PR #1, opened 2024-08-19
add `name` column in static-resources

aditya-mitra:
## Summary

- introduce name column in static-resources to find assets by their name

## References

closes https://tsu.atlassian.net/browse/IR-3155

## QA Steps

image1

image2

### DIFF
--- a/packages/client-core/src/admin/common/constants/resources.ts
+++ b/packages/client-core/src/admin/common/constants/resources.ts
@@ -27,7 +27,7 @@ import { t } from 'i18next'
 
 import { ITableHeadCell } from '../Table'
 
-type IdType = 'id' | 'key' | 'mimeType' | 'project' | 'action'
+type IdType = 'id' | 'name' | 'key' | 'mimeType' | 'project' | 'action'
 
 export type ResourceRowType = Record<IdType, string | JSX.Element | undefined>
 
@@ -37,6 +37,7 @@ interface IResourceColumn extends ITableHeadCell {
 
 export const resourceColumns: IResourceColumn[] = [
   { id: 'id', label: t('admin:components.resources.columns.id') },
+  { id: 'name', sortable: true, label: t('admin:components.resources.columns.name') },
   { id: 'key', sortable: true, label: t('admin:components.resources.columns.key') },
   { id: 'mimeType', sortable: true, label: t('admin:components.resources.columns.mimeType') },
   { id: 'project', sortable: true, label: t('admin:components.resources.columns.project') },
--- a/packages/client-core/src/admin/components/resources/AddEditResourceModal.tsx
+++ b/packages/client-core/src/admin/components/resources/AddEditResourceModal.tsx
@@ -27,7 +27,7 @@ import React, { useEffect } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import { PopoverState } from '@etherealengine/client-core/src/common/services/PopoverState'
-import { StaticResourceType, uploadAssetPath } from '@etherealengine/common/src/schema.type.module'
+import { staticResourcePath, StaticResourceType, uploadAssetPath } from '@etherealengine/common/src/schema.type.module'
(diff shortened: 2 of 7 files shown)


aditya-mitra:
Previous Discussion:

> Appu: Instead of introducing another column, can we modify the search criteria to search only the last string?
> Using regex in the query can work I think.

> Aditya: Regex would need to supported in our database - mysql. The other alternative is put such a regex in server or client on the database results. In this case, we would miss some search results because of pagination (for example, assume searching for vase produces 3 results with vase at the end of the string and 8 with vase at the middle of the string. In a pagination set of 10, there is strong possibility we will only see 2 results instead of 3 because of last one being in another page)

> Appaji: MySQL & MariaDB both of them support REGEXP. It isn't supported by feathers query operator. So, some workaround has to be made with action and modifying query at the server.
> For eg: In our database, below query searches through the file names, that has the word sky in them:
> image
> SELECT * FROM `static-resource` WHERE `key` REGEXP '\/[^\/]*(sky)[^\/]*$'
This will only search the string after the last slash, so only names are included in the search (project names, directories are not included).
> Test link: https://regex101.com/r/M5BGAX/2

> Josh: Let's include some unit tests for this please

> Josh: Does postgres support this?

> Appaji: It does: https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-POSIX-REGEXP. Syntax is a bit different.

aditya-mitra:
@CITIZENDOT Using separate column supports indexing and database search algorithms. Regex will loop over all the entries to test the validation.

We can use the regex if we have fewer entries. The static-resource table tends to be bigger in size, even if we just take the assets.

CITIZENDOT:
> A B-tree index can be used for column comparisons in expressions that use the =, >, >=, <, <=, or BETWEEN operators. The index also can be used for LIKE comparisons if the argument to LIKE is a constant string that does not start with a wildcard character. For example, the following SELECT statements use indexes:

https://dev.mysql.com/doc/refman/5.7/en/index-btree-hash.html

`LIKE` will use indices only if the pattern doesn't start with wildcard character. (but we have wildcard character at both start and end of the pattern). This can also be verified with `EXPLAIN` statement.

Query is using index when pattern is not starting wild card
- image

Query is not using index when pattern is starting wild card
- image

CITIZENDOT:
I dived into this little bit more, and found `FULLTEXT` index with `BOOLEAN MODE` query can be suited for our usecase. 

image

There is a limitation though (small one). Length of query string should be greater than or equal to 3. Docs say this length can be altered by setting [ngram_token_size](https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_ngram_token_size) at startup will fix this. But I couldn't reproduce this.

Btw, Using `BOOLEAN MODE` is a must. Otherwise database might consider some of the query string as noise.

Edit:
We're using MariaDB. The configuration variables are different from MySQL.

> - Words less than 4 (MyISAM) or 3 (InnoDB) characters in length will not be stored in the fulltext index. This value can be adjusted by changing the [ft_min_word_length](https://mariadb.com/kb/en/server-system-variables/#ft_min_word_len) system variable (or, for [InnoDB](https://mariadb.com/kb/en/innodb/), [innodb_ft_min_token_size](https://mariadb.com/kb/en/innodb-system-variables/#innodb_ft_min_token_size)).
> - Words longer than 84 characters in length will also not be stored in the fulltext index. This values can be adjusted by changing the [ft_max_word_length](https://mariadb.com/kb/en/server-system-variables/#ft_max_word_len) system variable (or, for [InnoDB](https://mariadb.com/kb/en/innodb/), [innodb_ft_max_token_size](https://mariadb.com/kb/en/xtradbinnodb-server-system-variables/#innodb_ft_max_token_size)).

https://mariadb.com/kb/en/full-text-index-overview/#excluded-results

aditya-mitra:
closing in favour of #316

PR #11, opened 2024-08-19
New GLTF ECS Loader

HexaField:
## References
closes #_insert number here_


HexaField:
@cla-bot check

HexaField:
@cla-bot check

packages/spatial/src/common/NameComponent.ts
```diff
@@ -63,3 +64,20 @@ export const NameComponent = defineComponent({
 
   entitiesByName: entitiesByName as Readonly<typeof entitiesByName>
 })
+
+export const recursiveNameLookup = (entity: Entity, targetName: string) => {
```
HexaField: can we just use `iterateEntityNode` for this?

packages/engine/src/avatar/AvatarBoneMatching.ts
```diff
@@ -645,6 +649,21 @@ export const recursiveHipsLookup = (model: Object3D) => {
   }
 }
 
+export const recursiveHipsLookupECS = (entity: Entity) => {
```
HexaField: can we use `iterateEntityNode` here

packages/engine/src/avatar/components/AnimationComponent.ts
```diff
@@ -36,3 +42,34 @@ export const AnimationComponent = defineComponent({
...
   })
 })
+
+export const useLoadAnimationFromBatchGLTF = (urls: string[], keepEntities = false) => {
+  const animations = urls.map((url) => useLoadAnimationFromGLTF(url, keepEntities))
+  const loadedAnimations = useHookstate(null as [AnimationClip[] | null, Entity][] | null)
+  useEffect(() => {
+    if (loadedAnimations.value || animations.some((animation) => !animation[0].value)) return
+    loadedAnimations.set(animations.map((animation) => [animation[0].get(NO_PROXY)!, animation[1]]))
+  })
```
HexaField: this useEffect should have dependencies

packages/engine/src/avatar/components/AnimationComponent.ts
```diff
@@ -36,3 +42,34 @@ export const AnimationComponent = defineComponent({
     animations: S.Array(S.Type<AnimationClip>())
   })
 })
+
+export const useLoadAnimationFromBatchGLTF = (urls: string[], keepEntities = false) => {
+  const animations = urls.map((url) => useLoadAnimationFromGLTF(url, keepEntities))
```
HexaField, line 59: can we guarantee that `urls` does not change length? if not, we need to use a subreactor to handle this

packages/engine/src/avatar/components/AnimationComponent.ts
```diff
@@ -36,3 +42,34 @@ export const AnimationComponent = defineComponent({
...
+  const animation = useHookstate(null as AnimationClip[] | null)
+  const animationComponent = useOptionalComponent(assetEntity, AnimationComponent)
+
+  useEffect(() => {
+    if (animation.value) return
+    if (!assetEntity) {
+      GLTFAssetState.loadScene(url, v4())
+      return
+    }
+  }, [useOptionalComponent(assetEntity, GLTFComponent)?.progress])
```
HexaField: prefer not to use hooks in deps array, we should declare this as a variable like we normally do

packages/engine/src/avatar/functions/avatarFunctions.ts
```diff
@@ -224,8 +226,12 @@ export const retargetAvatarAnimations = (entity: Entity) => {
  */
 export const setAvatarSpeedFromRootMotion = () => {
   const manager = getState(AnimationState)
-  const run = manager.loadedAnimations[preloadedAnimations.locomotion].animations[4] ?? [new AnimationClip()]
-  const walk = manager.loadedAnimations[preloadedAnimations.locomotion].animations[6] ?? [new AnimationClip()]
+  const animations = getComponent(
+    manager.loadedAnimations[preloadedAnimations.locomotion],
+    AnimationComponent
+  ).animations
+  const run = AnimationClip.findByName(animations, 'Run_RootMotion')
```
HexaField: let's make a note here to hande this generically in the future, and also put this magic string on a module level const

packages/engine/src/avatar/components/AvatarAnimationComponent.ts
```diff
@@ -103,58 +121,37 @@ export const AvatarRigComponent = defineComponent({
...
-      })
-
+      if (!rigComponent?.avatarURL?.value) return
+      setComponent(entity, GLTFComponent, { src: rigComponent.avatarURL.value })
       return () => {
-        removeEntity(helperEntity)
+        removeComponent(entity, GLTFComponent)
       }
-    }, [visible, debugEnabled, pending, rigComponent.normalizedRig])
+    }, [rigComponent?.avatarURL?.value])
```
HexaField: `avatarURL` is a redundant field, we can just set the `GLTFComponent` src directly

packages/engine/src/avatar/components/AvatarAnimationComponent.ts
```diff
@@ -103,58 +121,37 @@ export const AvatarRigComponent = defineComponent({
...
-      setupAvatarProportions(entity, model.asset as VRM)
-      setComponent(entity, AvatarRigComponent, {
-        vrm: model.asset as VRM,
-        avatarURL: model.src
-      })
+      if (gltfComponent?.progress?.value !== 100) return
+      console.log('Creating VRM')
+      const vrm = createVRM(entity)
+      setupAvatarProportions(entity, vrm)
+      rigComponent.vrm.set(vrm)
```
HexaField: now that we have the `VRMComponent`, having the `vrm` field on the AvatarRigComponent is redundant.
Alternatively, the VRMComponent could be seen as redundant - it was just an experiement that i think adds unnecessary complexity

packages/engine/src/avatar/systems/AvatarAnimationSystem.tsx
```diff
@@ -375,26 +333,45 @@ const Reactor = () => {
...
-        read: IKSerialization.readBlendWeight,
-        write: IKSerialization.writeBlendWeight
-      })
-
-      return () => {
-        networkState.networkSchema[IKSerialization.ID].set(none)
-      }
-    }, [])
-
-    if (!useMutableState(EngineState).viewerEntity.value) return null
```
HexaField, line 396: i think we still want to check that the viewer entity exists, or at least that there is a rendering context active. the original check here is to avoid loading these assets on pages that don't need them at load (such as the admin panel, which will use it if the user loads the avatar preview, but most of the time they don't). we could achieve this with maybe if a `useQuery([RendererComponent])`  is not empty

packages/engine/src/camera/components/AssetPreviewCameraComponent.ts
```diff
@@ -41,14 +43,15 @@ export const AssetPreviewCameraComponent = defineComponent({
...
+    const childMeshes = useChildrenWithComponents(previewCameraComponent.targetModelEntity.value, [MeshComponent])
     const cameraOrbitComponent = useComponent(entity, CameraOrbitComponent)
 
     useEffect(() => {
-      if (!modelComponent.scene.value) return
+      if (gltfComponent.progress.value !== 100) return
       cameraOrbitComponent.focusedEntities.set([previewCameraComponent.targetModelEntity.value])
       cameraOrbitComponent.refocus.set(true)
-    }, [modelComponent.scene, cameraOrbitComponent])
+    }, [gltfComponent.progress, childMeshes.length, cameraOrbitComponent])
```
HexaField: `.length` here is incorrect, the returned array itself should work as a reactive dep

packages/engine/src/avatar/components/AnimationComponent.ts
AidanCaruso: atm there is no use case I can think of for changing its length so I didn't bother with a subreactor. Currently its used for loading hard coded arrays of urls pointing to animation assets. not opposed to using a subreactor though as it might be useful in the future.

packages/engine/src/avatar/functions/avatarFunctions.ts
AidanCaruso: yeah this will need to be redone with upcoming animation system refactors, avatar animations should generally be handled generically, right now everything is magic strings and magic numbers.

packages/engine/src/avatar/components/AvatarAnimationComponent.ts
AidanCaruso: Hmm okay, I was actually thinking about breaking up the loading functionality from the rig component into the vrm component so as to allow avatar loading without querying by the avatar animation system. Indeed it might simplify things to just go with less components anyway.

packages/engine/src/gltf/GLTFLoader.test.tsx
```diff
@@ -0,0 +1,757 @@
...
+import { AssetLoaderState } from '../assets/state/AssetLoaderState'
+import { AnimationComponent } from '../avatar/components/AnimationComponent'
+import { GLTFComponent } from './GLTFComponent'
+import { GLTFDocumentState, GLTFNode, GLTFNodeState } from './GLTFDocumentState'
+import { getNodeUUID } from './GLTFState'
+import { KHRUnlitExtensionComponent, MaterialDefinitionComponent } from './MaterialDefinitionComponent'
+import { EXTMeshGPUInstancingComponent, KHRLightsPunctualComponent, KHRPunctualLight } from './MeshExtensionComponents'
+import { afterEach, beforeEach, describe, it } from 'vitest'
+
+const CDN_URL = 'https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/main/2.0'
```
HexaField: @MichaelEstes we should avoid remote dependencies for our tests. anything we use should be copied in the engine even if just for tests. any downtown or changes to the files mean our tests will start failing

packages/engine/src/gltf/GLTFLoader.test.tsx
```diff
@@ -0,0 +1,757 @@
...
+describe('GLTF Loader', () => {
+  beforeEach(async () => {
+    createEngine()
+  })
+
+  afterEach(() => {
+    return destroyEngine()
+  })
+
+  it('can load a mesh', () => new Promise<void>(done => {
```
HexaField: @MichaelEstes we should refactor these tests into less nested and async functions. if we have nested reactive logic for tests. we are more prone to timeouts as things get stuck, whereas as using `await act(() => rerender(<></>))` will cause all react updates to be handled  and we can immediately see what asserts fail.

HexaField:
@MichaelEstes the `ModelCompressPanel` has reference to `Heuristic.BUDGET` for LOD export, but that heuristic has been removed. are we wanting to add it back or change it to something else?

HexaField:
Looks like theres a weird regression with avatars. As soon as they load in, they go invisible. Went back a good few days of commits and it was still happening. Seems to be a race condition, sometimes doesn't happen.

MichaelEstes:
> @MichaelEstes the `ModelCompressPanel` has reference to `Heuristic.BUDGET` for LOD export, but that heuristic has been removed. are we wanting to add it back or change it to something else?

Converting model compression to use the GLTFComponent is what I was going to work on after finishing the thumbnail and that'll add back budget calculations

PR #15, opened 2024-08-19
added condition for the 'click to interact' to be enabled for the int…

MbfloydIR:
…eraction to call the callbacks

## Summary
https://tsu.atlassian.net/browse/IR-3404

## References
closes #_insert number here_


packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -319,8 +319,8 @@ export const InteractableComponent = defineComponent({
 
     InputComponent.useExecuteWithInput(
       () => {
+        if (!interactableComponent.clickInteract.value) return
```
SamMazerIR: this will likely break the E button interact due to the buttons.Interact alias being a combination of clicking and the E button

SamMazerIR (approved):

PR #18, opened 2024-08-19
Remove side effects from components

MichaelEstes:
## References
closes #_insert number here_


packages/engine/src/scene/components/GroundPlaneComponent.ts
```diff
@@ -43,16 +44,15 @@ export const GroundPlaneComponent = defineComponent({
 
   onInit(entity) {
     return {
-      color: new Color(),
+      color: 0xffffff as ColorRepresentation,
```
HexaField, line 47: with this change, make sure the studio still works correctly, as the editor might be expecting a color object

MichaelEstes: Ya, the editor color input has been updated to work with a color representation

packages/ui/src/primitives/tailwind/Color/index.tsx
```diff
@@ -50,7 +50,8 @@ export function ColorInput({
   textClassName,
   sketchPickerClassName
 }: ColorInputProp) {
-  const hexColor = typeof value.getHexString === 'function' ? '#' + value.getHexString() : '#000'
+  const color = new Color(value)
+  const hexColor = typeof color.getHexString === 'function' ? '#' + color.getHexString() : '#000'
```
HexaField: we can get rid of the inline conditional here, since getHexString is now guaranteed to always exist

HexaField (changes requested):

HexaField (approved):

HexaField:
looks like some other PR broke a test here

the-reality-engineer (approved):
Looks great!

PR #25, opened 2024-08-19
IR-3575 playmode on scene change

SamMazerIR:
## Summary
moving start/stop playmode into its own functions file, try exiting play mode on scene loading

## References
https://tsu.atlassian.net/browse/IR-3575

## QA Steps
open a scene, enter playmode, load a new scene without manually leaving playmode.

you should be able to properly see playmode has ended once the new scene is loaded, and you can enter/exit playmode as expected in the new scene

Repeat to your hearts content


AidanCaruso (approved):

HexaField:
This should not have been merged in. client-core cannot be imported by the spatial module

HexaField (commented):
I've reverted this PR. Please remake it. I think rather than having these functions be called by either the UI or wherverer it is used, we want these to be in a useEffect somewhere that only listens for changes to isEditing, and when isEditing is false, we spawn the avatar, otherwise despawn. we can then change isEditing to true when changing scenes. This creates a one to one dependency tree for what affects spawning the avatar or not.

PR #32, opened 2024-08-20
Non 3d page load optimizations

HexaField:
## References
closes #_insert number here_


aditya-mitra:
Why are we changing `Engine.instance.api` back to `API.instance`?

SamMazerIR (approved):

HexaField:
> Why are we changing `Engine.instance.api` back to `API.instance`?

Because we are trying to separate the core engine from the webapp. It also helps treeshaking, since `Engine` imports the ecs, and bitecs and threejs are quite large dependencies. This is increasing our load time on 2d pages significantly.

barankyle:
Seems to work fine, just need to fix linting and some type errors that check-errors is catching.

hanzlamateen (approved):

barankyle (approved):

PR #43, opened 2024-08-20
Update material on cascade change

MichaelEstes:
## References
closes #_insert number here_


packages/spatial/src/renderer/csm/CSM.ts
```diff
@@ -463,26 +459,23 @@ export class CSM {
...
+  getExtendedBreaks(): Vector2[] {
+    const target: Vector2[] = []
 
     for (let i = 0; i < this.cascades; i++) {
-      const amount = this.breaks[i]
+      const amount = this.breaks[i] || 0
       const prev = this.breaks[i - 1] || 0
-      target[i].x = prev
-      target[i].y = amount
+      target.push(new Vector2(prev, amount))
```
HexaField: why are we creating the vector here now? previously it was done lazily to avoid creaing garbage each frame

HexaField (changes requested):

packages/spatial/src/renderer/csm/CSM.ts
MichaelEstes: The only thing I could find to stop the Three errors when resizing the canvas was changing it so each shader's CSM uniform didn't reference the same array, having them reuse their unique arrays works too. I don't know why that fixed it though, if you have any ideas let me know

HexaField (approved):

PR #44, opened 2024-08-20
[IR-3907] move xr UI layer state and texture data store into hyperflux state

the-reality-engineer:
## References
closes #_insert number here_


packages/xrui/core/WebLayerManagerBase.ts
```diff
@@ -14,109 +14,33 @@ Software distributed under the License is distributed on an "AS IS" basis,
 WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for the
 specific language governing rights and limitations under the License.
 
-The Original Code is Infinite Reality Engine.
+The Original Code is Ethereal Engine.
```
HexaField, line 17: merge regression on license

PR #49, opened 2024-08-21
IR-3575 playmode on scene change better fix

SamMazerIR:
## Summary
playModeTool and sceneFunctions now react or use reactive state setting for isEditing state

## References
[https://tsu.atlassian.net/browse/IR-3575](https://tsu.atlassian.net/browse/IR-3575)

## QA Steps
enter a scene, enter playmode (with avatar) then load another scene

you should load into the new scene without issue, no longer in play mode and with the ability to start play mode again


packages/ui/src/components/editor/panels/Viewport/tools/PlayModeTool.tsx
```diff
@@ -47,46 +47,59 @@ import { HiOutlinePause, HiOutlinePlay } from 'react-icons/hi2'
...
-      // run all visual script logic
-      visualScriptQuery().forEach((entity) => dispatchAction(VisualScriptActions.execute({ entity })))
-      transformGizmoControlledQuery().forEach((entity) => removeComponent(entity, TransformGizmoControlledComponent))
-      //just remove all gizmo in the scene
-    }
+    getMutableState(EngineState).isEditing.set(!isEditing) //getState(EngineState).isEditing)
   }
 
+  useImmediateEffect(() => {
+    !isEditing.value ? onStartPlayMode() : onStopPlayMode()
```
HexaField: useImmediateEffect is unnecessary here, and we should call stop in a callback, such that if for whatever reason PlayModeTool was to unmount, it would still reverse the effect of spawning the avatar

HexaField (dismissed):

packages/ui/src/components/editor/panels/Viewport/tools/PlayModeTool.tsx
SamMazerIR: by callback do you mean the return of a useEffect?

HexaField: Yes

HexaField: we need to get rid of the conditional still

`!isEditing.value ? onStartPlayMode() : onStopPlayMode()`

we can instead have 

```ts
useEffect(() => {
  if (isEditing.value) return
  onStartPlayMode() 
  return () => onStopPlayMode()
}, [isEditing])
```

HexaField (dismissed):

the-reality-engineer (approved):

PR #54, opened 2024-08-21
Gltf loader 2.0 resource manager

MichaelEstes:
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/gltf/GLTFLoaderFunctions.ts
+++ b/packages/engine/src/gltf/GLTFLoaderFunctions.ts
@@ -444,7 +444,7 @@ const useLoadMaterial = (
   options: GLTFParserOptions,
   materialDef: ComponentType<typeof MaterialDefinitionComponent>
 ) => {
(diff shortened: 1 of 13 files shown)


packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -327,9 +330,17 @@ const useLoadBufferView = (options: GLTFParserOptions, bufferViewIndex?: number)
 
 const useLoadBuffer = (options: GLTFParserOptions, bufferIndex) => {
   const json = options.document
-  const result = useHookstate<ArrayBuffer | null>(null)
+  const loader = useHookstate(() => {
+    const fileLoader = new FileLoader(options.manager)
+    fileLoader.setResponseType('arraybuffer')
+    if (options.crossOrigin === 'use-credentials') {
```
HexaField: i think we can eliminate crossOrigin here altogether, we never set it to anything other than `anonymous`

HexaField (changes requested):

HexaField (approved):

PR #57, opened 2024-08-21
Made github-repo-access-refresh runnable as a job

barankyle:
## Summary

Logging in with GitHub was taking notably longer than other SSO methods because it was performing a refresh of the user's Github repo access. Made a new service to call this function that can be run as a job, and running it as a job on Kubernetes.

createExecutorJob now has an optional param to not wait for the job to finish before resolving for this and similar situations where you don't care to check if the job is finished. Default is to waitForFinish to be true.

## References
closes #_insert number here_


packages/server-core/src/user/strategies/github.ts
```diff
@@ -47,6 +51,24 @@ export class GithubStrategy extends CustomOAuthStrategy {
...
+      endTime: date,
+      returnData: '',
+      status: 'pending'
+    })
+
+    const jobBody = await getGithubRepoAccessRefreshJobBody(this.app, newJob.id, userId)
+    await this.app.service(apiJobPath).patch(newJob.id, {
+      name: jobBody.metadata!.name
+    })
+    const jobLabelSelector = `ir-engine/userId=${userId},ir-engine/release=${process.env.RELEASE_NAME},ir-engine/autoUpdate=false`
```
hanzlamateen, line 68: What is the use of `,ir-engine/autoUpdate=false` here?

packages/server-core/src/user/github-repo-access-refresh/github-repo-access-refresh.class.ts
```diff
@@ -31,13 +31,45 @@ import {
...
+    'scripts/refresh-gh-repo-access.ts',
+    '--userId',
+    userId,
+    '--jobId',
+    jobId
+  ]
+
+  const labels = {
+    'ir-engine/ghRepoAccessRefresh': 'true',
+    'ir-engine/autoUpdate': 'false',
```
hanzlamateen, line 63: What is the user of `'ir-engine/autoUpdate': 'false',`

hanzlamateen (changes requested):

packages/server-core/src/user/strategies/github.ts
barankyle: It looks like we set ir-engine/autoUpdate to true for when we make a cronjob, which I think only happens if someone sets a project to auto-update. So it's set to false if it's a one-off job.

hanzlamateen: https://github.com/ir-engine/ir-engine/blob/08d42e2d669735dbb0c7521e2199b7627deae097/packages/server-core/src/assets/model-transform/model-transform.helpers.ts#L47
https://github.com/theinfinitereality/irpro-multitenancy/blob/6c41fec421c3d3072c1d424d6726b005dc08b481/services/domain-provision-helper.ts#L136
https://github.com/theinfinitereality/irpro-multitenancy/blob/6c41fec421c3d3072c1d424d6726b005dc08b481/services/account-deletion-helper.ts#L35

I dont see it being set in any of these.

barankyle: I don't think it's necessary, as nothing is specifically looking for that label, it was just added for potential clarity, in case anyone was looking at a job and might benefit from seeing whether it was an auto-updating job. If MT jobs aren't adding that label at all it's fine.

hanzlamateen (approved):

PR #62, opened 2024-08-22
fix(IR-4170): adding id attributes for testing, scene and Modal component

kattsushi:
## Summary
fix(IR-4170): adding id attributes for testing, scene and Modal component

## Subtasks Checklist

- id for scene item : `default-scene`
- id for Modal component: `${parent-modal-id}-close-button` and `${parent-modal-id}-submit-button`

## References
closes Jira issue number #[IR-4170](https://tsu.atlassian.net/browse/IR-4170)

## QA Steps
Screenshot 2024-08-22 at 12 02 50 p m

Screenshot 2024-08-22 at 12 04 14 p m

### DIFF
--- a/packages/client-core/src/admin/components/scene/SceneItem.tsx
+++ b/packages/client-core/src/admin/components/scene/SceneItem.tsx
@@ -6,8 +6,8 @@ Version 1.0. (the "License"); you may not use this file except in compliance
 with the License. You may obtain a copy of the License at
 https://github.com/ir-engine/ir-engine/blob/dev/LICENSE.
 The License is based on the Mozilla Public License Version 1.1, but Sections 14
-and 15 have been added to cover use of software over a computer network and 
-provide for limited attribution for the Original Developer. In addition, 
+and 15 have been added to cover use of software over a computer network and
+provide for limited attribution for the Original Developer. In addition,
 Exhibit A has been modified to be consistent with Exhibit B.
 
 Software distributed under the License is distributed on an "AS IS" basis,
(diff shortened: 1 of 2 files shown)


packages/ui/src/primitives/tailwind/Modal/index.tsx
```diff
@@ -112,6 +121,7 @@ export const ModalFooter = ({
 }
 
 const Modal = ({
+  id,
```
aditya-mitra, line 124: this may not be an ideal case
footer can be hidden using the `hideFooter` prop
and also `id` on modal should go inside the modal's direct children

aditya-mitra (changes requested):

packages/ui/src/primitives/tailwind/Modal/index.tsx
kattsushi: I'm not entirely sure if I understood you correctly, but are you suggesting that I hide the footer and create a custom one inside the modal content? Honestly, I don't think it's really necessary to do that because modals could lose their consistency in different cases, and I think it goes a bit beyond the Open/Closed Principle. However, I did change the button identifiers to data-test-id=, and I think that works. What are your thoughts?

packages/client-core/src/admin/components/scene/SceneItem.tsx
```diff
@@ -102,7 +102,10 @@ export const SceneItem = ({
   }, [threeDotsContainRef])
 
   return (
-    <div className="col-span-2 inline-flex h-64 w-64 min-w-64 max-w-64 cursor-pointer flex-col items-start justify-start gap-3 rounded-lg bg-theme-highlight p-3 lg:col-span-1">
+    <div
+      id={`${sceneName === 'New-Scene' ? 'default-scene' : sceneName}`}
```
aditya-mitra: let's also change this to `data-test-id`

kattsushi: forgot that one, thx!

packages/ui/src/primitives/tailwind/Modal/index.tsx
aditya-mitra: How about we add a `data-test-id` to [the outermost div](https://github.com/ir-engine/ir-engine/blob/87ecd255f0c6a2420ac9a911c07183fa4f250fd6/packages/ui/src/primitives/tailwind/Modal/index.tsx#L131) ?

This div is always present and can be used for testing.

aditya-mitra (approved):

PR #65, opened 2024-08-22
IR-3984 viewer transition problems

dinomut1:
Fixes for bugs observed when switching between engine viewer canvases in the wizard. 

Pointer entities are generated in ClientInputHooks which get lost when viewer canvas gets unmounted. Since this data is only tracked in a one way reference from the PointerComponent to the camera entity associated with a viewer, the simplest fix is to detect stale pointers by checking if their camera entity still exists each frame

Also, the initialization of the viewer and of the origin have been split apart so that the scene doesn't need to be unloaded and reloaded every time a canvas is created or destroyed. Instead the base spatial environment is initialized on its own in initializeSpatialEngine, and a new function, initializeSpatialViewer, initializes the viewer. This way we can create and destroy viewers and preserve the underlying scene

Closes https://tsu.atlassian.net/browse/IR-3984

### DIFF
--- a/packages/editor/src/components/EditorContainer.tsx
+++ b/packages/editor/src/components/EditorContainer.tsx
@@ -56,6 +56,7 @@ import { FeatureFlags } from '@ir-engine/common/src/constants/FeatureFlags'
 import { EntityUUID } from '@ir-engine/ecs'
 import useFeatureFlags from '@ir-engine/engine/src/useFeatureFlags'
 import { EngineState } from '@ir-engine/spatial/src/EngineState'
+import { destroySpatialEngine, initializeSpatialEngine } from '@ir-engine/spatial/src/initializeEngine'
 import Button from '@ir-engine/ui/src/primitives/tailwind/Button'
 import 'rc-dock/dist/rc-dock.css'
 import { useTranslation } from 'react-i18next'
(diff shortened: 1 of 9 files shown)


SamMazerIR (approved):

packages/editor/src/components/EditorContainer.tsx
```diff
@@ -168,12 +169,19 @@ const EditorContainer = () => {
     }
   }, [scenePath.value])
 
-  const viewerEntity = useMutableState(EngineState).viewerEntity.value
+  useEffect(() => {
+    initializeSpatialEngine()
+    return () => {
+      destroySpatialEngine()
+    }
+  }, [])
```
Rezmason, line 177: I'd like to better understand how often this `useEffect` is run; it has no dependencies, but it initializes and destroys some things that look like a big deal, so to my undertrained eye it looks conspicuous

Rezmason (approved):

packages/editor/src/components/EditorContainer.tsx
Rezmason: "Runs only once, when the editor container is mounted."

packages/spatial/src/input/systems/ClientInputSystem.tsx
```diff
@@ -106,9 +107,16 @@ const execute = () => {
     getMutableComponent(eid, InputComponent).inputSources.set([])
   }
 
+  const stalePointers: Entity[] = []
+
   // update 2D screen-based (driven by pointer api) input sources
   for (const eid of pointersQuery()) {
     const pointer = getComponent(eid, InputPointerComponent)
+    // check if pointer camera entity still exists
```
heysokam, line 115: This PR will be merged soon: https://github.com/ir-engine/ir-engine/pull/26
These new `stalePointers` blocks of code are not tested, but probably should have specific cases created for them :writing_hand:

HexaField (approved):

packages/spatial/src/input/systems/ClientInputSystem.tsx
heysokam: I will update #26 then, since this was merged first into dev

PR #77, opened 2024-08-26
Remove MUI imports and use tailwind where necessary

CITIZENDOT:
## References
partially fixes https://tsu.atlassian.net/browse/IR-4236


packages/client/index.html
```diff
@@ -60,7 +59,7 @@
     </script>
 </head>
   <body>
-    <div id="root"  style='z-index: 1; margin: 0; width: 100%; height: 100%; position: absolute; pointer-events: none;'></div>
+    <div id="root"  style='z-index: 1; margin: 0; width: 100%; height: 100vh; position: absolute;'></div>
```
HexaField: Have you tested both location and studio after removing `pointer-events: none`? It would be good to get rid of it, but want to make sure this doesn't cause regressions.

CITIZENDOT: studio is fine, but locations is not responding to mouse events. i'll find some other way.

CITIZENDOT: fixed

HexaField (approved):

PR #80, opened 2024-08-26
Michael/ir 4091 video

MbfloydIR:
## Summary

updated the fit placement logic for 
- cover
- contain
- horizontal
- vertical 

https://tsu.atlassian.net/browse/IR-4091
https://tsu.atlassian.net/browse/IR-4090
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/scene/components/VideoComponent.tsx
+++ b/packages/engine/src/scene/components/VideoComponent.tsx
@@ -332,6 +332,17 @@ function VideoReactor() {
       uvScale.y = 1
       uvOffset.x = (1 - uvScale.x) / 2
     }
+    if (video.fit.value == 'contain') {
+      if (imageRatio > containerRatio) {
+        uvScale.y = imageRatio / containerRatio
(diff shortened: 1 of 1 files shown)


packages/engine/src/scene/components/VideoComponent.tsx
```diff
@@ -296,15 +301,55 @@ function VideoReactor() {
...
+    const containerHeight = video.size.height.value
+    const containerRatio = containerWidth / containerHeight
+
+    videoMesh.scale.x = containerWidth
+    videoMesh.scale.y = containerHeight
+
+    const imageSize = getTextureSize(videoMesh.material.uniforms.map.value as Texture | CompressedTexture)
+    const imageRatio = imageSize.x / imageSize.y || 1
+
+    let placement = 'horz'
```
SamMazerIR: would prefer this to be a bool/enum (enum only if there would be more options)

SamMazerIR (approved):

PR #81, opened 2024-08-26
Fix trigger state getting lost by regenerated colliders

dinomut1:
Fixes an issue where avatars cannot enter trigger colliders


HexaField (approved):

HexaField:
Let's add a regression test for this

heysokam:
The unit test that covers the isSensor property is here, which seems to be passing:
https://github.com/ir-engine/ir-engine/blob/6e2201ef436c68607e5eaaa95cc49a116baca7f3/packages/spatial/src/physics/classes/Physics.test.tsx#L1616-L1622
We likely need a separate test to ensure that `setCollisionGroups` was called.

dinomut1:
> The unit test that covers the isSensor property is here, which seems to be passing:
> 
> https://github.com/ir-engine/ir-engine/blob/6e2201ef436c68607e5eaaa95cc49a116baca7f3/packages/spatial/src/physics/classes/Physics.test.tsx#L1616-L1622
> 
> 
> We likely need a separate test to ensure that `setCollisionGroups` was called.

Yeah the only issue here was that the collision group was getting lost when the collider gets regenerated in the PreTransformSystem

PR #83, opened 2024-08-26
Ir 4123 compression with org name

Rezmason:
## Summary

Fixes a bug that broke model compression when we nested our projects within org folders.

closes https://tsu.atlassian.net/browse/IR-4123

## QA Steps
Compress a model, drag its integrated GLTF into a scene and confirm that the resulting new object in the hierarchy contains a "LOD-variant" child that has "Variant" components pointing to the different LOD levels you generated.


dinomut1 (dismissed):
Seeing an issue with the compression workflow. It's hard coded to only generate .glb files. We need to generate .gltf files instead

barankyle (approved):

Rezmason:
> Seeing an issue with the compression workflow. It's hard coded to only generate .glb files. We need to generate .gltf files instead

Fixed; the GLB-or-GLTF logic was tripping on our hashes

dinomut1 (approved):

PR #89, opened 2024-08-27
Components Schema

MichaelEstes:
## References
closes #_insert number here_


the-reality-engineer:
Looking great so far.  Much cleaner.

packages/ecs/src/schemas/JSONSchemaTypes.ts
```diff
@@ -0,0 +1,238 @@
...
+The Original Code is Infinite Reality Engine.
+
+The Original Developer is the Initial Developer. The Initial Developer of the
+Original Code is the Infinite Reality Engine team.
+
+All portions of the code written by the Infinite Reality Engine team are Copyright © 2021-2023 
+Infinite Reality Engine. All Rights Reserved.
+*/
+
+export const Kind = Symbol('Kind')
```
HexaField, line 26: If this implementation is a copy of derivative of typebox (or any library) we need to explicitly state as such at the top of each file (under our license) as well as include their license in the folder containing the bulk of the copy. We need to respect whatever their license states too.

HexaField:
If there are projects this affects (which there are) we should keep this as a draft until they have PRs up too.

packages/ecs/src/ComponentFunctions.ts
```diff
@@ -265,18 +346,24 @@ export const defineComponent = <
...
+export const getOptionalMutableComponent = <
+  Schema extends ComponentSchema,
+  InitializationType,
+  ComponentType,
+  JSON,
+  SetJSON
+>(
   entity: Entity,
-  component: Component<ComponentType, Record<string, any>, unknown>
+  component: Component<Schema, InitializationType, ComponentType, JSON, SetJSON, unknown>
```
HexaField: can we make a type instead of invoking `<Schema, InitializationType, ComponentType, JSON, SetJSON, unknown>` a lot

HexaField (changes requested):

packages/ecs/src/schemas/JSONSchemaTypes.ts
MichaelEstes: What counts as a derivative? I kept the Kind and static properties, but the rest is unique, Typebox was too inflexible to have typings for Required and NonSerialized fields

HexaField (approved):

PR #98, opened 2024-08-28
format only staged files

aditya-mitra:
## Summary

prettier format only staged files


package.json
```diff
@@ -103,7 +103,7 @@
   ],
   "lint-staged": {
     "*.{ts,tsx}": [
-      "prettier --write \"**/*.{ts,tsx}\""
+      "prettier --write"
```
HexaField, line 107: Hmm I'm not sure what's happening here. Can you test to see if changing the pattern in the lint-staged definition itself (L#105) needs to be changed? It was not formatting any files for me in projects, so maybe the extra pattern is unnecessary in the core engine but necessary in projects?

aditya-mitra: actually, `lint-staged` passes [the files to the command to work](https://github.com/lint-staged/lint-staged/blob/0ce5e1455e28d3376cb13de98b1023af1e60942b/lib/makeCmdTasks.js#L29)
these files in turn are passed to [the prettier command](https://arc.net/l/quote/wulgjnno)
thus, we can skip specifying the files

The prettier formatting on projects was non-functional because the git precommit hook was missing. I have added a new scripts/add_precommit.sh to add the git precommit hook in all projects (after npm install is finished) and lint staged files.

we can also remove the explicit [prettier config on mt](https://github.com/theinfinitereality/irpro-multitenancy/blob/dev/.prettierrc.json) because it almost the same as the root config

scripts/add-precommit.sh
```diff
@@ -0,0 +1,24 @@
...
+    if [ ! -e ".git/hooks/pre-commit" ]; then
+      rm .git/hooks/pre-commit
+    fi
+    echo "$pre_commit_commands" > .git/hooks/pre-commit
+    chmod ug+x .git/hooks/pre-commit
+    echo "added pre_commit hook in $directory"
+  fi
+}
+
+for directory in packages/projects/projects/ir-engine/* ;
```
HexaField: let's do this for all orgs too

HexaField (changes requested):
awesome :D

scripts/add-precommit.sh
aditya-mitra: done

HexaField (approved):

PR #100, opened 2024-08-28
studio: studio user profile

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-2520
## References
closes #_insert number here_


packages/editor/src/components/profile/Profile.tsx
```diff
@@ -0,0 +1,245 @@
...
+import { FaArrowRightToBracket } from 'react-icons/fa6'
+import { LuSettings } from 'react-icons/lu'
+import { MdOutlineKeyboardArrowDown } from 'react-icons/md'
+import { PiEyeBold, PiEyeClosedBold, PiTrashSimple } from 'react-icons/pi'
+
+const ProfileModal = ({ user }) => {
+  const { t } = useTranslation()
+  const selfUser = useHookstate(getMutableState(AuthState).user)
+
+  useEffect(() => {
```
barankyle, line 48: I don't think we need anything that deals with oauthConnectedState in this file, since we're not displaying any of that information. Setting it to the initial state on logout on line 100 is also unnecessary for that reason.

packages/editor/src/components/profile/Profile.tsx
```diff
@@ -0,0 +1,245 @@
...
+        <hr className="mx-4 text-[#A0A1A2] opacity-10" />
+        <div className="px-4 py-2">
+          <div className="grid grid-cols-1 gap-3">
+            <InputGroup
+              className="mr-0 flex flex-row items-center justify-start"
+              label="Invite Code"
+              labelClassName="text-left w-20"
+            >
+              <StringInput
+                type={showKeys.inviteCode.value ? 'text' : 'password'}
```
barankyle, line 141: Did someone request that these fields be hidden by default and then toggleable? We don't hide the invite code or userId anywhere else.

packages/editor/src/components/profile/Profile.tsx
```diff
@@ -0,0 +1,245 @@
...
+          <span className="text-md ml-4 flex cursor-pointer items-center gap-2 text-[#A0A1A2]">
+            <LuSettings />{' '}
+            <span className="text-xs" onClick={() => PopupMenuServices.showPopupMenu(UserMenus.Settings)}>
+              Settings
+            </span>
+          </span>
+          <span className="text-md ml-4 flex cursor-pointer items-center gap-2 text-[#A0A1A2]" onClick={handleLogout}>
+            <FaArrowRightToBracket /> <span className="text-xs">Log out</span>
+          </span>
+          <span className="text-md ml-4 flex cursor-pointer items-center gap-2 text-[#A0A1A2]">
```
barankyle, line 197: This button doesn't do anything. I don't know if it's supposed to be here - my take would be to make users go to the dashboard to delete their account and not have a delete account button in the studio, but I'm not in charge of UX decisions. If it is supposed to be here, then it needs to do something, and I would imagine there should be a confirmation prompt so a stray click doesn't wipe your account.

barankyle (changes requested):

achen5671:
closing, replacing with a much simpler pr https://github.com/ir-engine/ir-engine/pull/347/files with new designs

PR #105, opened 2024-08-28
avatar bone recursion

anselm:
## Summary

There is a subtle issue with some vrms, sometimes, not finding matching bone parts. There does appear to be a defect in the recursive bone search for hips that it returns on the first item of a collection. It may be that the bone order is not deterministic and that in some cases it causes a bone to fail.

It is not reproducible for me on my machine, my change however seems to work (as in not break the local instance on my machine). But there is a chance that it breaks dev.

## References
closes #_insert number here_


HexaField:
Looks good to me. Will wait on @AidanCaruso for final say.

AidanCaruso (approved):

PR #108, opened 2024-08-29
IR-4305: Instance table enhancements

MoizAdnan:
https://tsu.atlassian.net/browse/IR-4305


barankyle (approved):

hanzlamateen (changes requested):
Unable to see record if user was banned on a location and then for new instance, he does not appear in the dialog.

Do we need to leverage location-ban table for ban instead of user-kick?

hanzlamateen (approved):

PR #78, opened 2024-08-26
IR-4256 Fix model component save changes

AidanCaruso:
## Summary
Fixes model component save changes to unblock issues with saving lightmaps to materials

## References
closes #_insert number here_

PR #110, opened 2024-08-29
Model compression progress bar

Rezmason:
## Summary
The model transform function now reports its progress, and the compression panel now has a captioned progress bar instead of a spinner.

## References
closes https://tsu.atlassian.net/browse/IR-4271

### DIFF
--- a/packages/editor/src/components/assets/ModelCompressionPanel.tsx
+++ b/packages/editor/src/components/assets/ModelCompressionPanel.tsx
@@ -28,7 +28,10 @@ import { twMerge } from 'tailwind-merge'
 import { LoaderUtils } from 'three'
 
 import { API } from '@ir-engine/common'
-import { transformModel as clientSideTransformModel } from '@ir-engine/common/src/model/ModelTransformFunctions'
+import {
+  transformModel as clientSideTransformModel,
+  ModelTransformStatus
+} from '@ir-engine/common/src/model/ModelTransformFunctions'
 import { modelTransformPath } from '@ir-engine/common/src/schema.type.module'
 import { setComponent } from '@ir-engine/ecs/src/ComponentFunctions'
 import {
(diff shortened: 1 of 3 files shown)


dinomut1 (dismissed):
After running a default compression job on a file in "ir-props", the integrated prefab file is getting uploaded into /assets/assets/models rather than /assets/models
image

Rezmason:
> After running a default compression job on a file in "ir-props", the integrated prefab file is getting uploaded into /assets/assets/models rather than /assets/models

image

packages/editor/src/components/assets/ModelCompressionPanel.tsx
```diff
@@ -51,17 +54,32 @@ import { SourceComponent } from '@ir-engine/engine/src/scene/components/SourceCo
...
 import GLTFTransformProperties from '../properties/GLTFTransformProperties'
 
+const progressCaptions: Map<ModelTransformStatus, string> = new Map([
+  [ModelTransformStatus.Initializing, 'editor:properties.model.transform.status.initializing'],
+  [ModelTransformStatus.ProcessingTexture, 'editor:properties.model.transform.status.processingtexture'],
+  [ModelTransformStatus.Finalizing, 'editor:properties.model.transform.status.finalizing'],
+  [ModelTransformStatus.WritingFiles, 'editor:properties.model.transform.status.writingfiles'],
+  [ModelTransformStatus.Complete, 'editor:properties.model.transform.status.complete']
+])
+
```
aditya-mitra: in this case, the key value pairs are known. so an plain object would work here. we should utilise a map when we are dynamically adding or removing key values.

packages/engine/src/assets/compression/ModelTransformFunctions.ts
```diff
@@ -343,9 +343,24 @@ function hashBuffer(buffer: Uint8Array): string {
...
+export { Status as ModelTransformStatus }
+
 export async function transformModel(
   args: ModelTransformParameters,
-  onMetadata: (key: string, data: any) => void = (key, data) => {}
+  onMetadata: (key: string, data: any) => void = (key, data) => {},
+  onProgress: (progress: number, status: Status, numerator?: number, denominator?: number) => void = (
+    numerator,
+    denominator,
+    status
```
aditya-mitra:
```suggestion
    _numerator,
    _denominator,
    _status
```

we can use `_` underscore to denote unused variable in this case

aditya-mitra (dismissed):
few changes

packages/engine/src/assets/compression/ModelTransformFunctions.ts
Rezmason: These are in a function declaration that's a default value for a function parameter, so I'll just delete the default value and make the function parameter optional.

Believe it or not, I think the formatter wrote these out. I was using `() => {}` as the default value and I guess the formatter didn't like my arg erasure. It's a funny old world

packages/editor/src/components/assets/ModelCompressionPanel.tsx
Rezmason: TIL that the TypeScript Record type with an enum key is exhaustive, ie. it throws a useful compiler error if there are any missing enum keys in the declaration ✨

Rezmason:
> After running a default compression job on a file in "ir-props", the integrated prefab file is getting uploaded into /assets/assets/models rather than /assets/models 

Found to be caused by https://github.com/ir-engine/ir-engine/pull/78

dinomut1 (approved):

PR #113, opened 2024-08-29
V0.8(No Merge without QA approval) Disable login Link

DanielBelmes:
## References
closes #_insert number here_


Rezmason:
From a young age I was taught that, if someone comments out code, they ought to document it with their reasoning.

larryatinfinite (approved):

DanielBelmes:
> From a young age I was taught that, if someone comments out code, they ought to document it with their reasoning.

Very wise. Added

PR #114, opened 2024-08-29
Int v0.8 - updated conditional logic to prevent hover e interactions on proximity

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-4320

## References
closes #_insert number here_


SamMazerIR (approved):

packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -323,7 +324,13 @@ export const InteractableComponent = defineComponent({
     InputComponent.useExecuteWithInput(
       () => {
         const buttons = InputComponent.getMergedButtons(entity)
-        if (!interactableComponent.clickInteract.value && buttons.PrimaryClick?.pressed) return
+        if (
+          !activateUI ||
```
HexaField: This is scoping a global variable from a different context. We need a different solution.

HexaField (dismissed):

packages/engine/src/interaction/components/InteractableComponent.ts
MbfloydIR: implemented

packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -272,6 +273,8 @@ export const InteractableComponent = defineComponent({
     if (json.label) component.label.set(json.label)
     if (typeof json.uiActivationType === 'number' && component.uiActivationType.value !== json.uiActivationType)
       component.uiActivationType.set(json.uiActivationType)
+    if (typeof json.canInteract === 'boolean' && component.canInteract.value !== json.canInteract)
+      component.canInteract.set(json.canInteract)
```
SamMazerIR: I don't think we need to serialize canInteract. It should always represent a current state.

SamMazerIR (changes requested):
lets remove serialization of canInteract

packages/engine/src/interaction/components/InteractableComponent.ts
MbfloydIR: we need to serialize it so the user can check/uncheck the can interact editor ui checkbox, as well as loading the saved value when the scene is loaded

SamMazerIR (approved):

PR #116, opened 2024-08-29
IR-4006 incorrect highlight state when hitting escape

SamMazerIR:
## Summary
after some digging, the incorrect highlight state mentioned in IR-4006 is a result of document.activeElement focus when hitting Esc. Blurring the element if it is an HTMLElement (as others do not have blur()). This is being done in EditorControlSystem to fix it editor-wide when pressing Escape (as this was a visual issue on any element you could interact with)

## References
[https://tsu.atlassian.net/browse/IR-4006](https://tsu.atlassian.net/browse/IR-4006)

## QA Steps
select an element or hit a button (... buttons, publish button)

hit the escape key

observe there is no longer a white and blue border around the element like before


aditya-mitra (approved):
Works good ✅

PR #118, opened 2024-08-30
Improved login flow to prompt associations with existing accounts

barankyle:
## Summary

If a new account's identity provider is associated with an email address that is associated with other identity providers, the user is now prompted to associate their new login with the existing identity-provider. This only happens the first time; if they decline, then any subsequent logins will skip that step, and are kept as a separate account.

Fixed a bug with getting backend errors out of query params. The code was reading SearchParamState.error, which is a separate property that usually returns undefined. Now getting the .value of the state before reading the keyID.

Changed magiclink login flow to use ID as the /login identifier and token is now passed as a query param, instead of the token being the identifier. This should make it nearly impossible to brute-force guess a magiclink.

Resolves IR-3991

## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/src/user/components/MagicLink/AuthMagicLink.tsx
+++ b/packages/client-core/src/user/components/MagicLink/AuthMagicLink.tsx
@@ -27,23 +27,28 @@ import React, { useEffect } from 'react'
 import { useTranslation } from 'react-i18next'
 import { useLocation } from 'react-router-dom'
 
+import { config } from '@ir-engine/common/src/config'
 import { InstanceID } from '@ir-engine/common/src/schema.type.module'
-import Box from '@ir-engine/ui/src/primitives/mui/Box'
-import Container from '@ir-engine/ui/src/primitives/mui/Container'
-import Typography from '@ir-engine/ui/src/primitives/mui/Typography'
 
+import Button from '@ir-engine/ui/src/primitives/mui/Button'
+import LoadingView from '@ir-engine/ui/src/primitives/tailwind/LoadingView'
+import Text from '@ir-engine/ui/src/primitives/tailwind/Text'
+import { BsFillExclamationTriangleFill } from 'react-icons/bs'
 import { AuthService } from '../../services/AuthService'
 
-interface Props {
-  //auth: any
-  type: string
-  token: string
-  instanceId: InstanceID
-  path: string
-}
-
-const AuthMagicLink = ({ token, type, instanceId, path }: Props): JSX.Element => {
+const AuthMagicLink = (): JSX.Element => {
   const { t } = useTranslation()
+
+  const search = new URLSearchParams(useLocation().search)
+  const token = search.get('token') as string
+  const type = search.get('type') as string
+  const path = search.get('path') as string
+  const instanceId = search.get('instanceId') as InstanceID
+  const promptForConnection = search.get('promptForConnection') as string
+  const loginId = search.get('loginId') as string
+  const loginToken = search.get('loginToken') as string
+  const email = search.get('associateEmail') as string
+
   useEffect(() => {
     if (type === 'login') {
       let redirectSuccess = path ? `${path}` : null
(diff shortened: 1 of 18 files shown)


packages/client-core/src/common/services/RouterService.tsx
```diff
@@ -87,7 +87,7 @@ export const useSearchParamState = () => {
   /** Create a nested reactor to react to state changes */
   useEffect(() => {
     const SubReactor = (props: { keyID: string }) => {
-      const value = useHookstate(getMutableState(SearchParamState)[props.keyID]).value
+      const value = getMutableState(SearchParamState).value[props.keyID]
```
HexaField: why the change from useHookstate to getMutableState here? this means it will not trigger reactive updates to the react context

packages/client-core/src/user/components/MagicLink/AuthMagicLink.tsx
```diff
@@ -55,25 +59,73 @@ const AuthMagicLink = ({ token, type, instanceId, path }: Props): JSX.Element =>
     }
   }, [])
 
+  function doRedirect(connect = false) {
+    let redirect = /https:\/\//.test(path) ? path : config.client.clientUrl + path
+    console.log('added path', redirect)
```
HexaField: console logs

packages/client-core/src/user/components/MagicLink/AuthMagicLink.tsx
```diff
@@ -55,25 +59,73 @@ const AuthMagicLink = ({ token, type, instanceId, path }: Props): JSX.Element =>
...
+    if (instanceId != null) redirect += `?instanceId=${instanceId}`
```
HexaField, line 65: we have many more parameters than just `instanceId` (invites, roomcodes, studio project and scene path etc), do we want to make this generic somehow?

packages/server-core/src/user/login/login.class.ts
```diff
@@ -77,9 +79,62 @@ export class LoginService implements ServiceInterface {
...
+      const logins = await this.app.service(userLoginPath).find({
+        query: {
+          userId: identityProvider.userId
+        }
+      })
+      //Email identity-providers are created as type email, so new vs. existing logins can't be discerned by
+      //whether the current identity-provider is a guest. We're using logins + addToLogin being 0 as a proxy for
+      //a brand-new login with that email, which should trigger the auto-association, and any total greater than
+      //0 will be seen as an email that is established and shouldn't have the auto-association trigger
+      if (identityProvider.type === 'email' && logins.total + addToLogin === 0) {
```
HexaField: this is a little indirectional using a number for essentially a boolean operation, can we make `addToLogin` a boolean and the logic here can be `logins.total === 0 && !addToLogin` and `addToLogin` can be equal to `params!.query?.associate != null`

HexaField (changes requested):

packages/client-core/src/common/services/RouterService.tsx
barankyle: I can't remember why I changed that, and it was still working when I reverted it, so I've undone this.

packages/client-core/src/user/components/MagicLink/AuthMagicLink.tsx
barankyle: As discussed, this component is just used for resolving magiclink logins, which currently can only redirect to instances. If we enable redirects to e.g. a specific project/scene in the studio, we'll deal with it at that time.

packages/server-core/src/user/login/login.class.ts
barankyle: Fixed

packages/server-core/src/user/login/login.class.ts
```diff
@@ -77,9 +79,63 @@ export class LoginService implements ServiceInterface {
...
+      //whether the current identity-provider is a guest. We're using logins === 0 and !addToLogin as a proxy for
+      //a brand-new login with that email, which should trigger the auto-association, and that not being true
+      //will be seen as an email that is established and shouldn't have the auto-association trigger
+      if (identityProvider.type === 'email' && logins.total === 0 && !addToLogin) {
+        console.log('')
+        const existingIdentityProviders = await this.app.service(identityProviderPath).find({
+          query: {
+            $or: [
+              {
+                email: identityProvider.token
```
HexaField, line 123: just checking this is supposed to be .token and not .email - assuming that the token corresponds to an emal?

packages/server-core/src/user/strategies/apple.ts
```diff
@@ -118,7 +122,38 @@ export class AppleStrategy extends CustomOAuthStrategy {
...
+      if (entity.type === 'guest' && profile.email) {
+        const profileEmail = profile.email
+        const existingIdentityProviders = await this.app.service(identityProviderPath).find({
+          query: {
+            $or: [
+              {
+                email: profileEmail
+              },
+              {
+                token: profileEmail
```
HexaField: same as above, double checking the token here is supposed to be the same email field

HexaField:
Code looks good, we should probably have a few people test on QAT. cc @hanzlamateen for review too

packages/server-core/src/user/login/login.class.ts
barankyle: Yes. With identity-providers, `token` is the unique identifier for that IP. For email magiclink, `token` is the email address. For an SSO, it's `<sso_provider_name>:::<sso_identifier>`, e.g. `github:::1234abcd`, if your unique GitHub ID were 1234abcd.

The `email` field on identity-providers is used to store the email address associated with an SSO provider, e.g. if I'd created my GH account with `kyle@email.com`, that would be stored in the `email` field of my GH identity-provider, while `token` would be that `github:::1234abcd` value.

packages/server-core/src/user/strategies/apple.ts
barankyle: Yep. In this particular strategy, if you're logging in with Apple SSO, we're checking if the email address associated with your Apple account matches the email associated with any other SSO accounts (identity-provider.email), or any magiclink login with that email (identity-provider.token). As mentioned before, email magiclink identity-providers store the email address on the `token` field.

hanzlamateen (changes requested):
I deployed this branch on ST QAT and getting error when trying to access home page https://ir-engine-qat-dev.theinfinitereality.io/

https://github.com/theinfinitereality/ir-engine-devops/actions/runs/11683825467

image

barankyle:
> I deployed this branch on ST QAT and getting error when trying to access home page https://ir-engine-qat-dev.theinfinitereality.io/
> 
> https://github.com/theinfinitereality/ir-engine-devops/actions/runs/11683825467
> 
> image

This seems to be a problem in dev in general. I'm seeing it happen on the st-dev root page as well, so I don't think it's connected to anything I did.

HexaField (approved):
@barankyle i think that has had enough eyes and testing to warrant us merging it in. we can find any last bugs that come up after it has been merged in. we should definitely include it in the QA matrix

hanzlamateen (changes requested):
I am getting error following using following steps:

1. Signed up using github account.
2. From Dashboard sign out.
3. Signup using email
4. Open magic link
5. Click on yes in link account dialog.

Error video

https://github.com/user-attachments/assets/e95ebaf1-560b-4b9e-ab0b-efbe7bd528c2

barankyle:
I fixed the bug on the email linking. I also updated magiclink emails so that they include both the login-token ID and the token, which should make it harder to brute-force a magiclink login.

hanzlamateen:
@barankyle I used same scenarios as explained in my previous comment. This time, it takes me to /welcome page. Is this desired? I think when I press Yes, it should login as same user as my github account?

https://github.com/user-attachments/assets/e4d77425-993c-47a5-8543-b8dc174b3de8

barankyle:
> @barankyle I used same scenarios as explained in my previous comment. This time, it takes me to /welcome page. Is this desired? I think when I press Yes, it should login as same user as my github account?
> 
>  Recording.2024-11-13.183629.mp4

It was going back to the home page because of a bug that was triggering an error with the login flow. I've fixed that.

hanzlamateen (commented):
@barankyle I tried following 2 scenarios:

I am getting error following using following steps:

Scenario 1:
1. Signed up using github account.
2. From Dashboard sign out.
3. Signup using magiclink email
4. Open magic link
5. Link account dialog appeared.

Scenario 2:
1. Signed up using magiclink email.
2. From Dashboard sign out.
3. Signup using github
4. No link account dialog was shown

I think in scenario 2, step 4, dialog should appear as well?

hanzlamateen:
Converted to draft as GH checks are failing.

barankyle:
That failing test was something unrelated that looks to have been fixed in a recent commit to dev.

I fixed some other bugs I'd found, but I'm not seeing the problem of email signup followed by OAuth signup not offering the link option. Did you delete the account(s) you made in scenario 1 before doing scenario 2?

hanzlamateen:
@barankyle I just retried the scenario 2 and I was unable to see the Link dialog.

As you can see in attached video that in admin panel's user search you can see magic link user. Then I signup using github and no dialog is shown. Then attache screenshot shows the UTC timestamp on which the github user was created. I am in GMT +5 timezone.

https://github.com/user-attachments/assets/b52b1e05-0f3f-43ab-89f3-0be2fd968104

image

barankyle:
I looked at mt-qat-dev and it appeared that this branch hadn't been properly deployed. I `kubectl exec`'d into the API servers, and the github strategy looked like the dev branch - it was missing the email-matching code that's part of this branch. I deleted my identity-providers and also wasn't getting the association prompt after signing up with email and then signing up with GH.

The build status tab was showing that it had failed, with no actual build statuses for the commit of this branch. I think the builder was failing on something before it could even insert a build status. 

I rebased the branch around dev and deployed it successfully. I'm now getting the association prompt after clearing my GH identity provider.

hanzlamateen (approved):

PR #124, opened 2024-08-30
Move PopupMenu to ProjectsPage

CITIZENDOT:
so that pointer-events can be controlled by the parent component

## References
closes #_insert number here_


aditya-mitra (approved):
LGTM

PR #126, opened 2024-08-30
Ir 4294 create prefab fix

JT00y:
## References
https://tsu.atlassian.net/browse/IR-4294


AidanCaruso (approved):
looks good, this fixes the extra assets directory. 

as a note, this change is predicated on exportRelativeGLTF now concatenating the importFolder. if we're keeping that, it would also make sense to remove the editable 'assets/' start to the prefab path input field text.
image

dinomut1 (approved):

dinomut1:
Still seeing the issue with the cache not getting reset when a prefab is overwritten. That is out of scope for this ticket but we should resolve that soon

dinomut1:
https://tsu.atlassian.net/browse/IR-4294

dinomut1:
closing this as reverting gltf export fixes this

PR #139, opened 2024-08-30
IR-3622 fog fixes

SamMazerIR:
## Summary
Fixing fog issues with toggling visibility (would cause webgl errors and make most scene objects invisible)
and adding FogSettingsComponent to the component shelf
## References
[https://tsu.atlassian.net/browse/IR-3622](https://tsu.atlassian.net/browse/IR-3622)

## QA Steps
1. add a fog component to an entity using add component menu, or add a fog prefab through the add prefab menu
2. change fog type to Brownian Motion
3. set fog height and intensity to 1, and set color to something noticeable (this step is simply for visibility)
4. in the hierarchy view, click the eye icon on the right of the fog entity to toggle its visibility
5. observe that nothing in the scene disappears besides the fog effect
6. re-enable its visibility
7. observe fog is being applied again

### DIFF
--- a/packages/spatial/src/renderer/components/FogSettingsComponent.ts
+++ b/packages/spatial/src/renderer/components/FogSettingsComponent.ts
@@ -31,13 +31,16 @@ import {
   getOptionalComponent,
   removeComponent,
   setComponent,
-  useComponent
+  useComponent,
+  useOptionalComponent
 } from '@ir-engine/ecs/src/ComponentFunctions'
 import { useEntityContext } from '@ir-engine/ecs/src/EntityFunctions'
 import { FogComponent } from '@ir-engine/spatial/src/renderer/components/SceneComponents'
 
+import { useHookstate } from '@ir-engine/hyperflux'
 import { FogShaders } from '../FogSystem'
 import { initBrownianMotionFogShader, initHeightFogShader, removeFogShader } from './FogShaders'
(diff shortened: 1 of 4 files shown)


packages/spatial/src/renderer/components/FogSettingsComponent.ts
```diff
@@ -90,6 +93,17 @@ export const FogSettingsComponent = defineComponent({
   reactor: () => {
     const entity = useEntityContext()
     const fog = useComponent(entity, FogSettingsComponent)
+    const isVisible = useOptionalComponent(entity, VisibleComponent)
+    const retriggerFogType = useHookstate(0)
+
+    useEffect(() => {
```
HexaField, line 97: i think rather than the multiple condition based logic we already have and are extending here, we should make the main useEffect with the switch statement have it's effects be reversable, using `return () => {}` to undo whatever the logic is in the switch statements

HexaField (dismissed):

packages/spatial/src/renderer/components/FogSettingsComponent.ts
SamMazerIR: ah I see the way to do this better now, thanks. I finally understood how to fix this after lots of testing and didn't realize the solution could be simplified

SYBIOTE (approved):

PR #141, opened 2024-08-31
Fix model component export path

AidanCaruso:
## References
closes #_insert number here_


AidanCaruso:
Closed - merging this into int instead

PR #142, opened 2024-08-31
IR-4256 Include assets in model component export path

AidanCaruso:
## References
closes #_insert number here_

### DIFF
--- a/packages/ui/src/components/editor/properties/model/index.tsx
+++ b/packages/ui/src/components/editor/properties/model/index.tsx
@@ -108,7 +108,7 @@ export const ModelNodeEditor: EditorComponentType = (props) => {
       return
     }
     exporting.set(true)
-    const fileName = `${srcPath.value}.${exportType.value}`
+    const fileName = `assets/${srcPath.value}.${exportType.value}`
     exportRelativeGLTF(entity, srcProject.value, fileName).then(() => {
       const nuPath = pathJoin(config.client.fileServer, 'projects', srcProject.value, fileName)
       commitProperty(ModelComponent, 'src')(nuPath)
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/editor/properties/model/index.tsx
```diff
@@ -108,7 +108,7 @@ export const ModelNodeEditor: EditorComponentType = (props) => {
       return
     }
     exporting.set(true)
-    const fileName = `${srcPath.value}.${exportType.value}`
+    const fileName = `assets/${srcPath.value}.${exportType.value}`
```
HexaField, line 111: we either need another argument to `exportRelativeGLTF` or to split the filename when passed into the `File` constructor as it shouldn't have the path in it

HexaField (changes requested):

packages/ui/src/components/editor/properties/model/index.tsx
AidanCaruso: Yes. My original solution was to modify exportRelativeGLTF to break this up.

This broke prefab export, which I had forgotten also references exportRelativeGLTF, so it was reverted. Didn't want to make core changes close to release so I mirrored what we're already doing for prefabs and included assets in the relative path argument to make this not error out. I'll look at breaking it up again (in a way that doesn't bork prefabs).

dinomut1 (changes requested):
This is fixed on dev. Instead of this just fix the regex splitting done in the model node editor. See line 87 of this file on dev: https://github.com/ir-engine/ir-engine/blob/dev/packages/ui/src/components/editor/properties/model/index.tsx

packages/ui/src/components/editor/properties/model/index.tsx
dinomut1, line 111: We don't want to hard-code exports into the "assets" folder, to preserve flexibility with folder structure

dinomut1 (changes requested):

PR #145, opened 2024-09-03
refactor: Add LoadingView component to Select dropdown

MoizAdnan:
https://tsu.atlassian.net/browse/IR-2871

### DIFF
--- a/packages/ui/src/primitives/tailwind/Select/index.tsx
+++ b/packages/ui/src/primitives/tailwind/Select/index.tsx
@@ -33,6 +33,7 @@ import { useClickOutside } from '@ir-engine/common/src/utils/useClickOutside'
 import { useHookstate } from '@ir-engine/hyperflux'
 
 import Input from '../Input'
+import LoadingView from '../LoadingView'
 
 export type OptionValueType = string | number
 
@@ -56,6 +57,7 @@ export interface SelectProps<T extends OptionValueType> {
   errorBorder?: boolean
   searchDisabled?: boolean
   inputContainerClassName?: string
+  loading?: boolean
 }
 
 const Select = <T extends OptionValueType>({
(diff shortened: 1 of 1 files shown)


SYBIOTE (approved):

packages/ui/src/primitives/tailwind/Select/index.tsx
```diff
@@ -155,17 +158,21 @@ const Select = <T extends OptionValueType>({
...
-              disabled ? 'opacity-50' : ''
-            }`}
-            onClick={() => {
-              if (!disabled) {
-                toggleDropdown()
-              }
-            }}
-          />
+          loading ? (
+            <LoadingView spinnerOnly className="mr-2 h-6 w-6" />
```
aditya-mitra: `LoadingView` here with a fixed set of props might affect the component's dynamic nature and is tightly coupled. For example, if we need to provide text to the `LoadingView` we will need to add another prop in `SelectProps` which increases prop complexity.

aditya-mitra (changes requested):
Adding the `LoadingView` to `Select` makes the loading part tightly coupled.
We can instead pass an `endComponent` prop from `Select` down to `Input` component

packages/ui/src/primitives/tailwind/Select/index.tsx
```diff
@@ -155,17 +157,21 @@ const Select = <T extends OptionValueType>({
...
-            }`}
-            onClick={() => {
-              if (!disabled) {
-                toggleDropdown()
-              }
-            }}
-          />
+          endComponent ? (
+            endComponent
+          ) : (
```
aditya-mitra: one liner 💡

```suggestion
          endComponent ?? (
```

aditya-mitra (approved):
Looks good ✅

hanzlamateen (approved):

PR #148, opened 2024-09-03
Research on hyperflux build

DanielBelmes:
Current TODO:

- Decide api design(can be pretty fluid)
- Decide external and peer deps


DanielBelmes:
image
current build log snippet

packages/hyperflux/rollup.config.js
```diff
@@ -0,0 +1,80 @@
...
+// });
+
+const productionConfig = {
+  minifyInternalExports: true,
+}
+function customBuild() {
+  return esbuild({minify: true})
+}
+
+const externals = ["react", "react-reconciler", "react-reconciler/constants","ts-matches", "ts-toolbelt", "uuid", "@hookstate/core", "@hookstate/identifiable"]
```
HexaField: wonder if we can import this list from the package.json dependencies object keys

DanielBelmes: This is just to suppress a warning. We don't really need it but I liked having the list.

DanielBelmes: I am working on the vite config right now and for some reason it doesn't auto detect externals so I guess this is needed. Very weird

PR #149, opened 2024-09-03
Added support for cross-domain login.

barankyle:
For multi-domain setups, having to log in on each different domain is annoying. This saves the user's login information in the 'root' domain of the deployment, and on all domains, it sources the login credentials from an iframe running in the root domain using postMessage. This only gets skipped if a user has expressly denied cross-domain sharing of their login information, in which case localStorage will be used as the source for login credentials.

The iframe loads a new non-vite page in client/public, and does so in the root domain. Some user authorization flow is required on most browsers to enable this, using the requestStorageAccess API.

A new backend service, allowed-domains, takes in a domain on a query paramter and returns true if that domain is part of the deployment, and a 204 if not. By default this just returns the root domain, but this is extensible via projects' hooks to add other domains to the allowed list. It can also just be passed a variable isAllowed from a hook if fetching all domains would be cumbersome, and simply querying whether the domain is in a table would be simpler. The accessor iframe uses the response to determine whether to even attempt to access the cookies or prompt the user with requestStorageAccess.

Resolves IR-4430

## References
closes #_insert number here_


barankyle:
This is most easily tested by loading it and the same-named branch of irpro-multitenancy onto mt-qat-dev. When those have been deployed, go to e.g. the studio on the main client URL, then go to an account's subdomain. You should still be logged in as the same user as you were on the main client URL, and that should persist on any other subdomains as well.

Since they're all on the same root domain, this messaging through the iframe should always have access to localStorage and cookies. I'd tested this on minikube with completely different domains, in anticipation of when we'll people use outside domains, and that's when the other domains don't always inherently have access to third-party cookies and we have to request user permission to access them.

hanzlamateen (approved):

PR #150, opened 2024-09-03
refactor(IR-3746): refactor to have a custom radio button component a…

kattsushi:
## Summary
refactor(IR-3746): refactor to have a custom radio button component according the theming and designs

Screenshot 2024-09-03 at 12 14 24 p m

without state default

Screenshot 2024-09-03 at 12 14 18 p m

## References

closes #[IR-3746](https://tsu.atlassian.net/browse/IR-3746) number here_


packages/ui/src/primitives/tailwind/Radio/index.tsx
```diff
@@ -25,82 +25,103 @@ Infinite Reality Engine. All Rights Reserved.
...
-  labelClassName?: string
-  horizontal?: boolean
-  disabled?: boolean
-}) => {
+const Radio = <T extends OptionValueType>(props: RadioGroupProps<T>) => {
+  const { defaultChecked, disabled, name, onChange, options, horizontal, className } = props
+
+  const handleChange = (event: React.FormEvent<HTMLInputElement>) => onChange(event.currentTarget.value as T)
+
+  const defaultName = uuidv4()
```
aditya-mitra, line 109: this will change on every render 
we can just put an empty string in Line#116

packages/ui/src/primitives/tailwind/Radio/index.tsx
```diff
@@ -25,82 +25,103 @@ Infinite Reality Engine. All Rights Reserved.
...
-  value: string | number
-  onChange: React.ChangeEventHandler<HTMLInputElement>
-  selected: boolean
-  className?: string
-  labelClassName?: string
+import cx from 'clsx'
+
+export interface RadioProps {
   disabled?: boolean
+  defaultChecked?: boolean
```
aditya-mitra: the Radio input is always dependent on a parent component
so we can skip using `defaultChecked`
React recommend either using default value or value
we instead specify `value` to be mandatory

this is creating some situations like the one below: although the value is provided, the input is unchecked

image

packages/ui/src/primitives/tailwind/Radio/index.tsx
```diff
@@ -25,82 +25,103 @@ Infinite Reality Engine. All Rights Reserved.
...
+  <div className="flex flex-col">
+    <div className="flex items-start gap-2">
+      <div className="mt-1 grid place-items-center">
         <input
           type="radio"
-          checked={selected}
-          value={value}
-          name={label}
-          onChange={onChange}
+          id={id}
```
aditya-mitra: react provides a new [`useId`](https://react.dev/reference/react/useId) method which can be used for specifying unique ids. lets use that

packages/ui/src/primitives/tailwind/Radio/index.tsx
```diff
@@ -25,82 +25,103 @@ Infinite Reality Engine. All Rights Reserved.
...
-          labelClassName={labelClassName}
-          value={optionValue}
-          onChange={(event) => onChange(event.target.value as T)}
-          disabled={disabled}
-          description={description}
-        />
+      {options.map(({ label: optionLabel, value, description }, index) => (
+        <div key={value} className="flex items-center gap-2">
+          <RadioRoot
+            id={value}
```
aditya-mitra: let `useId` be specified on `RadioRoot` instead of here

aditya-mitra (changes requested):

packages/ui/src/primitives/tailwind/Radio/index.tsx
aditya-mitra: this will change of every render
we can just specify an empty string on Line#116

aditya-mitra (dismissed):
Awesome 🚀
This now matches closely with the design.

packages/ui/src/primitives/tailwind/Radio/index.tsx
kattsushi: Hey @aditya-mitra, thanks for your comments! I think I really don't want to change the API of the component (making value required) out of respect for the refactoring process, to avoid breaking anything. But I saw that the issue is already fixed you need to pass the value to pre-select a option of the radio button

packages/ui/src/primitives/tailwind/Radio/index.tsx
kattsushi: good call fixed!

packages/ui/src/primitives/tailwind/Radio/index.tsx
kattsushi: fixed! thanks! now looks good!

mikeplascdev (approved):

packages/ui/src/primitives/tailwind/Radio/index.tsx
```diff
@@ -23,84 +23,109 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
-  disabled,
-  description
-}: {
-  label: string
-  value: string | number
-  onChange: React.ChangeEventHandler<HTMLInputElement>
-  selected: boolean
-  className?: string
-  labelClassName?: string
+import cx from 'clsx'
```
CITIZENDOT, line 29: `clsx` is used in combination with material ui. `tailwind-merge` replaces this. Let's use `twMerge` instead.

packages/ui/src/primitives/tailwind/Radio/index.tsx
```diff
@@ -23,84 +23,109 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
+            defaultChecked={defaultChecked}
+            value={value}
+          />
+          <div
+            className={cx(
+              'pointer-events-none',
+              'col-start-1 row-start-1',
+              'h-2 w-2 rounded-full peer-checked:bg-blue-500',
+              'peer-checked:peer-disabled:bg-gray-400'
+            )}
```
CITIZENDOT, line 70: replace with `twMerge`

packages/ui/src/primitives/tailwind/Radio/index.tsx
```diff
@@ -23,84 +23,109 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
+          />
+        </div>
+        <label
+          htmlFor={radioId}
+          className={cx('text-start hover:cursor-pointer', {
+            'text-gray-400': disabled
+          })}
```
CITIZENDOT, line 77: replace with `twMerge`

CITIZENDOT (changes requested):

packages/ui/src/primitives/tailwind/Radio/index.tsx
CITIZENDOT: as per our discussion, we'll be using `clsx` along side with `tailwind-merge`. It's a small package compared with `tailwind-merge` (8.55KB vs 709KB) anyway.

CITIZENDOT (approved):

aditya-mitra (approved):
Works good ✅

PR #152, opened 2024-09-03
fix list view panel name alignment 

SYBIOTE:
## Summary
fix list view panel name alignment 
the fallback icon styling was causign problems, fixed 

## References
closes #_insert number here_


SYBIOTE:
current branch made from int, deleting this

PR #157, opened 2024-09-04
Fix gizmo rotation update

SYBIOTE:
This pull request fixes the issue with the gizmo rotation update. The `EulerInput` component was not correctly updating the quaternion and euler values, causing incorrect rotation behavior. This PR ensures that the quaternion and euler values are properly updated when changes are made to the component's input.


packages/ui/src/components/editor/input/Euler/index.tsx
```diff
@@ -43,66 +43,71 @@ type EulerInputProps = {
   unit?: string
 }
 
+const temp_Euler = new Euler()
+const temp_Quat = new Quaternion()
```
aditya-mitra: we should only use camelCase in ts files

```suggestion
const tempEuler = new Euler()
const tempQuat = new Quaternion()
```

packages/ui/src/components/editor/input/Euler/index.tsx
```diff
@@ -43,66 +43,71 @@ type EulerInputProps = {
...
+
 /**
  * FileIEulerInputnput used to show EulerInput.
  *
  * @type {Object}
```
aditya-mitra: we can remove this jsdoc type

packages/ui/src/components/editor/input/Euler/index.tsx
```diff
@@ -43,66 +43,71 @@ type EulerInputProps = {
...
   const quaternion = useHookstate(props.quaternion)
-  const euler = useHookstate(new Euler().setFromQuaternion(quaternion.value, 'YXZ'))
+  const euler = useHookstate(temp_Euler.setFromQuaternion(props.quaternion, 'YXZ'))
 
   useEffect(() => {
-    euler.value.setFromQuaternion(quaternion.value, 'YXZ')
+    euler.set(temp_Euler.setFromQuaternion(quaternion.value, 'YXZ'))
   }, [props.quaternion])
 
   const onSetEuler = useCallback(
```
aditya-mitra: we can remove the `useCallback` here since it will be memoized by react compiler

packages/ui/src/components/editor/input/Euler/index.tsx
```diff
@@ -43,66 +43,71 @@ type EulerInputProps = {
...
  * @type {Object}
  */
 export const EulerInput = (props: EulerInputProps) => {
   const quaternion = useHookstate(props.quaternion)
-  const euler = useHookstate(new Euler().setFromQuaternion(quaternion.value, 'YXZ'))
+  const euler = useHookstate(temp_Euler.setFromQuaternion(props.quaternion, 'YXZ'))
 
   useEffect(() => {
-    euler.value.setFromQuaternion(quaternion.value, 'YXZ')
+    euler.set(temp_Euler.setFromQuaternion(quaternion.value, 'YXZ'))
```
aditya-mitra: `tempEuler` persists the changes. We can use `new Euler().<methodName>` without needing tempEuler (or tempQuartenion)

```suggestion
    euler.set(new Euler().setFromQuaternion(quaternion.value, 'YXZ'))
```

aditya-mitra (dismissed):
we might not need the temp variables

packages/ui/src/components/editor/input/Euler/index.tsx
SYBIOTE: using a temp euler to not have to create a new one every time EulerInput gets rerendered

packages/ui/src/components/editor/input/Euler/index.tsx
SYBIOTE: done

packages/ui/src/components/editor/input/Euler/index.tsx
SYBIOTE: done

SYBIOTE:
> we might not need the temp variables

i prefer keeping them, creating a new euler or quat every time the component gets rerendered is wasteful

packages/ui/src/components/editor/input/Euler/index.tsx
SYBIOTE: tempEuler will be overritten before the euler value gets set

HexaField:
> > we might not need the temp variables
> 
> i prefer keeping them, creating a new euler or quat every time the component gets rerendered is wasteful

For uniqueness reasons I would actually prefer new objects each time. It keeps it simpler to reason about and the performance hit is negligible for just one value.

SYBIOTE:
> > > we might not need the temp variables
> > 
> > 
> > i prefer keeping them, creating a new euler or quat every time the component gets rerendered is wasteful
> 
> For uniqueness reasons I would actually prefer new objects each time. It keeps it simpler to reason about and the performance hit is negligible for just one value.

ok, ill revert the change

HexaField (approved):

PR #159, opened 2024-09-04
fix parent transform rigidbody scaling

HexaField:
## References
closes #_insert number here_


DanielBelmes:
Closed to be based off int

PR #165, opened 2024-09-04
Stop media on MediaElementComponent removal

MichaelEstes:
## References
closes #_insert number here_


DanielBelmes:
Closed as it's a part of v0.8 release and needs to be int based

PR #169, opened 2024-09-04
V0.8 studio: asset panel render by height

achen5671:
## References
closes #_insert number here_

### DIFF
--- a/packages/ui/src/components/editor/panels/Assets/container/index.tsx
+++ b/packages/ui/src/components/editor/panels/Assets/container/index.tsx
@@ -569,7 +569,10 @@ const AssetPanel = () => {
     }
   }
 
-  useEffect(() => staticResourcesPagination.skip.set(0), [searchText])
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


aditya-mitra:
we should always base off from `dev` branch
since commits are cherry-picked into `int` branch

achen5671:
> we should always base off from `dev` branch since commits are cherry-picked into `int` branch

I was told we should branch of int right now, cc: @DanielBelmes

DanielBelmes:
Yup sorry I was hoping it would get to you through Rahul and Josh. @aditya-mitra all fixes for V0.8 will occur in int and be brought over to dev now. We found it's a more consistent release path than what we were doing earlier.

achen5671:
@DanielBelmes @aditya-mitra after discussions, looks like we are going to move it so it branches off dev. Let me know if there are any objection

achen5671:
closing infavor of https://github.com/ir-engine/ir-engine/pull/183

PR #181, opened 2024-09-05
IR-4343 Ambient light culling issue

SamMazerIR:
## Summary
adding useEffect along with useOptionalComponent to remove FrustumCullCameraComponent from AmbientLightComponent when it exists. Ambient lights should not be culled based on being within the frustum

## References
[https://tsu.atlassian.net/browse/IR-4343](https://tsu.atlassian.net/browse/IR-4343)

## QA Steps
Create a new scene

Add Ambient Light to the scene

Change the Color to Red

Change intensity from 1 to 10

Change the ambient lighting position to 100,100,100

Observe lighting effect still active in scene


HexaField (dismissed):
Hmm, while this change may be find, I'd prefer to change whatever logic it is that is adding it in the first place. It seems like it is `generateEntityJsonFromObject` as a result of loading with the ModelComponent, which is odd for a scene entity. Can you look into where this is coming from? Might be as simple as adding an `if (object.isMesh)` condition to the `setComponent(entity, FrustumCullCameraComponent` call

SamMazerIR:
> Hmm, while this change may be find, I'd prefer to change whatever logic it is that is adding it in the first place. It seems like it is `generateEntityJsonFromObject` as a result of loading with the ModelComponent, which is odd for a scene entity. Can you look into where this is coming from? Might be as simple as adding an `if (object.isMesh)` condition to the `setComponent(entity, FrustumCullCameraComponent` call

done

HexaField (approved):

PR #185, opened 2024-09-06
Update customer Service help button design

SYBIOTE:
## References
closes #_insert number here_


packages/editor/src/components/EditorContainer.tsx
```diff
@@ -246,17 +247,26 @@ const EditorContainer = () => {
         })}
       </div>
       <PopupMenu />
-      {!isWidgetVisible && initialized && (
```
aditya-mitra, line 249: any reason why we removed these conditions?

aditya-mitra (dismissed):

packages/editor/src/components/EditorContainer.tsx
jose-galvan: removing `isWidgetVisible` is causing an issue where both the widget and the button are visible simultaneously. When the widget is displayed, the button should be hidden to avoid overlap

image


`initialized` flag ensures that the widget is fully initialized and the chat can be opened. If removed, the button will become unresponsive when the Zendesk script hasn’t fully loaded or if the Zendesk configuration is missing, incomplete, or disabled.

SYBIOTE: i removed it for testing, im sure i reverted that, ill add it back if its not in there

SYBIOTE: updated

jonathan-casarrubias (approved):

jose-galvan (approved):

PR #188, opened 2024-09-06
[IR-3423] Unit Test - Spatial Physics coverage to 80%

heysokam:
## References
https://tsu.atlassian.net/browse/IR-3423


packages/spatial/src/physics/systems/PhysicsPreTransformSystem.test.ts
```diff
@@ -0,0 +1,569 @@
...
+  function assertNotDirty(entity: Entity, id: number = 0): void {
+    if (!id) {
+      assert.equal(TransformComponent.dirtyTransforms[entity], false)
+      return
+    }
+    if (TransformComponent.dirtyTransforms[entity] === undefined) return
+    assert.equal(TransformComponent.dirtyTransforms[entity], false)
+  }
+
+  describe('setDirty', () => {
```
HexaField: A setDirty test is unnecessary, as the function definition exists only to minimize garbage. We only need to test functionality unique to the module.

packages/spatial/src/physics/systems/PhysicsPreTransformSystem.ts
```diff
@@ -195,3 +195,11 @@ export const PhysicsPreTransformSystem = defineSystem({
   insert: { after: TransformDirtyUpdateSystem },
   execute
 })
+
+export const _PhysicsPreTransformFunctions = {
```
HexaField: Let's remove the underscore from this definition name. Even though these functions are unlikely to be used externally, we should keep the namespace clear.

HexaField (changes requested):

packages/spatial/src/physics/systems/PhysicsPreTransformSystem.ts
heysokam: Fixed with 89c3e14

packages/spatial/src/physics/systems/PhysicsPreTransformSystem.test.ts
heysokam: Fixed with 89c3e14

HexaField (approved):

PR #202, opened 2024-09-09
pin `three-mesh-bvh` version

CITIZENDOT:
## Summary

Pins `three-mesh-bvh` version because it's breaking tests.

## References
closes #_insert number here_


AidanCaruso:
We'll have to fix the tests with later versions of three-mesh-bvh at some point as the gltf loader PR bumps three-mesh-bvh to the latest version for improvements to indirect bvh support. Which tests is this version breaking?

CITIZENDOT:
[This PR's](https://github.com/theinfinitereality/irpro-multitenancy/pull/475) tests are failing due to the `three-mesh-bvh` latest version issue. Can we merge this until the gltf loader PR is ready?

Test link: https://github.com/theinfinitereality/irpro-multitenancy/actions/runs/10772137926/job/29869577765?pr=475

AidanCaruso (changes requested):
We should pin this at the last working version, 0.7.1 is 6 releases behind and lacks some performance improvements (and indirect mode) present in later versions. Does version 0.7.6 (which is one behind current) present the same errors? It's also really weird that the error only happens on one project.

CITIZENDOT:
updated version to `0.7.6`.

Also, I noticed when `@ir-engine/instanceserver` imports are removed in the mentioned PR, this error won't occur. So, I suspect something in the `@ir-engine/instanceserver` is importing `three-mesh-bvh` which is causing the error.

AidanCaruso (approved):
Looks good, testing with 0.7.6 works

PR #205, opened 2024-09-09
Fix hierarchy panel reordering

SYBIOTE:
This pull request fixes an issue with the hierarchy panel reordering. The `reparentObject` function has been updated to include an additional parameter `after` which allows for reordering of entities in the hierarchy. The `dropItem` function in the `HierarchyTreeNode` component has also been modified to handle placing entities before or after a specific node. These changes ensure that the hierarchy panel reordering functionality works correctly.

temporary fix for now, i propose a deeper refactor

### DIFF
--- a/packages/editor/src/functions/EditorControlFunctions.test.tsx
+++ b/packages/editor/src/functions/EditorControlFunctions.test.tsx
@@ -669,11 +669,12 @@ describe('EditorControlFunctions', () => {
       const node2Entity = UUIDComponent.getEntityByUUID(node2UUID)
       const sourceID = getComponent(nodeEntity, SourceComponent)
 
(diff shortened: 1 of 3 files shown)


aditya-mitra (commented):
I am doing a full refactoring of the hierarchy panel which should fix this issue.
I think that we introduction of the `after` argument in `EditorControlsFunctions.reparentObject` can be avoided.

SYBIOTE:
> I am doing a full refactoring of the hierarchy panel which should fix this issue. I think that we introduction of the `after` argument in `EditorControlsFunctions.reparentObject` can be avoided.

thats necessary, we can put this in as a temp fix for now while you work on the same

dinomut1 (approved):

PR #208, opened 2024-09-10
fix: share instance by invite

jose-galvan:
## Summary
This update resolves the issue with sharing an instance via invite. The problem stemmed from the incorrect format of the `inviteCode` in the schema, and the query for retrieving existing invites included unnecessary properties, causing the request to fail.

## References
closes #_insert number here_

## QA Steps
- Go to an existing location
- Click on the share button 
- Input an email address 
- Click on send 
It should send an invite to the recipient. 
image


aditya-mitra (approved):
Looks good ✅

PR #218, opened 2024-09-11
Fixed verification of GitHub JWT signing tokens.

barankyle:
## Summary

Verification of GH JWTs was mistakenly using the apps' secret. A GitHub private key can now be stored in
authentication-settings.oauth.github.privateKey, and the GH JWT auth gets the key from there.

Fixed some bugs with populating JWT config in updateAppConfig.ts

Fixed a naming bug when downloading a project without an org in the name.

## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/projects/project/project.hooks.ts
+++ b/packages/server-core/src/projects/project/project.hooks.ts
@@ -259,14 +259,14 @@ const checkIfProjectExists = async (context: HookContext<ProjectService>) => {
   }
 
   const data: ProjectData[] = Array.isArray(context.data) ? context.data : [context.data]
-
   const projectName = data[0].name!
+  const slashIndex = projectName.indexOf('/')
 
-  const orgName = projectName.slice(0, projectName.indexOf('/'))
-
-  const cleanedProjectName = cleanString(projectName.slice(projectName.indexOf('/')))
-
-  context.projectName = `${orgName}/${cleanedProjectName}`.toLowerCase()
+  if (slashIndex > -1) {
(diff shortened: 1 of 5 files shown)


packages/server-core/src/projects/project/project.hooks.ts
```diff
@@ -259,14 +259,14 @@ const checkIfProjectExists = async (context: HookContext<ProjectService>) => {
...
-  const orgName = projectName.slice(0, projectName.indexOf('/'))
-
-  const cleanedProjectName = cleanString(projectName.slice(projectName.indexOf('/')))
-
-  context.projectName = `${orgName}/${cleanedProjectName}`.toLowerCase()
+  if (slashIndex > -1) {
+    const orgName = projectName.slice(0, projectName.indexOf('/'))
+    const cleanedProjectName = cleanString(projectName.slice(projectName.indexOf('/')))
+    context.projectName = `${orgName}/${cleanedProjectName}`.toLowerCase()
+  } else context.projectName = projectName.toLowerCase()
```
hanzlamateen, line 269: In what scenario this else block will be used?

hanzlamateen (approved):

packages/server-core/src/projects/project/project.hooks.ts
barankyle: With multitenancy, we're enforcing an [org]/[projectName] naming convention. But in a single-tenant environment, it should be viable to not use org names. When I was testing, I re-used a saved query that was sending the project name without an org name, and it was turning e.g. irpro-stripe into irpro-strip/e.

PR #219, opened 2024-09-11
IR-4344 uploaded files extension tolower

SamMazerIR:
## Summary
adding functionality to return a new File object with tolower on its extension, calls to this for upload files button and drag and drop files

## References
[https://tsu.atlassian.net/browse/IR-4344](https://tsu.atlassian.net/browse/IR-4344)

## QA Steps
Open dev env

Create a new scene

Upload mp3 file that has file extension of all caps ".MP3" (any supported file type works, just make it all caps as that was the root issue being fixed)

Observe that the uploaded file now displays with ".mp3" in lowercase (or whatever extension it should be)

Drag uploaded mp3 file from files tab to scene (again, any relevant file type, as long as it can be dragged into viewport - e.g. mp3, mp4...)

Observe the app behavior working as intended


MichaelEstes (approved):

aditya-mitra:
We should combine this with https://github.com/ir-engine/ir-engine/pull/257/files as both of them are modifying files in bulk. A common function such as `cleanFileName(file:File):File` would be helpful

HexaField:
Furthermore, we should include this logic in the backend file-browser and/or static-resource services so we can guarantee we are standardising everywhere

SamMazerIR:
merged this PR for the immediate bugfix, but can you make a ticket with details of the work @HexaField or @aditya-mitra ?

lets be explicit which functions to roll into "cleanFileName(file:FIle):File" and where it should live, then for Josh where would that need to be implemented specifically?

HexaField:
> merged this PR for the immediate bugfix, but can you make a ticket with details of the work @HexaField or @aditya-mitra ?
> 
> lets be explicit which functions to roll into "cleanFileName(file:FIle):File" and where it should live, then for Josh where would that need to be implemented specifically?

https://tsu.atlassian.net/browse/IR-4766

PR #228, opened 2024-09-12
[IR-4553] Unit Test - Spatial: Physics Integration Tests

heysokam:
## References
https://tsu.atlassian.net/browse/IR-4553


heysokam:
@HexaField I think this covers all of the test cases we talked about, but let me know if it is missing any.
Otherwise it should be ready to go.

HexaField (approved):

PR #229, opened 2024-09-12
studio: warning for unsupported browser

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-4471
## References
closes #_insert number here_


packages/editor/src/pages/EditorPage.tsx
```diff
@@ -76,5 +84,33 @@ export const EditorPage = () => {
...
+        PopoverState.showPopupover(
+          <Modal
+            onSubmit={() => true}
+            onClose={() => PopoverState.hidePopupover()}
+            className="w-[50vw] max-w-2xl"
+            hideFooter
+          >
+            <div className="flex flex-col gap-2">
+              <span className="flex items-center gap-2">
+                <FcInfo /> Browser Not Supported!
```
aditya-mitra: use translation strings

packages/editor/src/hooks/useBrowserCheck.ts
```diff
@@ -0,0 +1,22 @@
...
+      if (!isChrome && !isSafari) {
+        setUnsupported(true)
+      }
+    }
+
+    detectBrowser()
+  }, [])
+
+  return unsupported
+}
```
aditya-mitra: this can be transformed into a normal function instead of a hook
because this returns the output only when the component mounts which is the same as a normal return from a function

aditya-mitra (changes requested):

aditya-mitra (approved):
LGTM

(suggestion: for changes to the ui, adding few screenshots will make the review process faster)

achen5671:
> LGTM
> 
> (suggestion: for changes to the ui, adding few screenshots will make the review process faster)

good idea! will do :)

PR #231, opened 2024-09-12
chore(ir-engine): implemented location find service params hook

jonathan-casarrubias:
## Summary
This hook function ensure that location find service can't be executed with empty query params object, followed by a validation that only allowed params can be included in the object.
## Breaking Changes
N/A
## References
closes #_insert number here_
https://tsu.atlassian.net/browse/IR-3891


packages/server-core/src/social/location/location.hooks.ts
```diff
@@ -222,14 +223,48 @@ const duplicateNameError = async (context: HookContext<LocationService>) => {
...
+    '$limit',
+    '$sort',
+    'sceneId',
+    'projectId',
+    'name',
+    'id'
+  ]
+  const keys: string[] = Object.keys(context.params.query || {})
+  if (!keys.length) throw new BadRequest(`Query can't have empty params object`)
+  if (!keys.map((key) => allowed.includes(key)).reduce((p, c) => (!p || !c ? false : true)))
```
jose-galvan: `schemaHooks.validateQuery(locationQueryValidator)` already makes sure the query only includes parameters from `locationQuerySchema` and throws an error if there's anything extra.

Screenshot 2024-09-12 at 4 50 14 p m

jose-galvan (changes requested):

packages/server-core/src/social/location/location.hooks.ts
jonathan-casarrubias: @hanzlamateen I've reviewed what @jose-galvan is mentioning by analyzing what `schemaHooks.validateQuery(locationQueryValidator)` hook is doing under the hood and my understand is that he is right, unless we are missing something it seems that validating the parameters again if action != admin is redundant.

I introduced my implementation based off the Story conversation we had in Jira and during the Stand Up call, but I totally missed what Jose found and is mentioning. 

Regardless, since you explained the requirements I would like to know what your opinion is.

I still believe that if isAction('admin') then we should keep verifyScope('location', 'read'), there's no discussion on that one, but....

!isAction('admin') then... Do we really need to verify the query params to ensure only params being used currently are allowed? IMO potentially it's redundant, the only benefit I can think of, is that locationQueryValidator allows for all parameters defined in the schema, while within my hook I'm addressing your requirement that only parameters that are currently being sent from every location call are allowed, for example my validation would reject the call if `maxUsersPerInstance`, `isFeatured` or `isLobby` are part of the query, the reason is that not a single existing Location.find call is using those parameter, while the schema validator would allow access, because these properties are part of the schema, that's the only difference I see.

There still is 1 more requirement to be discussed and that is Not Allowing empty query params when calling Location.find service.

So, I believe the end result should be something like

```tsx
iff(
  isProvider('external'),
  iffElse(isAction('admin'), [verifyScope('location', 'read')], [verifyEmptyQueryParams])
)
```
thoughts?

jonathan-casarrubias: A couple thoughts more,

First thought is that a hook to validate that the query params object isn't empty is super generic, I won't be surprised if a hook making that validation already exists, if it does exist already I would appreciate if you mention which hook is, so I can simply leverage it instead of creating a new one for the same purpose.

Second thought is that it would be super helpful to have a documentation listing all the existing hooks, it would save a lot of time when working on stories like this to first check the documentation of available hooks in order to use those already existing and only attempt to create new hooks if the functionality isn't listed in the documentation, so if the documentation already exist I would appreciate if you can provide it, if it doesn't exist it would be a great idea to coordinate with Armando to create it.


Thanks

hanzlamateen: So you mentioned `!isAction('admin')` in 1st comment. Its important to notice we have an overloaded scenario in MT that introduces `action: 'account'`.

https://github.com/theinfinitereality/irpro-multitenancy/blob/b1c7af5e227663c689e06f0268a9f5c27fd25ac7/services/hooks/social/location.hooks.ts#L76

We dont have an existing hook that checks for empty nor there is a documentation. Maybe you relevant one here: https://hooks-common.feathersjs.com/hooks.html

We can leverage schema validator and we should get rid of the properties that are not used. Like you mentioned (maxUsersPerInstance, isFeatured or isLobby). You can search for Type.Union and Type.Intersection in our code base to see how we can have multiple combinations.

Though its important to mention we also want to restrict some scenarios like following to be done from backend only or from admin. Here %like allows searching through api, we dont want any user to search locations without enough scopes.

https://github.com/ir-engine/ir-engine/blob/a35c1b9355c056b8b58c07a18ffe6492e5a45a71/packages/server-core/src/projects/project/project.hooks.ts#L455
https://github.com/ir-engine/ir-engine/blob/a35c1b9355c056b8b58c07a18ffe6492e5a45a71/packages/client-core/src/user/components/UserMenu/menus/LocationMenu.tsx#L80

I think `LocationMenu` is not being used anywhere so it might be redundant.

hanzlamateen: @jonathan-casarrubias FYI

kattsushi (approved):

packages/server-core/src/hooks/is-action.ts
```diff
@@ -28,6 +28,9 @@ import { HookContext } from '../../declarations'
 export default (...params: string[]) => {
   const args = Array.from(params)
   return (hook: HookContext): boolean => {
-    return hook.params?.query && args.includes(hook.params.query.action)
+    return (
+      (hook.params?.query && args.includes(hook.params.query.action)) ||
+      (hook.params?.actualQuery && args.includes(hook.params?.actualQuery?.action))
```
hanzlamateen:
```suggestion
      (hook.params?.actualQuery && args.includes(hook.params.actualQuery.action))
```

packages/server-core/src/social/location/location.hooks.ts
```diff
@@ -222,14 +224,74 @@ const duplicateNameError = async (context: HookContext<LocationService>) => {
...
     all: [schemaHooks.validateQuery(locationQueryValidator), schemaHooks.resolveQuery(locationQueryResolver)],
-    find: [discardQuery('action'), discardQuery('studio'), sortByLocationSetting],
+    find: [
+      // Empty query object in findLocation is not allowed
+      rejectEmptyQuery,
+      // Persist query to avoid hook collitions or malformed knex query instances
+      persistQuery,
+      // Clean up query of non db relevant parameters
+      discardQuery('action'),
+      discardQuery('studio'),
```
hanzlamateen:
```suggestion
```
I think we can remove this `discardQuery('studio'),` as I dont see it used anywhere in code.

packages/server-core/src/social/location/location.hooks.ts
```diff
@@ -222,14 +224,74 @@ const duplicateNameError = async (context: HookContext<LocationService>) => {
...
+      // On external calls. evaluate action property presence and restrict queries according
+      iff(
+        isProvider('external'),
+        iffElse(hasAction, [verifyScope('location', 'read'), restrictProtectedQuery], [restrictPublicQuery])
```
hanzlamateen: So it seems like hasAction is running `verifyScope('location', 'read')` for all actions which is wrong.

`verifyScope('location', 'read')` should work only if user is an admin. If a user has `location:read` then he can read all locations in our system. Even those that belong to another user.

For end users using studio and dashboard, they only have `editor:write` scope. So `verifyScope('location', 'read')` should only be performed for `action: admin` requests and not other actions.

Right now, if you were testing this flow with account created through seed data locally, then that user has all admin scopes as well as a MT account. For testing this scenarios. Try to open an incognito window and do a new user signup. Then try accessing locations from studio and dashboard. Because this user will only have `editor:write` scope and truly represents an end user.

hanzlamateen (changes requested):

packages/server-core/src/social/location/location.hooks.ts
jonathan-casarrubias: @hanzlamateen I've always used the sign up section to create users, don't know how to create seeded users, regardless I cleaned up my environment `npm run dev-reinit`, I ran the project, using magic link I created a new user from Sign Up:

image
(See ID: 43b3fd8c-3f89-4c39-870b-6909d62df8af)

Then if I go and see the scopes associated with this user

image

Am I missing something? I just want to understand because you said this user will only have editor:write.

Aside of that I did test the different locations related sections (admin/location, dashboard/location, studio/publish location "viewer") and everything was working as expected before I sending the PR last night.

Anyway I still made the change you requested in case that I'm missing something

```ts
iff(isAction('admin'), verifyScope('location', 'read')).else(verifyScope('editor', 'write'))
```
Which after my testes and as I assumed it would work exactly the same because of this:

image

In deed it worked just the same before the changes, so I don't mind pushing the changes, I just want to understand the difference because I don't see any since both scopes `location:read` and `editor:write` are associated with the user I just created and not just `editor:write` as you said it should be.

jonathan-casarrubias:
@hanzlamateen this PR and https://github.com/theinfinitereality/irpro-multitenancy/pull/501 have to be merged at the same time or MT Locations will get broken until both are in dev

mikeplascdev (approved):

packages/server-core/src/social/location/location.hooks.ts
hanzlamateen: So, after dev-reinit. The first user is automatically made an admin. I hope you have your .env.local setup as mentioned in readme below.

https://github.com/theinfinitereality/irpro-multitenancy/blob/dev/readme.md
image

The screenshot you showed had 1st user getting all the scope, which does not depict a typical end user. So once you have this user.

Open an incognito window and there signup using a different SSO/Email. Now this user will be 2nd user in the system and it should only have editor:write scope and it depicts a real end user using our platform.

packages/server-core/src/social/location/location.hooks.ts
```diff
@@ -222,14 +225,80 @@ const duplicateNameError = async (context: HookContext<LocationService>) => {
...
+      persistQuery,
+      // Clean up query of non db relevant parameters
+      discardQuery('action'),
+      // On external calls. evaluate action property presence and restrict queries according
+      iff(
+        isProvider('external'),
+        iffElse(
+          hasAction,
+          [
+            iff(isAction('admin'), verifyScope('location', 'read')).else(verifyScope('editor', 'write')),
```
hanzlamateen, line 294:
```suggestion
            iffElse(isAction('admin'), verifyScope('location', 'read'), verifyScope('editor', 'write')),
```

Just a humble request, since we use iffElse all across our codebase instead of iff().else(). Can you make this change.

hanzlamateen (changes requested):

packages/server-core/src/social/location/location.hooks.ts
jonathan-casarrubias: @hanzlamateen not sure why, but when iffElse is deep nested, intellisense is complaining while the exact same implementation at a top level isn't complaining, see the following:

image

that's not even the reason I decided to use the iff().else() approach, I just believe that it is unnecessary to allocate in memory 2 arrays object instances (and everything that is included within its prototype) while each array object is only containing a single hookFunction, but I don't mind to address your change request it's just that for some reason intellisense is complaining even though these are supposed to be equivalent implementations.

Oddly enough, according to the documentation it supposedly should work if using conditional hooks as part of the Array<hooksFuncTrue> collection:

image


thoughts?

hanzlamateen (changes requested):
I signed up a fresh account in new tab, unable to publish location from a new scene.

> Make sure to try with a non admin account as I did

https://github.com/user-attachments/assets/f05116dc-f06a-405e-b8c8-4998890a7ff1

```
[server] [server] [server-core:log-error] Error in location service, before hook, find method. Error: Unauthorized read action on location, stacktrace: Forbidden: Unauthorized read action on location
[server] [server]     at /home/hanzlamateen/irengine/packages/server-core/src/hooks/verify-scope.ts:54:18
[server] [server]     at processTicksAndRejections (node:internal/process/task_queues:95:5)
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:47:26
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:51:7
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:47:26
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:46:27
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:51:7
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:47:26
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:51:7
[server] [server]     at async LocationService.<anonymous> (/home/hanzlamateen/irengine/node_modules/@feathersjs/schema/src/hooks/resolve.ts:97:5)
[server] [server]     at async LocationService.<anonymous> (/home/hanzlamateen/irengine/node_modules/@feathersjs/schema/src/hooks/resolve.ts:97:5)
[server] [server]     at async LocationService.<anonymous> (/home/hanzlamateen/irengine/node_modules/@feathersjs/schema/src/hooks/resolve.ts:150:5)
[server] [server]     at async LocationService.logError (/home/hanzlamateen/irengine/packages/server-core/src/hooks/log-error.ts:42:5)
[server] [server]     at async runMethod (/home/hanzlamateen/irengine/node_modules/@feathersjs/transport-commons/src/socket/utils.ts:118:38) Forbidden: Unauthorized read action on location
[server] [server]     at /home/hanzlamateen/irengine/packages/server-core/src/hooks/verify-scope.ts:54:18
[server] [server]     at processTicksAndRejections (node:internal/process/task_queues:95:5)
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:47:26
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:51:7
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:47:26
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:46:27
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:51:7
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:47:26
[server] [server]     at async /home/hanzlamateen/irengine/node_modules/feathers-hooks-common/dist/index.js:51:7
[server] [server]     at async LocationService.<anonymous> (/home/hanzlamateen/irengine/node_modules/@feathersjs/schema/src/hooks/resolve.ts:97:5)
[server] [server]     at async LocationService.<anonymous> (/home/hanzlamateen/irengine/node_modules/@feathersjs/schema/src/hooks/resolve.ts:97:5)
[server] [server]     at async LocationService.<anonymous> (/home/hanzlamateen/irengine/node_modules/@feathersjs/schema/src/hooks/resolve.ts:150:5)
[server] [server]     at async LocationService.logError (/home/hanzlamateen/irengine/packages/server-core/src/hooks/log-error.ts:42:5)
[server] [server]     at async runMethod (/home/hanzlamateen/irengine/node_modules/@feathersjs/transport-commons/src/socket/utils.ts:118:38) {
[server] [server]   code: 403,
[server] [server]   className: 'forbidden',
[server] [server]   type: 'FeathersError',
[server] [server]   hook: ContextClass {
[server] [server]     self: LocationService {
[server] [server]       update: [Function],
[server] [server]       find: [Function],
[server] [server]       get: [Function],
[server] [server]       create: [Function],
[server] [server]       patch: [Function],
[server] [server]       remove: [Function],
[server] [server]       hooks: [Function (anonymous)],
[server] [server]       _events: [Object: null prototype],
[server] [server]       _eventsCount: 4,
[server] [server]       _maxListeners: undefined,
[server] [server]       setMaxListeners: [Function: setMaxListeners],
[server] [server]       getMaxListeners: [Function: getMaxListeners],
[server] [server]       emit: [Function (anonymous)],
[server] [server]       addListener: [Function: addListener],
[server] [server]       on: [Function: addListener],
[server] [server]       prependListener: [Function: prependListener],
[server] [server]       once: [Function: once],
[server] [server]       prependOnceListener: [Function: prependOnceListener],
[server] [server]       removeListener: [Function: removeListener],
[server] [server]       off: [Function: removeListener],
[server] [server]       removeAllListeners: [Function: removeAllListeners],
[server] [server]       listeners: [Function: listeners],
[server] [server]       rawListeners: [Function: rawListeners],
[server] [server]       listenerCount: [Function: listenerCount],
[server] [server]       eventNames: [Function: eventNames],
[server] [server]       docs: [Object],
[server] [server]       publish: [Function: publish],
[server] [server]       registerPublisher: [Function: registerPublisher],
[server] [server]       _emit: [Function: emit],
[server] [server]       [Symbol(@feathersjs/transport-commons/publishers)]: {}
[server] [server]     },
[server] [server]     arguments: [ [Object] ],
[server] [server]     event: null,
[server] [server]     type: 'before'
[server] [server]   }
[server] [server] }
```

jonathan-casarrubias:
@hanzlamateen it should be working now, the problem was that a modal from admin is being used in studio to publish locations, so when I assigned the action: 'admin' because it's an admin app component, when used in studio a regular user didn't have the right permissions, I had to change it to action: studio in order for a regular user to be granted with permissions to use that modal.

jonathan-casarrubias:
@hanzlamateen you know what i'll add an extra parameter to the modal to pass the action so we better pass the action from outside the modal, so we can distinguish admin from studio from dashboard

jonathan-casarrubias:
@hanzlamateen ok so IMO this is a better solution, I added the action parameter and passed from each app accordingly and it's correctly working

image

hanzlamateen:
> @hanzlamateen ok so IMO this is a better solution, I added the action parameter and passed from each app accordingly and it's correctly working
> 
> image

Yes I agree, passing action is a better option.

hanzlamateen (approved):

PR #233, opened 2024-09-13
IR-4540 changing camera lerping targets and lerp methodology

SamMazerIR:
## Summary
changing infinite approach lerping toward targets in for timed A to B lerp that ends with always being on target after the specified time. Time is set as a const

There is a case for revisiting all cases of the "smootheLerpAlpha" function, as it is being used incorrectly. It is meant to provide a somewhat logarithmic graph but only if you pass it an accumulating value for time. It is being used with deltaTime instead, which is the time since last frame (and performance dependent)...basically meaning that you would have a roughly fixed value from smootheLerpAlpha given a stable framerate, which would never approach 1.

This also no longer lerps each frame from current position/rotation to target pos/rot (meaning it would effectively accelerate the lerp over time rather than have a fixed speed)

## References
[https://tsu.atlassian.net/browse/IR-4540](https://tsu.atlassian.net/browse/IR-4540)

## QA Steps
in studio hit the play button after moving the camera to wherever you want in the viewport (drag left click or hold right click and use wasd to move)

watch as the camera smoothly lerps to the avatar follow position (3rd person over the shoulder by default) and doesn't first travel to the avatar's feet or do anything weird

after lerping finishes in 1 second the camera should follow rotations and movements of the camera/avatar as usual


packages/spatial/src/camera/components/FollowCameraComponent.ts
```diff
@@ -378,7 +388,11 @@ const computeCameraFollow = (cameraEntity: Entity, referenceEntity: Entity) => {
   )
 
   mx.lookAt(direction, empty, upVector)
-  cameraTransform.rotation.setFromRotationMatrix(mx)
+
+  //slerp using rotationLerp value, this is reset to zero every time the follow target changes
+  const camRot = cameraTransform.rotation.clone()
+  camRot.slerpQuaternions(follow.originalRotation ?? camRot, new Quaternion().setFromRotationMatrix(mx), lerpVal)
```
HexaField: need to avoid `.clone` and `new` calls in hot code paths

packages/spatial/src/camera/components/FollowCameraComponent.ts
```diff
@@ -195,32 +216,43 @@ const tempVec1 = new Vector3()
...
   let isInsideWall = false
 
-  const offsetAlpha = smootheLerpAlpha(follow.offsetSmoothness, getState(ECSState).deltaSeconds)
   const targetOffset =
-    follow.mode === FollowCameraMode.FirstPerson ? follow.firstPersonOffset : follow.thirdPersonOffset
-  follow.currentOffset.lerp(targetOffset, offsetAlpha)
+    follow.mode === FollowCameraMode.FirstPerson
+      ? follow.firstPersonOffset
+      : follow.thirdPersonOffset.y === 0
+      ? new Vector3(0, cameraTransform.position.y, 0)
```
HexaField: avoid `new`

HexaField (dismissed):

packages/spatial/src/camera/components/FollowCameraComponent.ts
SamMazerIR: avoided now

MichaelEstes (approved):

PR #234, opened 2024-09-13
develop scene complexity checks and warnings

SYBIOTE:
Add RenderMonitorSystem and EditorWarningState
to monitor scene complexity and provide warnings


SYBIOTE:
https://github.com/user-attachments/assets/96f313bf-06ad-4fb9-ae69-d7c90f02603c

packages/engine/src/scene/constants/SceneConstants.ts
```diff
@@ -0,0 +1,44 @@
...
+  shaderComplexityWeight: 1.0
+}
+
+// these are thresholds for the scene complexity
+export const SceneComplexity = {
+  VeryLight: { label: 'VeryLight', value: 50000 } as const,
+  Light: { label: 'Light', value: 100000 } as const,
+  Medium: { label: 'Medium', value: 150000 } as const,
+  Heavy: { label: 'Heavy', value: 200000 } as const,
+  VeryHeavy: { label: 'VeryHeavy', value: 300000 } as const
```
aditya-mitra:
```suggestion
  VeryLight: { label: 'Very Light', value: 50000 } as const,
  Light: { label: 'Light', value: 100000 } as const,
  Medium: { label: 'Medium', value: 150000 } as const,
  Heavy: { label: 'Heavy', value: 200000 } as const,
  VeryHeavy: { label: 'Very Heavy', value: 300000 } as const
```

packages/editor/src/systems/RenderMonitorSystem.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+    useEffect(() => {
+      // these thresholds are to be adjusted  based on experimentation
+      let warning = t('editor:warnings.sceneComplexity', { sceneComplexity: SceneComplexity.VeryHeavy.label })
+      if (sceneComplexityScore.value < SceneComplexity.VeryLight.value) return
+      if (sceneComplexityScore.value < SceneComplexity.Light.value) return
+      if (sceneComplexityScore.value < SceneComplexity.Medium.value) return
+      if (sceneComplexityScore.value < SceneComplexity.Heavy.value)
+        warning = t('editor:warnings.sceneComplexity', { sceneComplexity: SceneComplexity.Heavy.label })
+      getMutableState(EditorWarningState).warning.set(warning)
```
aditya-mitra, line 125: the state we set here is a simple string. we can instead use `NotificationService.dispatchNotify` here directly. This will avoid us having `useEffect`s

packages/editor/src/components/EditorContainer.tsx
```diff
@@ -76,6 +78,18 @@ export const DockContainer = ({ children, id = 'editor-dock', dividerAlpha = 0 }
...
+const onEditorWarning = (warning) => {
+  console.warn(warning)
+  NotificationService.dispatchNotify(warning, {
+    variant: 'warning'
+  })
+
+  // popover design doesnt match the figma designs, we use notification for now
+  /*PopoverState.showPopupover(
+    <WarningDialog title={t('editor:warning')} description={warning || t('editor:warningMsg')} />
+  )*/
```
aditya-mitra, line 91: design is using a notification
image

hence, we can remove this comment

packages/editor/src/components/EditorContainer.tsx
```diff
@@ -76,6 +78,18 @@ export const DockContainer = ({ children, id = 'editor-dock', dividerAlpha = 0 }
   )
 }
 
+const onEditorWarning = (warning) => {
+  console.warn(warning)
```
aditya-mitra, line 83: this console warning can be removed since notification message has the same text

packages/ui/src/components/tailwind/WarningDialog/index.tsx
aditya-mitra, line 1: lets remove the warning dialog since design is also using a notification message

aditya-mitra (dismissed):
nice work with the counting flow!

packages/editor/src/systems/RenderMonitorSystem.tsx
SYBIOTE: i agree, i decided to build out the warning system to allow us to dispatch warnings from anywhere in the code base more easily

packages/editor/src/components/EditorContainer.tsx
SYBIOTE: i left it in in case we want to build a customized modal around it later, i would suggest keeping it in, it will reduce dev effort later

packages/editor/src/components/EditorContainer.tsx
SYBIOTE: this is so that the warning is logged in the console, and we can extract the logs for other purposes later, and the warning stays in there

packages/ui/src/components/tailwind/WarningDialog/index.tsx
SYBIOTE: kept it in incase we want a different customized design, the notification based is the first iteration, we may have a different design later and the warning dialog will be changed to suit it accordingly

packages/editor/src/systems/RenderMonitorSystem.tsx
```diff
@@ -0,0 +1,130 @@
...
+    SceneComplexityWeights.shaderComplexityWeight * params.shaderComplexity
+  return complexity
+}
+
+// count number of lines in vertex and fragment shaders
+function getShaderComplexity(resources: Record<string, any>): number {
+  const countLines = (code?: string): number => (code || '').split('\n').length
+  const totalComplexity = Object.values(resources)
+    .filter((resource) => resource.type === ResourceType.Material)
+    .flatMap((resource) => (Array.isArray(resource.asset) ? resource.asset : [resource.asset]))
```
jonathan-casarrubias: It really isn't necessary to have the ternary operator, nor wrapping resource.asset into a new array, by directly returning resource.asset should work.

Consider the following link as reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

packages/editor/src/systems/RenderMonitorSystem.tsx
```diff
@@ -0,0 +1,130 @@
...
+    .filter((material) => material.isShaderMaterial || material.isMeshStandardMaterial)
```
jonathan-casarrubias: There are 2 filtering iterations over the same collection, consider optimizing into a single iteration.

packages/editor/src/systems/RenderMonitorSystem.tsx
SYBIOTE: oh, cool ill change it then

SYBIOTE: thats great, thanks for pointing it out

packages/editor/src/systems/RenderMonitorSystem.tsx
SYBIOTE: yeah that makes sense since the flatmap doesnt exist anymore

packages/editor/src/systems/RenderMonitorSystem.tsx
SYBIOTE: applied

packages/editor/src/systems/RenderMonitorSystem.tsx
SYBIOTE: applied

MbfloydIR (approved):

HexaField:
Requesting review from @dinomut1 @MichaelEstes

packages/editor/src/systems/RenderMonitorSystem.tsx
```diff
@@ -0,0 +1,134 @@
...
+    const sceneComplexityScore = useHookstate(0)
+
+    const renderInfoState = useHookstate(getMutableState(RenderInfoState))
+    const resourceState = useHookstate(getMutableState(ResourceState))
+    const lightQuery = useQuery([LightTagComponent, VisibleComponent, SourceComponent])
+
+    useEffect(() => {
+      const params = {
+        vertices: resourceState.totalVertexCount.value,
+        triangles: renderInfoState.info.triangles.value,
```
MichaelEstes, line 100: I'm not sure that we want to use the render info triangle count here. The reason for storing our own vertex count is because not all geometry should be considered part of the scene budget (Things like transform gizmos and debug meshes).

HexaField: yes, we need to iterate from the loaded scene gltf entity and only include deeply nested entities that belong to the scene

SYBIOTE: roger that, this would still give us a very close estimate, since its only a part of calculating the scene complexity, it would not affect the final calculation all that much

MichaelEstes (approved):

HexaField (approved):

PR #235, opened 2024-09-13
Signaling service, logic and transport for p2p instances

HexaField:
## Summary

- [x] convert 'currentUsers' column on `instance` table to resolver
- [x] add `peerId` and `peerIndex` column to `instance-attendance` table
- [x] `peerIndex` now incremented by `instance-attendance` table
- [x] add p2p instance provisioning logic
- [x] improve types for feathers primus web sockets
- [x] implement `instance-signaling` service methods that creates/patches `instance-attendance` table
- [x] implement `instance-signaling` service method that routes messages to a specific peer and respects feathers-sync
- [x] Implement p2p network transport
- [x] Implement p2p network logic
- [x] Implement p2p webrtc data channels
- [x] Implement p2p webrtc media channels
- [x] Event source peer state

## References
closes
- IR-4449
- IR-4450
- IR-4451
- IR-4901
- IR-4945


barankyle:
Saw a couple of issues when testing this locally:

1. If User A starts their audio, User B mutes the audio consumer, then User A stops and starts their audio, User B hears their audio again. This does not happen for video, User B's video consumer never starts playing again as long as User B has it 'muted'.
2. I'm fairly consistently seeing issues when a completely new user comes into a location with at least two other users present and doesn't render at least one other user's avatar and media bubble. The correct number of peers is present in NetworkState, it looks like it makes WebRTC connections to every other user, but FilteredUsersState.nearbyLayerUsers is not correct, e.g. it only has one userId when there are two other users present. I'm unsure what triggers avatar rendering, or lack thereof, but they're most likely connected. Refreshing doesn't always solve the problem. 
3. I see this error pretty consistently when any user joins, it doesn't seem to cause any immediate problems but should maybe be addressed?
```
Warning: Cannot update a component (`UserMediaWindows`) while rendering a different component (`MediaIconsBox`). To locate the bad setState() call inside `MediaIconsBox`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
    at reactor (https://localhost:3000/@fs/home/kyle/Development/Forks/ir-engine/packages/client-core/src/media/MediaStreamState.ts:51:19)
    at <anonymous> (https://localhost:3000/@fs/home/kyle/Development/Forks/ir-engine/packages/hyperflux/src/functions/ReactorFunctions.tsx:57:10)
    at Suspense
    at ReactorContainer (https://localhost:3000/@fs/home/kyle/Development/Forks/ir-engine/packages/hyperflux/src/functions/ReactorFunctions.tsx:117:25)
```

barankyle:
I did not see the issue with other users' avatars and bubbles not appearing when I disabled P2P - everyone showed up every time. I did see an issue with the consumer mute state. If I muted someone else's audio/video, the button changed to the muted state (red, crossed through), but the audio/video did not stop. If I clicked the button again, the button went back to the unmuted state, but the audio/video itself was then muted.

barankyle:
I'm also seeing an issue with P2P off. If User A enters a location, they get assigned to an instance. User B enters, and they get assigned to the same instance. If one of them leaves and then goes back to that location, the instance-provision is returning a different instance.

I think this is because we're not updating the currentUsers column in the `instance` table any more, but there's probably still some logic looking at that. I see when a user leaves, the `ended` column for both active instanceservers is getting changed from false to true. This might be happening in instance-signaling.ts around line 232, where on instance-attendance patches, we're making the instance record `ended: true` if currentUsers is 0.

I see that the currentUsers column is supposed to not exist any more. It looks like the migration has the wrong comparison for dropping that column - it's only dropping the column if it doesn't exist. But anything referencing currentUsers will need to be updated to remove currentUsers; they're probably being hidden by the fact that the migration isn't removing that column.

HexaField:
Haven't yet fixed the audio muting/unmuting bug, but I think everything else should be addressed now.

barankyle:
Still seeing the reversed consumer mute/unmute behavior when in server-SFU mode. P2P mode works as intended.

I also see an issue in server-SFU mode where if Client B has their video enabled and then leaves, it's breaking some of the state of Client A. Client A will not show a user bubble for Client B upon their return, nor any other clients that join after Client B left.

I've occasionally seen new clients not make every peer media connection properly when there were more than two clients in an instance, or maybe it's just not rendering the bubble. This was annoyingly intermittent. World connections all appeared to be made every time.

In P2P mode, if you start an instanceserver with P2P_INSTANCE_ENABLED not 'true', but the API server is started with that var as 'true', the instanceservers are getting a feathers-sync instance-signaling message and crashing with 


```
[12:51:41.067] FATAL: UNCAUGHT EXCEPTION.
    component: "server-core:config"
    err: {
      "type": "Error",
      "message": "Can not find service 'instance-signaling'",
      "stack":
          Error: Can not find service 'instance-signaling'
              at Feathers.defaultService (/home/kyle/Development/Forks/ir-engine/node_modules/@feathersjs/feathers/src/application.ts:58:11)
              at Feathers.service (/home/kyle/Development/Forks/ir-engine/node_modules/@feathersjs/feathers/src/application.ts:68:27)
              at Feathers.<anonymous> (/home/kyle/Development/Forks/ir-engine/node_modules/feathers-sync/lib/core.js:28:25)
              at Feathers.emit (node:events:517:28)
              at Feathers.emit (node:domain:489:12)
              at msgFromRedisHandler (/home/kyle/Development/Forks/ir-engine/node_modules/feathers-sync/lib/adapters/redis.js:22:11)
              at PubSub._PubSub_emitPubSubMessage (/home/kyle/Development/Forks/ir-engine/node_modules/@redis/client/dist/lib/client/pub-sub.js:294:9)
              at PubSub.handleMessageReply (/home/kyle/Development/Forks/ir-engine/node_modules/@redis/client/dist/lib/client/pub-sub.js:233:93)
              at Object.onReply (/home/kyle/Development/Forks/ir-engine/node_modules/@redis/client/dist/lib/client/commands-queue.js:40:87)
              at RESP2Decoder.write (/home/kyle/Development/Forks/ir-engine/node_modules/@redis/client/dist/lib/client/RESP2/decoder.js:119:26)
              at RedisCommandsQueue.onReplyChunk (/home/kyle/Development/Forks/ir-engine/node_modules/@redis/client/dist/lib/client/commands-queue.js:154:72)
              at RedisSocket.<anonymous> (/home/kyle/Development/Forks/ir-engine/node_modules/@redis/client/dist/lib/client/index.js:420:84)
              at RedisSocket.emit (node:events:517:28)
              at RedisSocket.emit (node:domain:489:12)
              at Socket.<anonymous> (/home/kyle/Development/Forks/ir-engine/node_modules/@redis/client/dist/lib/client/socket.js:201:42)
              at Socket.emit (node:events:517:28)
              at Socket.emit (node:domain:489:12)
              at addChunk (node:internal/streams/readable:368:12)
              at readableAddChunk (node:internal/streams/readable:341:9)
              at Socket.Readable.push (node:internal/streams/readable:278:10)
              at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
              at TCP.callbackTrampoline (node:internal/async_hooks:128:17)
```

Looks like it's because the instance-signaling service just returns if p2pEnabled is false, so I guess this is just a weird local edge case when changing the config.

packages/network/src/EntityNetworkState.tsx
```diff
@@ -161,35 +157,48 @@ const EntityNetworkReactor = (props: { uuid: EntityUUID }) => {
...
-  }, [networkState.peers, networkState.users])
+  }, [isOwner, isAuthorInNetwork])
 
   return null
 }
+
+/**
+ * Get a deterministic network ID scoped to each owner peer
+ */
+const useNetworkID = (uuid: EntityUUID) => {
```
the-reality-engineer, line 199: Awesome

the-reality-engineer (approved):
Only potential issue I'm seeing is how we are handling HyperFlux.store.userID

We assign the id in AuthService.ts after the user is authenticated, but we are relying on HyperFlux.store.userID (or Engine.instance.userID) in several places within reactors, and it's not a reactive state. I don't think we are guaranteeing anywhere that the user is authenticated before we use the `userID` value, or before we dispatch actions using it (and even if we are, seems like something we should make reactive just in case). This was already implemented like this before this PR, but it stood out to me as a potential problem.

HexaField:
> Only potential issue I'm seeing is how we are handling HyperFlux.store.userID
> 
> We assign the id in AuthService.ts after the user is authenticated, but we are relying on HyperFlux.store.userID (or Engine.instance.userID) in several places within reactors, and it's not a reactive state. I don't think we are guaranteeing anywhere that the user is authenticated before we use the `userID` value, or before we dispatch actions using it (and even if we are, seems like something we should make reactive just in case). This was already implemented like this before this PR, but it stood out to me as a potential problem.

Yeah I have been aware of this for a while. We can do this another time, since it currently doesn't cause any known issues.

PR #241, opened 2024-09-16
studio: update browser detection

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-4471
https://tsu.atlassian.net/browse/IR-4570

## References
closes #_insert number here_

## QA Steps
Screenshot 2024-09-16 at 11 36 00 AM


packages/editor/src/functions/browserCheck.ts
```diff
@@ -25,12 +25,7 @@ Infinite Reality Engine. All Rights Reserved.
 
 export const isSupportedBrowser = () => {
   const userAgent = navigator.userAgent as any
```
aditya-mitra: why are we typecasting to `any` type here?
if it comes out to be a separate type, can you try `window.navigator.userAgent`

aditya-mitra (changes requested):

aditya-mitra (changes requested):
also lets store this in localstorage to avoid prompting the user once selected

packages/editor/src/functions/browserCheck.ts
```diff
@@ -24,13 +24,13 @@ Infinite Reality Engine. All Rights Reserved.
 */
 
 export const isSupportedBrowser = () => {
-  const userAgent = navigator.userAgent as any
+  const userAgent = window.navigator.userAgent
   const winNav = window.navigator as any
 
   const isGoogleChrome =
     /Chrome/.test(userAgent) && typeof winNav.userAgentData !== 'undefined'
       ? winNav.userAgentData.brands?.some((item) => item.brand === 'Google Chrome')
```
achen5671: @aditya-mitra I think we can keep this here since we refactor this so it wont crash the app if `brands` is null. I added an extra step to check for a other chromium browsers to be safe. 

Let me know if you would prefer to remove the use of useAgentData completly although I think it's good to have for now

packages/editor/src/pages/EditorPage.tsx
```diff
@@ -60,7 +61,7 @@ export const EditorPage = () => {
   const [params] = useSearchParams()
   const { scenePath, projectName } = useHookstate(getMutableState(EditorState))
   const supportedBrowser = useHookstate(isSupportedBrowser)
-  const acknowledgedUnsupportedBrowser = useHookstate(false)
+  const acknowledgedUnsupportedBrowser = useHookstate(localStorage.getItem('acknowledgedUnsupportedBrowser') ?? false)
```
aditya-mitra: can you do it using state definition? (using `syncStateWithLocalStorage`)
 (example: https://github.com/ir-engine/ir-engine/blob/641da655a0365ddb28c9f9a288a54ffd6948cf1c/packages/editor/src/services/HierarchyNodeState.ts/#L42)

packages/editor/src/functions/browserCheck.ts
aditya-mitra: `userAgentData` is [unavailable in a lot of browsers](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData#browser_compatibility) (including all mobile browsers)
we can avoid the ternary conditions (separate testing again for safari) and for maintenance purposes just narrow it down to `window.navigator.userAgent`

aditya-mitra (changes requested):

packages/editor/src/functions/browserCheck.ts
achen5671: cool, updated

aditya-mitra (approved):
awesome 🚀
(now i like my browser)

PR #242, opened 2024-09-16
studio: max user exceeded error on publish

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-4242
## References
closes #_insert number here_

## QA Steps
Screenshot 2024-09-16 at 11 46 47 AM


aditya-mitra (changes requested):
looks good on the client side

can you also a validation on the server side as well?
(it would be inside location.resolvers.ts and also in schema if this is a strict upper limit)

barankyle (approved):

packages/server-core/src/social/location/location.resolvers.ts
```diff
@@ -137,7 +140,15 @@ export const locationPatchResolver = resolve<LocationType, HookContext>({
   updatedBy: async (_, __, context) => {
     return context.params?.user?.id || null
   },
+  maxUsersPerInstance: resolvedMaxUserPerInstance,
   updatedAt: getDateTimeSql
 })
 
+async function resolvedMaxUserPerInstance(value, location) {
+  if (location.maxUsersPerInstance > MAX_USER_PER_INSTANCE) {
+    throw new BadRequest('You have entered higher than the allowed max users. Please enter a number between 1 and 10.')
```
aditya-mitra, line 149: `MAX_USER_PER_INSTANCE` is 100 but we are adding a range upto 10

aditya-mitra (dismissed):

packages/server-core/src/social/location/location.resolvers.ts
achen5671: yea, I confirmed with ross yesterday. I changed it to 100 because test was failing. Ill push up a new commit to fix

PR #249, opened 2024-09-16
Fix root cookie accesor

DanielBelmes:
## Summary
Because we are json.parsing the data we were attempting to parse data from other iframes loaded from extensions. So if a customer has an extension that does postmessage it breaks auth. This fixes it by just sending data as an object and passing in a "source" which is one way that people mark post message data. We can discuss ways to identify the source.

## References
closes #_insert number here_


barankyle (approved):

DanielBelmes:
Was testing locally while in merge queue and saw some weird behavior so doing another testing pass.

packages/client-core/src/user/services/AuthService.ts
```diff
@@ -144,17 +145,17 @@ const waitForToken = async (win, clientUrl): Promise<string> => {
     )
     const getIframeResponse = function (e) {
       if (e.origin !== clientUrl) return
-      if (e?.data) {
+      if (e?.data && e.data.source === iframeSource) {
```
jonathan-casarrubias: I'm assuming that validating e.data.source === iframeSource may implicitly assume that e.data.data exists and that it's a string, but it still leaves room for unreliability.

It's just safer to avoid any assumption and to explicitly validate the specific parameters or variables that are being used within the logical block.

My suggestion is to explicitly validate that e.data.data exists and that it in fact is a string, by making explicit validations, the code is future proof, even if for some reason APIs change down the road.

```suggestion
      if (e?.data && e.data.source === iframeSource && typeof e.data?.data === 'string') {
```

packages/client-core/src/user/services/AuthService.ts
```diff
@@ -224,20 +228,20 @@ const getToken = async (): Promise<string> => {
...
     } else {
-      iframe.style.display = 'block'
+      iframe.style.visibility = 'visible'
       return await new Promise((resolve) => {
         const clickResponseListener = async function (e) {
           if (isRootCookieAncestorMessage(e)) {
             try {
               window.removeEventListener('message', clickResponseListener)
-              const parsed = !e.data ? {} : JSON.parse(e.data)
+              const parsed = !e.data || e.data.source !== iframeSource ? {} : e.data
```
jonathan-casarrubias: The "parsed" constant name isn't semantic, nothing is being parsed during the process of assigning a value, not the end of the world but semantic is the practice that promotes code readability and consistency, sometimes it even avoids unnecessary mutations and/or data duplicity.

packages/client-core/src/user/services/AuthService.ts
```diff
@@ -614,9 +618,9 @@ export const AuthService = {
         const clientUrl = config.client.clientUrl
         const getIframeResponse = function (e) {
           if (e.origin !== clientUrl) return
-          if (e?.data) {
+          if (e?.data && e.data.source == iframeSource) {
             try {
-              const value = JSON.parse(e.data)
+              const value = e.data
```
jonathan-casarrubias: Along the lines of pondering semantics, this one goes for consistency, same operation having 2 different constant names, neither is semantic nor consistent.

packages/client/public/root-cookie-accessor-template.html
```diff
@@ -1,141 +1,169 @@
...
-                    .split("; ")
-                    .find((row) => row.startsWith(`allowedDomains=`))
-                allowedDomains = allowedDomainCookie ? JSON.parse(allowedDomainCookie.split("=")[1]) : []
-            } catch (err) {
-                allowedDomains = []
+      let origin
+      const source = "ir-engine-root-cookie"
+      const processMessage = async (e) => {
+        origin = e.origin
+        const payload = e?.data && typeof e.data === 'string' ? JSON.parse(e.data) : null
```
jonathan-casarrubias: Why are we parsing e.data.data as a string in other places, while here we are still parsing e.data?

mikeplascdev (approved):
LGTM

packages/client/public/root-cookie-accessor-template.html
```diff
@@ -1,141 +1,169 @@
...
+        if (payload == null) return
```
jonathan-casarrubias: If payload is null, the subsequent code logic isn't executed and if it isn't executed something won't correctly work, therefore there's no difference between having this validation or not having it at all.

In order to extract value from this validation, failure recovery has to be implemented if possible, if it isn't possible then creating a log report for traceability has to be implemented.

packages/client/public/root-cookie-accessor-template.html
```diff
@@ -1,141 +1,169 @@
...
+        }
+      }
+      document.addEventListener('DOMContentLoaded', async (event) => {
+        document.getElementById('confirmButton').onclick = async (event) => {
+          try {
+            const requestStorageAccessResponse = await document.requestStorageAccess()
+            if (requestStorageAccessResponse == null) {
+              const cookieSet =
+                document.cookie &&
+                document.cookie
```
jonathan-casarrubias: Document.cookie is being processed multiple times all over the place, creating and allocating in memory each time a new collection (split), iterating multiple times (find), mutating iterations (replace), etc.

As a general rule, each time we realize that we are doing the same operation more than once, we must reconsider an alternate approach.

The best way to avoid iterating the same data over and over is by creating a key-value map once, enabling data values to be accessible without the need of iterating the collection each time that a value is required.

```suggestion
   // We iterate just once and allocate in memory once
   const cookiesMapObj = document.cookie.split('; ').reduce((accumulator, current) => {
  const values = current.split('=')
  return {...accumulator, values[0]: values[1]}
}, {})
```

Now each time we need to access a cookie data value we can simply do it by the key reference

```suggestion
 someFunction(cookiesMapObj['user-token'])
```

Now we avoided tons of iterations, duplicated code, unnecessary complexity, misused memory allocation and a it's lot less prone to failure.

jonathan-casarrubias (changes requested):

packages/client-core/src/common/iframeCOM.ts
```diff
@@ -0,0 +1,89 @@
...
+  private messageQueue: Map<
+    string,
+    [(value: MessageResponse | PromiseLike<MessageResponse>) => void, (reason?: any) => void]
+  > = new Map()
+
+  constructor(iframeId: string, targetOrigin: string) {
+    this.iframeId = iframeId
+    this.targetOrigin = targetOrigin
+    this.iframe = document.getElementById(iframeId) as HTMLIFrameElement
+    window.addEventListener('message', this.handleMessage.bind(this))
```
jonathan-casarrubias, line 54: First of all, I'd like to recognize that today's updates are a lot more elegant, great work.

Now, I tried to find a place where this event listener is removed but couldn't find any window.removeEventListener for this.handleMessage, it might potentially be a memory leak.

packages/client-core/src/user/services/AuthService.ts
```diff
@@ -133,116 +137,87 @@ const resolveWalletUser = (credentials: any): UserType => {
...
+  if (error?.data?.invalidDomain) {
+    try {
+      localStorage.setItem('invalidCrossOriginDomain', 'true')
+    } catch (err) {
+      console.log('Was not able to read invalid Domain messaging', err)
+    }
+  }
+}
+
+const waitForToken = async (win: Window, clientUrl: string): Promise<string> => {
```
jonathan-casarrubias: Combining await with then, finally and catch in the same promise it's an anti-pattern.

For example if something goes wrong within the then handler, it won't throw an exception at the waitForToken level, therefore it's impossible to handle exceptions when calling waitForToken

```ts
try {
  await waitForToken()
} catch() {
  // this will never be executed even if internally an error occurs
}
```

There are many resources available on internet explaining why these shouldn't be combined.

packages/client-core/src/user/services/AuthService.ts
```diff
@@ -133,116 +137,87 @@ const resolveWalletUser = (credentials: any): UserType => {
...
-              } else {
-                const token = await waitForToken(win, clientUrl)
-                iframe.style.display = 'none'
-                resolve(token)
-              }
-            } catch (err) {
-              //Do nothing
+          if (e.origin !== config.client.clientUrl || e.source !== iframe.contentWindow) return
+          try {
+            window.removeEventListener('message', clickResponseListener)
```
jonathan-casarrubias: This listener is being removed before executing the operation, if internally something goes wrong, the listener is no longer available potentially resulting in a poor user experience.

packages/client-core/src/user/services/AuthService.ts
```diff
@@ -133,116 +137,87 @@ const resolveWalletUser = (credentials: any): UserType => {
...
+              localStorage.setItem('skipCrossOriginCookieCheck', 'true')
+              iframe.style.visibility = 'hidden'
+              resolve('')
+            } else {
+              const token = await waitForToken(win, clientUrl)
+              iframe.style.visibility = 'hidden'
+              resolve(token)
             }
+          } catch (err) {
+            //Do nothing
```
jonathan-casarrubias, line 224: This catch won't ever be executed if anything goes wrong inside waitForToken for the reasons I mentioned before.

If this logic is triggered by a click event listener (user interaction) we need this catch to be executed on exceptions and to provide the user with feedback.

packages/client/public/root-cookie-accessor-template.html
```diff
@@ -1,141 +1,177 @@
...
-                        document.cookie = `allowedDomains=${JSON.stringify(allowedDomains)}; SameSite=None; Secure`
-                        await processMessage(e)
-                    }
-                } else {
-                    parent.postMessage(JSON.stringify({invalidDomain: true}), e.origin)
+            message.data = result
+            parent.postMessage(message, e.origin)
+            break
+          case 'remove':
+            document.cookie = document.cookie
```
jonathan-casarrubias, line 43: I see that you already implemented the getCookies() function and the map object is now being used in most places, but is there a reason to parse and iterate the cookies again in this line?

jonathan-casarrubias (dismissed):

packages/client/public/root-cookie-accessor-template.html
DanielBelmes: Easier to not have to rebuild all the cookie syntax from getCookies. Less complexity and I didn't mind inefficiency of something that leads to logging out.

packages/client-core/src/common/iframeCOM.ts
DanielBelmes: I'll add a destroy concept. Not sure where it's gonna execute yet.

packages/client-core/src/user/services/AuthService.ts
DanielBelmes: I've found an article going over it and will iron out the anti pattern.

PR #251, opened 2024-09-17
feat(tables): Fixed vertical scroll and now the tables fit on the screen sizes

kattsushi:
## Summary
feat(tables): Fixed vertical scroll and now the tables fit on the screen sizes

- Applied truncation ellipsis  to both 'start', 'middle', and 'end' positions as needed for better user experience as need it like commit Sha columns.
- Composition for copyable and links truncated text
- Ability to copy on some cases the content truncated
- Fixed Action buttons alignments 
- Fixed Vertical scroll and now the tables fit on the screen sizes

example:

https://github.com/user-attachments/assets/f1654961-8396-4576-9444-589c5447f7d2

## References
related to Jira [#IR-3741](https://tsu.atlassian.net/browse/IR-3741) Issue


packages/client/public/root-cookie-accessor.html
barankyle, line 1: This file is generated from root-cookie-accessor-template.html, but shouldn't be committed, as it is deployment-specific. It was originally not included in the .gitignore, but is now; you might have started this branch before its inclusion in the .gitignore. Please remove this from the branch.

kattsushi: got it!

packages/client-core/src/admin/components/avatar/AvatarTable.tsx
```diff
@@ -129,12 +129,18 @@ export default function AvatarTable({ search }: { search: string }) {
...
+            startIcon={<HiTrash className="place-self-center text-theme-iconRed" />}
+          />
         </div>
       )
     }))
 
-  return <DataTable query={adminAvatarQuery} columns={avatarColumns} rows={createRows(adminAvatarQuery.data)} />
+  return (
+    <DataTable
+      className="h-[calc(100vh_-_318px)]"
```
hanzlamateen: If we have to add this property on all tables, can we make it a behavior of `DataTable` control by default? Or if required we can make it toggle able through a property `isFullScreen` or `isResponsive`?

packages/ui/src/primitives/tailwind/TruncatedLink/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  return truncators[ellipsisPosition]()
+}
+
+type TruncatedTextProps = {
+  text: string
+} & TruncateTextOptions
+
+export const TruncatedCopyableText: FC<TruncatedTextProps> = (props) => {
```
hanzlamateen: Instead of have separate controls like `TruncatedLink` & `TruncatedLinkText` maybe have a single `TruncatedText` control and then introduce additional features through properties and/or a `variant` property.

Also then update this file's name to match component name.

hanzlamateen (dismissed):
Can you make horizontal scrollbar size similar to vertical scrollbar.

packages/client-core/src/admin/components/avatar/AvatarTable.tsx
kattsushi: Yes, by default the table adjusts its height to fit the number of rows it can render (up to a limit). However, the desired behavior is for the tables to always fit the screen size. The way I made that possible was by calculating the height of each table page because each page has a different layout, such as header filters, tabs, or page tools in the top. This ensures there are no unexpected vertical scrolls or hidden content.

One possible solution could be to introduce 4 different sizes, for example: `sm`, `md`, `lg`, and `xl`. However, as you can see, most table pages don't match exactly across sizes, and this could lead to issues like unexpected vertical scrolls, hidden content, etc.

packages/ui/src/primitives/tailwind/TruncatedLink/index.tsx
kattsushi: thanks! already addressed

packages/client-core/src/admin/components/avatar/AvatarTable.tsx
kattsushi: i think the options with the variants by size is cleaner just implemented, what do you think ?

kattsushi:
> Can you make horizontal scrollbar size similar to vertical scrollbar.

thanks good catch, already fixed

packages/client-core/src/admin/common/Table.tsx
```diff
@@ -112,14 +115,16 @@ const DataTable = ({ query, columns, rows }: DataTableProps) => {
...
+    <div className="relative">
       {query.status === 'pending' && (
         <div className="absolute left-1/2 top-1/2 flex h-8 -translate-x-1/2 -translate-y-1/2 items-center">
           <LoadingView className="mx-1 h-8 w-8" />
           <Text className="mx-1">{t('common:table.refetching')}</Text>
         </div>
       )}
-      <Table containerClassName={`${query.status === 'pending' && 'opacity-50'} h-[calc(100%_-_160px)]`}>
+      <Table
+        containerClassName={twMerge(`${query.status === 'pending' && 'opacity-50'} table-size-${size}`, className)}
```
aditya-mitra:
```suggestion
        containerClassName={twMerge(query.status === 'pending' && 'opacity-50', `table-size-${size}`, className)}
```

packages/ui/src/primitives/tailwind/TruncatedText/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+*/
+
+import React, { AnchorHTMLAttributes, FC } from 'react'
+import CopyText from '../CopyText'
+import Tooltip from '../Tooltip'
+
+type TruncateTextOptions = {
+  ellipsisChar?: string
+  visibleChars?: number
+  ellipsisPosition?: 'start' | 'middle' | 'end'
```
aditya-mitra: ellipsis are always at the end (by definition they mean _so forth_)
hence we the ellipsis position would always be at the end

packages/ui/src/primitives/tailwind/TruncatedText/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+}
+
+export const truncateText = (text: string, options: TruncateTextOptions = {}) => {
+  const { visibleChars = 3, ellipsisPosition = 'middle', ellipsisChar = '...' } = options
+
+  if (text.length <= visibleChars * 2) return text
+
+  const truncators: Record<NonNullable<TruncateTextOptions['ellipsisPosition']>, () => string> = {
```
aditya-mitra:
```suggestion
  const truncators = {
```
type will be deduced as it is a constant

packages/ui/src/primitives/tailwind/TruncatedText/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  return truncators[ellipsisPosition]()
+}
+
+type TruncatedTextProps = {
+  text: string
+  variant?: 'copy' | 'text'
+} & TruncateTextOptions
+
+const TruncatedText: FC<TruncatedTextProps> = (props) => {
+  const { variant = 'copy', text, visibleChars = 3, ellipsisPosition = 'middle', ellipsisChar = '...' } = props
```
aditya-mitra: by default the variant should be `text`
since by definition we are calling it Truncated**Text**

packages/ui/src/primitives/tailwind/Table/index.tsx
```diff
@@ -57,7 +57,7 @@ const TableHeadRow = ({
   children: JSX.Element | JSX.Element[]
 }) => {
   const twClassName = twMerge('text-left capitalize', className)
-  const twClassNameThead = twMerge('sticky top-0 z-10 bg-theme-table-secondary', theadClassName)
+  const twClassNameThead = twMerge('sticky top-[-2px] z-10 bg-theme-table-secondary', theadClassName)
```
aditya-mitra, line 60: why did we add -2px?
(i think it should be a page layout that would be causing the issue if it has moved downward)

packages/client/src/themes/utilities.css
```diff
@@ -1 +1,42 @@
...
+
+::-webkit-scrollbar-thumb {
+  background: #888;
+  border-radius: 10px;
+}
+
+/* Firefox */
+html {
+  scrollbar-width: thin; /* Tamaño más delgado del scrollbar */
+  scrollbar-color: #888 #f1f1f1; /* Color del thumb y fondo */
```
aditya-mitra: we can remove the comments

packages/client-core/src/admin/adminRoutes.tsx
```diff
@@ -6,8 +6,8 @@ Version 1.0. (the "License"); you may not use this file except in compliance
 with the License. You may obtain a copy of the License at
 https://github.com/ir-engine/ir-engine/blob/dev/LICENSE.
 The License is based on the Mozilla Public License Version 1.1, but Sections 14
-and 15 have been added to cover use of software over a computer network and 
-provide for limited attribution for the Original Developer. In addition, 
+and 15 have been added to cover use of software over a computer network and
+provide for limited attribution for the Original Developer. In addition,
```
aditya-mitra: Hey Andres, can you recheck your vscode editorconfig or prettier config for which these lines are getting changed during commits?
(with them absent, your PR looks more concise and shorter)

packages/client/src/themes/utilities.css
```diff
@@ -1 +1,42 @@
...
+  height: calc(100vh - 402px);
+}
+
+.table-size-lg {
+  height: calc(100vh - 386px);
+}
+
+.table-size-xl {
+  height: calc(100vh - 318px);
+}
```
aditya-mitra: we should ask about this to the designers what the appropriate table size is
if there is height, we can simply make a record object of height name and pixels

packages/client-core/src/admin/common/Table.tsx
aditya-mitra: since we are using `table-size-` tailwind utility at only 1 place, we can convert the size chart to a record object
for example,
```
{
 lg: 'h-[calc(100vh - 386px)]'
...
}
```

aditya-mitra (dismissed):
like your truncated text component - it can be easily utilized at different places

just a few changes

packages/ui/src/primitives/tailwind/Table/index.tsx
kattsushi: this is because now the header is fixed with the vertical scroll and the rows appears a bit above the the header

packages/ui/src/primitives/tailwind/TruncatedText/index.tsx
kattsushi: Yes, after thinking about it again, you're right. It should be named 'truncator' instead of 'ellipsis'. There are cases where the text should be truncated in the middle, like 'abc213...zbcd1' for IDs, for example, to shorten long IDs.

packages/client/src/themes/utilities.css
kattsushi: The issue here is that the behavior isn't for a fixed-height table layout. The calculation for the space taken by the utility tabs or filters above each page's table is different. To achieve the desired behavior, where the tables fit the remaining height of the screen, I have to calculate and adjust the pixels for each case, while also ensuring vertical scrolling within the tables,

Regarding the 'record object of height' in Tailwind it won't work in this case because it's a custom calculation in runtime, and Tailwind doesn't have a way to build classes and CSS for that. Let me think a bit more about how I can improve this by using the utility layers from Tailwind.

kattsushi: I made it work by extending the height theme configuration. This is cleaner.

```typescript
  theme: {
    extend: {
      height: {
        'table-size-xs': 'calc(100vh - 436px)',
        'table-size-sm': 'calc(100vh - 418px)',
        'table-size-md': 'calc(100vh - 402px)',
        'table-size-lg': 'calc(100vh - 386px)',
        'table-size-xl': 'calc(100vh - 318px)',
        'table-size-full': '100%',
      },
```

```typescript
const TABLE_HEIGHT_SIZES = {
  xs: 'h-table-size-xs',
  sm: 'h-table-size-sm',
  md: 'h-table-size-md',
  lg: 'h-table-size-lg',
  xl: 'h-table-size-xl',
  full: 'h-table-size-full'
}
```

packages/client-core/src/admin/common/Table.tsx
kattsushi: Thanks, the solution is above

packages/client-core/src/admin/adminRoutes.tsx
kattsushi: done! thx

packages/ui/src/primitives/tailwind/Table/index.tsx
aditya-mitra: @CITIZENDOT your thoughts?

CITIZENDOT (changes requested):
https://github.com/user-attachments/assets/83ccd41f-97b3-4326-a3e7-770be62062d9

I tested with dummy data, and Account's Table is still not fitting in the screen's height & Table's header is not fixed. Resource Table is a good example of how it should look like. I haven't tested other tables.

kattsushi:
> 2024-09-21.19-41-55.mp4 
> I tested with dummy data, and Account's Table is still not fitting in the screen's height & Table's header is not fixed. Resource Table is a good example of how it should look like. I haven't tested other tables.

Hey Appu that's why that part of the code is in another PR in the multitenansy project https://github.com/theinfinitereality/irpro-multitenancy/pull/490 here

CITIZENDOT:
> Hey Appu that's why that part of the code is in another PR in the multitenansy project [theinfinitereality/irpro-multitenancy#490](https://github.com/theinfinitereality/irpro-multitenancy/pull/490) here

Sorry, I missed that.
I just tested with that branch now, and everything works great. Nice work!

CITIZENDOT (approved):

PR #252, opened 2024-09-17
Add translation strings

CITIZENDOT:
## References
Needed for https://tsu.atlassian.net/browse/IR-4446


packages/ui/src/primitives/tailwind/Modal/index.tsx
```diff
@@ -123,7 +124,9 @@ export const ModalFooter = ({
 const Modal = ({
   id,
   title,
-  onClose,
+  onClose = () => {
```
HexaField: we should avoid having primitives importing state if possible

HexaField (changes requested):

packages/ui/src/primitives/tailwind/Modal/index.tsx
CITIZENDOT: Almost all modals has the same function set to `onClose`. Hence I set it as default prop. Shall I undo it?

HexaField: Yeah it would be nice to do that, but we definitely don't want to import client-core into these pure primitives. So we should revert

CITIZENDOT: noted.

HexaField (approved):

PR #260, opened 2024-09-17
IR-4351 positional audio max distance disabled state

SamMazerIR:
## Summary
this bug was technically a communication/knowledge issue. The 'Max Distance' is only used when the 'Distance Model' is set to 'Linear'. I've added a disabled state to the input field for the max distance as well as an updated visual and tooltip
image

making NumericInput respond to disabled and appropriate styling, making positional audio maxDistance respect distance model visually (disables unless distance model is Linear)

## References
[https://tsu.atlassian.net/browse/IR-4351](https://tsu.atlassian.net/browse/IR-4351)

## QA Steps
on a positional audio component change the distance model between linear and anything else, you should see max distance becomes disabled both visually and for input unless the distance model is set to linear. You should also note that hovering over the info icon of max distance will display a relevant message depending on the disabled state for max distance


DanielBelmes (approved):
Did check to see if design signed off on it. They did.

PR #261, opened 2024-09-17
Add thumbnail and envmap to publish modal

SYBIOTE:
This pull request adds the functionality to include a thumbnail and envmap in the publish modal.. It also adds new code to the `AddEditLocationModal` component to handle the thumbnail and loading screen image inputs.


packages/editor/src/services/SceneThumbnailState.tsx
```diff
@@ -0,0 +1,126 @@
...
+  initial: () => ({
+    thumbnailURL: null as string | null,
+    thumbnail: null as File | null,
+    uploadingThumbnail: false,
+    loadingScreenURL: null as string | null,
+    loadingScreenImageData: null as ImageData | null,
+    uploadingLoadingScreen: false,
+    resolution: 2048
+  }),
+  CreateThumbnail: async () => {
```
HexaField: camelCase for function names

HexaField (changes requested):

packages/editor/src/services/SceneThumbnailState.tsx
SYBIOTE: done

HexaField:
Looks good, pending discussion https://tsu.atlassian.net/browse/IR-4674?focusedCommentId=44101

HexaField (changes requested):
Rather than it being empty, we should show the existing thumbnail beside the new one (same for envmap) such that we can make a comparison.

Also, it looks like gizmos are rendering into the thumbnail. We need to make sure they are turned off.

packages/editor/src/services/SceneThumbnailState.tsx
```diff
@@ -0,0 +1,126 @@
...
+  uploadThumbnail: async () => {
+    const sceneThumbnailState = getMutableState(SceneThumbnailState)
+    if (!sceneThumbnailState.thumbnail.value) return
+    sceneThumbnailState.uploadingThumbnail.set(true)
+    const editorState = getState(EditorState)
+    const projectName = editorState.projectName!
+    const currentSceneDirectory = getState(EditorState).scenePath!.split('/').slice(0, -1).join('/')
+    const { promises } = uploadProjectFiles(projectName, [sceneThumbnailState.thumbnail.value], [currentSceneDirectory])
+    const [[savedThumbnailURL]] = await Promise.all(promises)
+    commitProperty(SceneSettingsComponent, 'thumbnailURL')(savedThumbnailURL)
```
HexaField: we need to pass the correct entity into this explicitly, otherwise if the entity is not selected, this function crashes

HexaField (changes requested):

packages/editor/src/services/SceneThumbnailState.tsx
SYBIOTE: done

SYBIOTE:
> Rather than it being empty, we should show the existing thumbnail beside the new one (same for envmap) such that we can make a comparison.
> 
> Also, it looks like gizmos are rendering into the thumbnail. We need to make sure they are turned off.

done

HexaField:
Screenshot from 2024-10-10 09-19-50

Let's make these side by side, rather than on top of eachother. Also the next needs to be larger.

SYBIOTE:
> Screenshot from 2024-10-10 09-19-50
> 
> Let's make these side by side, rather than on top of eachother. Also the next needs to be larger.

them being side by side, exceeds the bounds of the modal and initiates scroll behavior
will require a deeper refactor of the image preview primitive

HexaField:
> them being side by side, exceeds the bounds of the modal and initiates scroll behavior will require a deeper refactor of the image preview primitive

Let's add support for that then. It's just not passable vertically stacked.

SYBIOTE:
> > them being side by side, exceeds the bounds of the modal and initiates scroll behavior will require a deeper refactor of the image preview primitive
> 
> Let's add support for that then. It's just not passable vertically stacked.

done

HexaField (changes requested):
few issues
- gizmos appear in thumbnail
- thumbnail takes main camera position, rather than scene preview camera
- previous thumbnail shows the previously generated one, not the currently saved one

Screenshot from 2024-10-24 11-10-52

SYBIOTE:
fixed the save issue, but i have absolutely no idea why the renderer takes up the viewer camera and not the scene preview camera, we are explicitly assigning it in the take screenshot code as well 
@HexaField

HexaField (changes requested):
We definitely need the camera being oriented incorrectly fixed, and also need to ensure the modal is big enough / the previews are small enough that they are properly centered and the modal doesn't have a horizontal scroll.

SYBIOTE:
preview screen shot has stopped working too

SYBIOTE:
made sure that the image is taken form the scene preview camera, made the size of thumbnail modal smaller

HexaField (changes requested):
it still isn't taking the scene preview camera's position

HexaField:
It also seems like when I move the scene preview camera around, the engine camera snaps to it

SYBIOTE:
> It also seems like when I move the scene preview camera around, the engine camera snaps to it

that has always been the case? we dont want that? cool ill fix it

HexaField (changes requested):
Switching scenes does not clear previous state
Screenshot from 2024-11-14 21-05-43

SYBIOTE:
> Switching scenes does not clear previous state Screenshot from 2024-11-14 21-05-43

thanks for pointing that out, implemented the same

HexaField (approved):
Awesome, we got there!

PR #266, opened 2024-09-18
Migrate from CJS to ESM

HexaField:
## Summary

based on https://github.com/ir-engine/ir-engine/pull/148

## References
closes #_insert number here_


package.json
```diff
@@ -45,9 +46,9 @@
     "check-eslint": "eslint --quiet .",
     "checkout-dev": "lerna exec 'git checkout dev' --parallel --no-bail",
     "clean-node-modules": "npx rimraf node_modules && npx rimraf package-lock.json && npx lerna exec npx rimraf node_modules && npx lerna exec npx rimraf package-lock.json",
-    "clone-project": "cross-env ts-node --swc scripts/clone-project.ts",
-    "create-root-package-json": "cross-env ts-node --swc scripts/create-root-package-json",
-    "create-project": "cross-env ts-node --swc scripts/create-project",
+    "clone-project": "cross-env vite-node scripts/clone-project.ts",
+    "create-root-package-json": "cross-env vite-node scripts/create-root-package-json",
+    "create-project": "cross-env vite-node scripts/create-pr  oject",
```
heysokam: Is this `create-pr  oject` part intentional or typo?

DanielBelmes: Does not look intentional :)

HexaField: Nope. Don't know how that happened.

packages/client-core/package.json
```diff
@@ -58,7 +59,7 @@
     "i18next": "21.6.16",
     "image-palette-core": "0.2.2",
     "jszip": "^3.10.1",
-    "lodash": "4.17.21",
+    "lodash-es": "4.17.21",
```
SamMazerIR, line 62: did we switch to a Spanish language version?

DanielBelmes:
Looks good. Is there plans to fix the failing test?

HexaField:
They started failing all of a sudden. Have to look at the merged commits to see why. :(

packages/engine/src/scene/systems/SceneObjectSystem.tsx
```diff
@@ -159,7 +156,6 @@ export function setupObject(obj: Object3D, entity: Entity, forceBasicMaterials =
 }
 
 const groupQuery = defineQuery([GroupComponent])
-const updatableQuery = defineQuery([UpdatableComponent, CallbackComponent])
```
DanielBelmes: Did you provide an alternative for this(updateableComponent)? We are currently using this in https://github.com/theinfinitereality/irpro-advanced-materials/ for animating materials for example the WaterStylizeMaterial and so this branch is breaking that functionality.

HexaField: You can just define a system. Pretty easy to fix.

DanielBelmes: I believe the use case is that each individual material instance has it's own callback function. @dinomut1 first used useExecute I believe but was told not to democratically. But if it's gonna be removed it would be good to fix the dependency in the update PR for advanced materials. But personally I'd just revert the removal as it's a useful paradigm for now.

HexaField: we can just query for a component and call whatever updates we want. thats the point of an ECS. it's tech debt and not a pattern we should keep

dinomut1: Not in scope to remove updateable component as part of a PR to migrate from cjs to esm. We do not have an agreed upon replacement for advanced materials, which are not components and thus cannot be queried.

DanielBelmes: I don't believe this change is related to this ticket and is being done in an incorrect way. Historically we try to ensure a PR that is merged in won't break deployments. In its current state any deployment with our core packages installed would break. The best answer is to revert this change and to record a ticket for this work to be done seperatly when we have more bandwidth. Then we can remove this tech debt and refactor its many dependencies at the same time.

dinomut1 (dismissed):
Need to keep updateable component for now. Not in scope of PR

barankyle (approved):

PR #278, opened 2024-09-19
IR-4605: Add "save" and "revert" buttons in hierarchy panel

dinomut1:
Closes
https://tsu.atlassian.net/browse/IR-4605

When the contents of a model file are modified, "save" and "revert" buttons appear in the hierarchy panel which easily let the user save or undo their changes

image


HexaField:
so awesome

dinomut1:
expediting this merge as artists are having problems with save workflow for making templates

aditya-mitra (approved):
Made a few changes to match our designs
Looks good!

PR #280, opened 2024-09-20
Michael/ir 4091 video

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-4091
https://tsu.atlassian.net/browse/IR-4664
## References
closes #_insert number here_


packages/engine/src/scene/components/VideoComponent.tsx
```diff
@@ -112,13 +112,36 @@ export const VideoComponent = defineComponent({
...
     fit: ContentFitTypeSchema('contain'),
+    useLetterBox: S.Bool(false),
     projection: ProjectionSchema,
     mediaUUID: S.EntityUUID(),
     // internal
     videoMeshEntity: S.Entity(),
     texture: S.Nullable(S.Type<VideoTexturePriorityQueue>())
   }),
 
+  onSet: (entity, component, json) => {
```
HexaField: i dont think we need the onset anymore

HexaField (changes requested):

HexaField (dismissed):
this is what the `fit` field should already accomplish. we should avoid making new fields and functions if `fit` is sufficient

packages/engine/src/scene/components/VideoComponent.tsx
MbfloydIR: Talked with @MichaelEstes, the onSet is temporarily needed, other wise the video player doesn't work.

MbfloydIR:
> this is what the `fit` field should already accomplish. we should avoid making new fields and functions if `fit` is sufficient

Removed the letterbox property and made the letterboxing effect always on

packages/engine/src/scene/components/VideoComponent.tsx
MichaelEstes: Michael found an issue with component schemas for components that rely on three objects being cloned when set. The onSet can be removed when https://github.com/ir-engine/ir-engine/pull/285 is merged in

MbfloydIR: on set has now been removed

Rezmason (approved):
LGTM!

PR #285, opened 2024-09-20
Component schema initializer / reuse three objects in onSet

MichaelEstes:
## References
closes #_insert number here_


dinomut1 (approved):
Fixes critical issues with scene changes randomly rotating objects, and sometimes entities teleporting far away

PR #291, opened 2024-09-23
Tcr/sample data test id and test

TCRodriguez:
## Summary
This PR brings in `data-testid` attributes to the front-end for the `studio`: specifically, the `Files` and `Assets` panels.

It also brings in a new test file to assert that these elements get rendered and thus ensures that any automation script written from here on can rely on these test IDs to remain in place (generally speaking).

## Subtasks Checklist
N/A

## Breaking Changes
N/A

## References
closes [IR-4738](https://tsu.atlassian.net/browse/IR-4738)

## QA Steps
N/A

[IR-4738]: https://tsu.atlassian.net/browse/IR-4738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/editor/tests/setupTestDependencies.ts
```diff
@@ -0,0 +1,43 @@
...
+  fallbackLng: 'en',
+  resources: {
+    en: {
+      translation: {}
+    }
+  },
+  interpolation: { escapeValue: false }
+})
+
+HyperFlux.store = <HyperStore>{
```
HexaField: why are we creating a hyperflux store here? the test setup context does not need it. also, the store is created with `createHyperStore`

packages/ui/src/primitives/tailwind/Slider/index.tsx
```diff
@@ -35,12 +35,25 @@ export interface SliderProps {
...
+const Slider = ({
+  value,
+  min = 0,
+  max = 100,
+  step = 1,
+  width = 200,
+  onChange,
+  onRelease,
+  className,
+  dataTestIdForValueInput,
```
HexaField: i'd very much prefer for us not to include these test IDs in our primitive library. can we keep them upstream in the UI?

HexaField (changes requested):

HexaField:
Please add the Jira ticket number

packages/editor/tests/setupTestDependencies.ts
TCRodriguez: When I ran test without `HyperFlux.store` I get the following error:
```
     TypeError: Cannot read properties of undefined (reading 'stateMap')
```

That being said, I myself may be missing some knowledge on how to properly mock/stub things in unit tests like this, so any guidance would be very much appreciated! 🙏🏽

HexaField: You need to run either `createHyperStore` or `createEngine` (depending what you're testing) inside the `before` hook of the test

aditya-mitra:
cross posting [this](https://theinfinitereality.slack.com/archives/C070ZGB7Q9E/p1727154258636109?thread_ts=1727137611.518249&cid=C070ZGB7Q9E) here

can we do testing on element existence on e2e skipping unit tests?
specifically unit tests are likely to change and harder to maintain on the ui side due to their changing nature. also when we will be doing e2e tests, we will be checking for existence. component and e2e tests should be good enough for both studio and locations imo

TCRodriguez:
> can we do testing on element existence on e2e skipping unit tests? specifically unit tests are likely to change and harder to maintain on the ui side due to their changing nature. also when we will be doing e2e tests, we will be checking for existence. component and e2e tests should be good enough for both studio and locations imo

@aditya-mitra For now, we can definitely omit the `test.tsx` files to expedite the process of adding the `data-testid` attributes. I know there is a significant UI refactor currently happening so it's likely any UI component tests would have to change significantly as well.

However, I still believe that, eventually, component tests (rather than Playwright scripts) should be responsible for verifying the existence of these attributes. My concern is that if we rely solely on Playwright tests, we may not catch missing or removed `data-testid` attributes until a Playwright test fails:

	1.	UI components are refactored or altered, and changes are merged into dev.
	2.	A Playwright test that relies on the `data-testid` attribute fails because the locator is broken.
	3.	Someone then has to open a new branch, add the missing `data-testid` attributes (duplicating work), and wait for the PR to be merged before the Playwright tests pass again.

In cases where there are significant UI changes (e.g., replacing a slider with a checkbox or changing an `<input>` to a `<select>`), it’s understandable that associated Playwright flows would need reworking as well as any documented testing steps in the associated test cases.

But for simpler changes, like moving a file to another folder (e.g., moving `index.tsx` from `packages/ui/src/components/Assets/container` to `packages/editor/src/panels/assets`), the `data-testid` attributes, if already in place, should remain unaffected.

As for your concern on mitigating the overhead of maintaining the UI tests, I completely understand, but I hope I've made the case here that the work investment would be worth it. I myself would be more than willing to take help take ownership of maintaining these tests. Also, unless I'm mistaken, there is an acknowledged gap in the UI component tests specifically around the studio panels and elsewhere in the `editor` package? So these tests will get added eventually, and it'd be quick to just include some form of `data-testid` existence check there.

This is likely fit for a longer discussion, and perhaps a better place for this to get documented 😅 . 

Tagging @HexaField and @heysokam and @hanzlamateen for reference.

packages/editor/tests/setupTestDependencies.ts
TCRodriguez: Just to close this thread, I'll be removing the `toolbar.test.tsx` entirely for now, and plan to come back to this in the near future. Refer to my comment [here](https://github.com/ir-engine/ir-engine/pull/291#issuecomment-2375814339).

I will keep this in mind, however, for when I revisit this!

packages/ui/src/primitives/tailwind/Slider/index.tsx
TCRodriguez: @HexaField So I went ahead and found another way to do this by "chaining" off of a more "parent" `data-testid` attribute like so:
```
  await page.getByTestId('files-panel-view-options-list-font-size-group').locator('input').first().fill('85');
  await page.getByTestId('files-panel-view-options-list-font-size-group').locator('input').nth(1).fill('54');
```

This test ID is now on the `<InputGroup>` component, which is defined in `packages/ui/src/components/editor/input/Group/index.tsx` and called in `packages/editor/src/panels/files/toolbar.tsx`.  

While it definitely works from our Playwright perspective, I still feel that in this case I'd want to be able to use the more robust `getByTestId()` method with a granular `data-testid` on the `<input>` itself since using the `nth()` locator relies on the _position_ of the elements to remain the same and so is much more subject to change as the UI gets worked on. We also lose context when working in the automation script as well as readability:

`await page.getByTestId('files-panel-view-options-list-font-size-group').locator('input').nth(1).fill('54')`

vs.

`await page.getByTestId('files-panel-view-options-list-font-size-slider-input').fill('54')`


Now, I know in this case I'm really splitting hairs with my argument, but I figured we're discussing how we move forward -- there are many more `data-testid` attributes that are going to get added so its best to nail the procedure down now.

Either way, let me know what you think 🙏🏽 .

HexaField:
Something went wrong with your merge strategy, please rebase and fix so the diff is correct

packages/ui/src/primitives/tailwind/Slider/index.tsx
HexaField: What if we did a middleground by adding generic IDs to the ui primitives, so we don't have to resort to magic numbers, but also don't have to pass in IDs explicitly.

TCRodriguez: If I'm understanding correctly, is this what you mean?

```
      <input
        min={min}
        max={max}
        value={value}
        onChange={handleInputChange}
        onBlur={() => onRelease && onRelease(value)}
        className="h-8 w-14 rounded bg-neutral-900 text-center text-sm font-normal leading-[21px] text-neutral-400"
        data-testid="slider-text-value-input"
      />
```
      
     
So `data-testid` is now `slider-text-value-input` instead of `data-testid={dataTestIdForValueInput || ''`? 
      



      
This way, we don't include them as props here:
      
```
const Slider = ({
  value,
  min = 0,
  max = 100,
  step = 1,
  width = 200,
  onChange,
  onRelease,
  className,
  dataTestIdForValueInput, // here
  dataTestIdForSlider // here
}: SliderProps) => {

```

?

HexaField: Yup, so we can query for the higher level component, and then query into it for the desired inner element.

TCRodriguez:
Closing this branch to address the `verification/cla-signed` issue.

PR #294, opened 2024-09-23
The file browser now always shows the scroll bar if its contents vertically overflow.

Rezmason:
closes https://tsu.atlassian.net/browse/IR-4697 .


aditya-mitra (approved):
LGTM ✅

PR #295, opened 2024-09-23
Box selection

JT00y:
## Summary
make box selection to drag a marquee to select entities in viewport
## References
closes #_insert number here_


packages/spatial/src/camera/systems/CameraOrbitSystem.tsx
```diff
@@ -37,21 +37,21 @@ import {
...
 } from '@ir-engine/ecs'
-import { getState, isClient } from '@ir-engine/hyperflux'
+import { getMutableState, getState, isClient } from '@ir-engine/hyperflux'
 import { TransformComponent } from '@ir-engine/spatial'
 import { CameraComponent } from '@ir-engine/spatial/src/camera/components/CameraComponent'
 import { CameraOrbitComponent } from '@ir-engine/spatial/src/camera/components/CameraOrbitComponent'
 import { Vector3_Up } from '@ir-engine/spatial/src/common/constants/MathConstants'
 import { GroupComponent } from '@ir-engine/spatial/src/renderer/components/GroupComponent'
 
+import { SelectionBoxState } from '@ir-engine/editor/src/panels/viewport/tools/SelectionBoxTool'
```
DanielBelmes: Couple things. We can't import editor into spatial or it will be a circular dependency. I think what you are trying to do here is that if the tool is selected orbiting won't run. What you want to use in editor is InputState.setCapturingEntity(YourEntity) We can also consider other helper InputState functions. But this would remove the need to modify code in spatial like you have and then that test will stop failing.

DanielBelmes (changes requested):

packages/spatial/src/input/state/InputState.ts
```diff
@@ -37,7 +37,8 @@ export const InputState = defineState({
     scroll: new Vector2(),
     capturingEntity: UndefinedEntity,
     inputMeshes: new Set([] as Entity[]),
-    inputBoundingBoxes: new Set([] as Entity[])
+    inputBoundingBoxes: new Set([] as Entity[]),
+    capturingCameraOrbitEnbaled: true
```
heysokam: `Enbaled` typo :writing_hand:

packages/editor/src/panels/viewport/tools/SelectionBoxTool.test.tsx
dinomut1, line 1: Please add additional test for the box not selecting an object that is outside the frustum, and a test for selecting multiple objects

dinomut1 (changes requested):

dinomut1 (approved):

DanielBelmes (approved):

PR #307, opened 2024-09-24
The buttons in a modal dialog should be right aligned, and populate from right to left

Rezmason:
Closes [IR-4580](https://tsu.atlassian.net/browse/IR-4580)

[IR-4580]: https://tsu.atlassian.net/browse/IR-4580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aditya-mitra (changes requested):
image


this affects other modals

CITIZENDOT:
Something like this will work. If two buttons exist, they would be positioned at their respective ends. If only one button exists, it would be centered.

https://codepen.io/CITIZENDOT/pen/NPKQvXa

PR #308, opened 2024-09-24
Fix removing particle behavior

Rezmason:
Removing a particle behavior by splicing at its index, rather than filtering it out by value. This also assures that only one behavior is removed, even if there are several identical behaviors.

Closes [IR-4694](https://tsu.atlassian.net/browse/IR-4694).

[IR-4694]: https://tsu.atlassian.net/browse/IR-4694?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aditya-mitra (approved):
Looks good ✅

PR #310, opened 2024-09-24
Fix contextmenu file selection

Rezmason:
If the file browser's context menu is opened while files are already selected, it will add the current file to the selection if it isn't already included. This membership test needs to use `Array::some` for references to objects, not `Array::includes`.

Closes [IR-4216](https://tsu.atlassian.net/browse/IR-4216).

[IR-4216]: https://tsu.atlassian.net/browse/IR-4216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/editor/src/panels/files/contextmenu.tsx
+++ b/packages/editor/src/panels/files/contextmenu.tsx
@@ -225,7 +225,7 @@ export function FileContextMenu({
           size="small"
           fullWidth
           onClick={() => {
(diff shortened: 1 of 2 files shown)


aditya-mitra (approved):
LGTM

PR #312, opened 2024-09-25
IR-4714 Merge coil setting to engine settings

mshafiqmk:
## Summary

Description: This PR migrates the Coil settings from the **coil-setting** table to the **engine-setting** table. The following changes have been made:

### Migration Script:

Created a new migration script to move existing data from **coil-setting** to **engine-setting**.
The migration script also drops the coil-setting table after the data has been moved.

### Code Update:

Updated **updateAppConfig.ts** to read and write Coil settings from the engine-settings table instead of the coil-setting table.

### Seed File Update:
Updated the **engine-setting** seed file to insert  Coil settings into the engine-settings table.

## References
closes #_insert number here_


barankyle (approved):

packages/client/scripts/getCoilSettings.ts
```diff
@@ -42,11 +42,17 @@ export const getCoilSetting = async () => {
 
   const coilSetting = await knexClient
     .select()
-    .from<CoilSettingType>(coilSettingPath)
-    .then(([dbCoil]) => {
-      if (dbCoil) {
-        return dbCoil
+    .from<EngineSettingType>(engineSettingPath)
```
hanzlamateen, line 45: We need to update this file to be getEngineSetting and then make it get only those keys which are required from where this method is being called.

hanzlamateen (changes requested):

packages/server-core/src/setting/engine-setting/engine-setting.seed.ts
```diff
@@ -93,3 +84,20 @@ export async function seed(knex: Knex): Promise<void> {
     }
   }
 }
+
+async function generateSeedData(
+  items: { key: string; value: string }[],
+  category: EngineSettingType['category'],
+  type: EngineSettingType['type']
```
hanzlamateen: Maybe you can make the default `type` here as private?

hanzlamateen (changes requested):

packages/client/scripts/getCoilSettings.ts
mshafiqmk: @hanzlamateen done.

packages/server-core/src/setting/engine-setting/engine-setting.seed.ts
mshafiqmk: @hanzlamateen done

hanzlamateen (approved):

PR #313, opened 2024-09-25
new designs for checkbox

aditya-mitra:
## Summary

figma link: https://www.figma.com/design/ln2VDACenFEkjVeHkowxyi/iR-Engine-Design-Library-File?node-id=2786-21102&node-type=frame&t=TlQtKBH49KjD5Efr-0

## References

closes https://tsu.atlassian.net/browse/IR-4752

## QA Steps

image


packages/ui/src/primitives/tailwind/Checkbox/index.tsx
```diff
@@ -23,60 +23,112 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
-      <HiCheck onClick={handleChange} className="absolute m-0.5 hidden h-3 w-3 text-white peer-checked:block" />
+        tabIndex={0}
+        ref={ref}
+      >
+        <FaCheck
+          onClick={handleChange}
+          className={twMerge(
+            'absolute h-3 w-3 transition-transform duration-200 ease-in-out',
+            disabled ? 'cursor-not-allowed text-[#42454D]' : 'text-[#5F7DBF]',
+            isCheckedInternal ? 'scale-100' : 'scale-0'
```
aditya-mitra: we can use `checked && !indeterminate` instead of introducing another state

packages/ui/src/primitives/tailwind/Checkbox/index.tsx
```diff
@@ -23,60 +23,112 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
+          )}
+        />
+
+        <FaMinus
+          onClick={handleChange}
+          className={twMerge(
+            'absolute h-3 w-3 transition-transform duration-200 ease-in-out',
+            disabled ? 'cursor-not-allowed text-[#42454D]' : 'text-[#5F7DBF]',
+            isIndeterminateInternal ? 'scale-100' : 'scale-0'
```
aditya-mitra: same here

packages/ui/src/primitives/tailwind/Checkbox/index.tsx
CITIZENDOT: internal state is used for following purpose:
- when indeterminate state is set to true: checked should be false (even if props.checked is set to true)
- when checked is set to true: indeterminate should be false (even if props.indeterminate is set to true)

so, newly set property takes precedence and overrides the old property. this logic is managed by internal states.

packages/ui/src/primitives/tailwind/Checkbox/index.tsx
CITIZENDOT: https://github.com/ir-engine/ir-engine/pull/313#discussion_r1782166449

CITIZENDOT (approved):

PR #317, opened 2024-09-25
v0.8.1 NO MERGE WITHOUT LARRY Fix memory leak from useAncestorWithComponents

DanielBelmes:
## Summary
https://tsu.atlassian.net/browse/IR-4761
https://tsu.atlassian.net/browse/IR-4734?atlOrigin=eyJpIjoiYTZlMjllMWUwYzlkNDI0Y2E4M2E0MTYwYzVlODlkYzMiLCJwIjoiamlyYS1zbGFjay1pbnQifQ

## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/transform/components/EntityTree.tsx
+++ b/packages/spatial/src/transform/components/EntityTree.tsx
@@ -423,7 +423,7 @@ export function useAncestorWithComponents(
       unmounted = true
       root?.stop()
     }
-  }, [entity, componentsString, includeSelf, parentEntity])
+  }, [entity, componentsString, includeSelf, parentEntity?.value])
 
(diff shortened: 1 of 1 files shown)


packages/spatial/src/transform/components/EntityTree.tsx
```diff
@@ -423,7 +423,7 @@ export function useAncestorWithComponents(
       unmounted = true
```
DanielBelmes, line 423: I also wonder why this memory wasn't cleaning. Maybe there is something with StartReactor that we need to clean up. @HexaField @MichaelEstes

HexaField (approved):

PR #318, opened 2024-09-25
Fix: storage permission query

jose-galvan:
## Summary
The `navigator.permissions.query()` method throws an error when a browser does not support the requested permission. This behavior is observed in Safari 17.5, causing the app to hang at the loading screen.

This update introduces error handling for unsupported permission queries, ensuring the app continues functioning smoothly without getting stuck.

**See [TypeError](https://developer.mozilla.org/en-US/docs/Web/API/Permissions/query#exceptions)**

## References
closes #_insert number here_

### DIFF
--- a/packages/client/public/root-cookie-accessor-template.html
+++ b/packages/client/public/root-cookie-accessor-template.html
@@ -58,7 +58,7 @@
               const hasAccess = await document.hasStorageAccess()
               let storageAccessPermission
               try {
-                await navigator.permissions.query({
+                storageAccessPermission = await navigator.permissions.query({
                   name: 'storage-access'
                 })
               } catch (error) {
--- a/packages/client/public/root-cookie-accessor-template.html
+++ b/packages/client/public/root-cookie-accessor-template.html
@@ -56,9 +56,14 @@
               break
             } else {
               const hasAccess = await document.hasStorageAccess()
-              const storageAccessPermission = await navigator.permissions.query({
-                name: 'storage-access'
-              })
+              let storageAccessPermission = { state: 'denied' } 
+              try {
+                await navigator.permissions.query({
+                  name: 'storage-access'
+                })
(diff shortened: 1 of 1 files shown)


barankyle (approved):

packages/client/public/root-cookie-accessor-template.html
```diff
@@ -56,9 +56,14 @@
               break
             } else {
               const hasAccess = await document.hasStorageAccess()
-              const storageAccessPermission = await navigator.permissions.query({
-                name: 'storage-access'
-              })
+              let storageAccessPermission = { state: 'denied' }
```
hanzlamateen: I am just wondering why are we setting it as `denied` towards the beginning? Should it be set as `denied` in catch block? Otherwise use the response of `navigator.permissions.query` method?

packages/client/public/root-cookie-accessor-template.html
```diff
@@ -56,9 +56,15 @@
               break
             } else {
               const hasAccess = await document.hasStorageAccess()
-              const storageAccessPermission = await navigator.permissions.query({
-                name: 'storage-access'
-              })
+              let storageAccessPermission
+              try {
+                await navigator.permissions.query({
```
barankyle: I think we need to attempt to set storageAccessPermission with the result of this, otherwise if it succeeds, storageAccessPermission will still be null and the if statement on line 68 will never get triggered properly.

barankyle (changes requested):

packages/client/public/root-cookie-accessor-template.html
jose-galvan: Apologies for missing that earlier. That was my intention from the start.

barankyle (approved):

hanzlamateen (approved):

PR #320, opened 2024-09-26
fix(IR-4586): showing error on profile username when is greater than 60 characters

kattsushi:
…

## Summary
fix(IR-4586): showing error on profile username when is greater than 60 characters

image

## References
Jira issue [#IR-4586](https://tsu.atlassian.net/browse/IR-4586)


packages/client-core/i18n/en/user.json
```diff
@@ -192,6 +192,7 @@
       "lbl-settings": "Settings",
       "lbl-username": "Username",
       "usernameError": "Invalid username",
+      "usernameLengthError": "Username is too long",
```
jonathan-casarrubias: I suggest to be more explicit and to provide the user with feedback that indicates how to avoid the problem, example:

`Username must have less than X characters`

jonathan-casarrubias (changes requested):

packages/client-core/i18n/en/user.json
kattsushi: good point! already addresed ty

image

jonathan-casarrubias (approved):

packages/client-core/src/user/components/UserMenu/menus/ProfileMenu.tsx
```diff
@@ -74,6 +74,7 @@ import styles from '../index.module.scss'
 const termsOfService = config.client.tosAddress ?? '/terms-of-service'
 
 const logger = multiLogger.child({ component: 'engine:ecs:ProfileMenu', modifier: clientContextParams })
+const USERNAME_MAX_LENGTH = 60
```
aditya-mitra: is there a backend validation for it as well?
we should add it as well

aditya-mitra (changes requested):

packages/client-core/src/user/components/UserMenu/menus/ProfileMenu.tsx
kattsushi: done ty

mikeplascdev (approved):

hanzlamateen (approved):

PR #328, opened 2024-09-26
Implement Camera Gizmo

SYBIOTE:
Implements a camera gizmo which allows for easy manipulation of the scene camera


DanielBelmes:
Still not working when I have an object selected but otherwise looks really good.

DanielBelmes:
still have to review code

SYBIOTE:
> Still not working when I have an object selected but otherwise looks really good.

yeah the transform gizmo's input component, takes over the camera gizmo's i don't quite understand why that happens

SYBIOTE:
> > Still not working when I have an object selected but otherwise looks really good.
> 
> yeah the transform gizmo's input component, takes over the camera gizmo's i don't quite understand why that happens

yup absolutely no idea what's happening, i understand that the presence of the transform gizmo is somehow messing with the input of the camera gizmo but i have no idea why
@DanielBelmes

SamMazerIR:
pushed a fix for input heuristics. should be good now regardless of whether you're selecting an object

SamMazerIR (approved):

HexaField (changes requested):
when i click the camera preview entity the scene is scaled up

SYBIOTE:
> when i click the camera preview entity the scene is scaled up

i dont see the issue on my end

SYBIOTE:
@HexaField , any update on this

HexaField:
> @HexaField , any update on this

Try the sky station. It might have something to do with the scene preview camera being disabled. All you need to do to reproduce is open the sky station in the studio and select the scene preview camera entity.

SYBIOTE:
seems like a separate issue? the zooming happens when you select the scene preview camera when its invisible, if you make the scene preview camera visible again, it will revert back to appropriate zoom level

SYBIOTE:
the problem exists in dev as well, can we solve the bug separately from this PR?
@HexaField

HexaField:
> the problem exists in dev as well, can we solve the bug separately from this PR?
> @HexaField 

Gotcha. Yes, no worries then.

HexaField (approved):

PR #336, opened 2024-09-27
IR-4693 remove beziers issue fix

SamMazerIR:
## Summary
fixing issue with PaginatedList that caused no feedback when adding or removing Beziers from the ParticleSystemComponent. Also fixed the twin PaginatedList (there are two, both being used in different elements that end up being used in ParticleSystem but not directly in the same place)

we apparently have two PaginatedList.tsx files in the engine, and ParticleSystem's node editor is using one at the root level as well as ValueGenerator which uses the OTHER PaginatedList. The only difference appears to be the HTML return value of the elements, one using a Well and Grid while the other is mostly just divs. I'm sure there's a styling reason for both but I suggest we modify their names to differentiate them better. Would like your input as to what to rename one or both.

ir-engine/packages/editor/src/components/layout/PaginatedList.tsx

-  ONLY used in ParticleSystemNodeEditor, twice

ir-engine/packages/ui/src/components/editor/layout/PaginatedList.tsx
- used in ValueGenerator (ir-engine/pakcages/ui/src/components/editor/input/Generator/Value/index.tsx)and several other places but NOT in ParticleSystemNodeEditor directly (used in Elements that it DOES use though, hence why I missed this)

## References
[https://tsu.atlassian.net/browse/IR-4693](https://tsu.atlassian.net/browse/IR-4693)

## QA Steps
go to a particle system, scroll until you see start life/start size or any of those with a "type" set to interval by default
image

change it to type bezier
add beziers and remove them with the buttons that appear, notice you can add enough to add a new page, scroll pages with the arrow keys or the page buttons that populate


Rezmason (approved):
One `PaginatedList` should be enough for everybody.

PR #339, opened 2024-09-27
rename skybox from ee_skybox to ir_skybox

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-4782
## References
closes #_insert number here_


packages/projects/default-project/public/scenes/apartment.gltf
```diff
@@ -262,7 +262,7 @@
       }
     },
     {
-      "name": "EE_skybox",
+      "name": "iR_skybox",
```
HexaField: these should all just be `Skybox`

HexaField (changes requested):
This needs to respect the

packages/projects/default-project/public/scenes/apartment.gltf
MbfloydIR: implemented

HexaField (approved):

PR #346, opened 2024-09-30
`data-testid`s for studio

TCRodriguez:
## Summary
This PR brings in `data-testid` attributes to the front-end for the `studio`: specifically, the `Files` and `Assets` panels.

It also brings in a new test file to assert that these elements get rendered and thus ensures that any automation script written from here on can rely on these test IDs to remain in place (generally speaking).

## Subtasks Checklist
N/A

## Breaking Changes
N/A

## References
closes [IR-4738](https://tsu.atlassian.net/browse/IR-4738)

## QA Steps
N/A

[IR-4738]: https://tsu.atlassian.net/browse/IR-4738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TCRodriguez:
@cla-bot

TCRodriguez:
@cla-bot check

packages/editor/src/panels/files/fileitem.tsx
```diff
@@ -139,11 +139,12 @@ function TableView({ file, onClick, onDoubleClick, drag, drop, isOver, onContext
       onContextMenu={onContextMenu}
       onClick={onClick}
       onDoubleClick={onDoubleClick}
+      data-testid="files-panel-file-item"
     >
       {availableTableColumns
         .filter((header) => selectedTableColumns[header])
         .map((header, idx) => (
-          <td key={idx} style={{ fontSize: `${fontSize}px` }}>
+          <td key={idx} style={{ fontSize: `${fontSize}px` }} data-testid={`files-panel-file-item-${header}`}>
```
aditya-mitra, line 147: how about using `idx` instead of `header`?
`idx` is more non-changing

packages/editor/src/panels/files/toolbar.tsx
```diff
@@ -146,7 +160,10 @@ const ViewModeSettings = () => {
             </div>
             <div className="flex-col">
               {availableTableColumns.map((column) => (
-                <InputGroup label={t(`editor:layout.filebrowser.table-list.headers.${column}`)}>
+                <InputGroup
+                  label={t(`editor:layout.filebrowser.table-list.headers.${column}`)}
+                  dataTestId={`files-panel-view-mode-list-options-column-${column}`}
```
aditya-mitra, line 165: same here. lets use indices

packages/ui/src/primitives/tailwind/Slider/index.tsx
aditya-mitra, line 1: i think we avoid testing on the the ui primitives because they are base level components. using them, makes data-testid non unique

aditya-mitra (changes requested):
most of the editor changes look good.
we need to revert the changes on ui folder

packages/ui/src/primitives/tailwind/Slider/index.tsx
TCRodriguez: @aditya-mitra Ah yes, this is actually something I've discussed with @HexaField on the old PR [here](https://github.com/ir-engine/ir-engine/pull/291#discussion_r1772094599). We came to the middle ground of placing generic `data-testid` attributes on the primitives instead of specific ones, that way we can properly "chain" into them through Playwright's locator engine from a parent component. If you look at the comment thread, you'll see the examples I mention that supports the need to put them here.

But to summarize, without some kind of descriptive attribute to hook into we'd need to rely on the `nth()` locator feature which relies on the position of the element, which isn't reliable long-term.

packages/editor/src/panels/files/fileitem.tsx
TCRodriguez: @aditya-mitra Ah yes, that's a good point! I'll make the change.

TCRodriguez: @aditya-mitra So for this, the reason I went with `header` is so that the `data-testid` is descriptive. If we use `idx`, we'd have to map out what each index position corresponds to -- which isn't that big of a deal, but if we can avoid that I'd prefer that. 

This is what we get when we use `idx` instead of `header`:
image

As you can see in the screenshot, `files-panel-file-item-0` doesn't tell us much. However, if we use `header`:
image

Now, we know exactly the element we're getting, which in this case is the **name** of the "file item". And this would repeat for the rest of the columns (`Type`, `Date Modified`, etc.)

packages/editor/src/panels/files/toolbar.tsx
TCRodriguez: See my comment above.

packages/editor/src/panels/files/fileitem.tsx
aditya-mitra: awesome explanation with visuals!
got it. we can keep headers then

aditya-mitra (approved):
Looks good ✅

PR #349, opened 2024-09-30
add underscores in filename validation

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-4621
## References
closes #_insert number here_


packages/common/tests/regex.test.ts
```diff
@@ -48,7 +48,6 @@ describe('regex.test', () => {
   describe('INVALID_FILENAME_REGEX', () => {
     it('should not match invalid filenames', () => {
       const invalidFilenames = [
-        'hello_world',
```
aditya-mitra, line 51: can you also a add a test with `_` in the _should match valid filenames_ test?

aditya-mitra (dismissed):

packages/common/tests/regex.test.ts
Rezmason: I'll do it 😎

Rezmason: Added https://github.com/ir-engine/ir-engine/pull/349/commits/32225efcc4dbdb579a6a231601621ab744a59733

Rezmason: Apparently `_` should not be a valid filename. Also, `example.com` should also be invalid, but it isn't. Investigating.

Rezmason: with insight offered by David and Kyle, we’ll support the following conditions for filenames.

A filename is valid if:
- it has 4 to 64 characters
- its first and last character are alphanumeric
- any other characters are either alphanumeric, “_”, “-”, or “.”

barankyle (approved):

PR #352, opened 2024-10-01
Fixed issues with invites

barankyle:
## Summary

Instance invite menu would throw errors if one tried to send an invite before connected to an instance. Button to open menu is now disabled until one has connected to an instance.

accept-invite would error out when creating a new identity-provider because of better enforcement of validators. params.query.passcode was being used by identity-provider.create but was not a valid query param for that process. Deleting params.query.passcode after it is no longer needed.

Added alternative placeholder text to invite menu's email/phone input based on which magiclink services are enabled.

Resolves IR-4321 and IR-4846

## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/i18n/en/user.json
+++ b/packages/client-core/i18n/en/user.json
@@ -125,7 +125,11 @@
     "share": {
       "lbl-share": "Share",
       "ph-phoneEmail": "Invite by email or phone number",
+      "ph-email": "Invite by email",
+      "ph-phone": "Invite by phone number",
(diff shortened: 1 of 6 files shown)


larryatinfinite:
@barankyle QA has cleared INT so we can continue testing next releases features since the build has progressed to UAT. For this, can you base this change off of MT-STG instead? https://github.com/ir-engine/ir-engine/pull/352

barankyle:
I've changed the base to the stg branch.

HexaField (approved):

PR #363, opened 2024-10-03
[IR-4892] tst: Refactor our custom assertion functions into one file

heysokam:
## References
https://tsu.atlassian.net/browse/IR-4892


heysokam:
@HexaField Should be good to go :+1:

We definitely want to port this functionality to [`vitest/expect.extend`](https://vitest.dev/guide/extending-matchers.html) at some point,
but this should work in the meantime.

HexaField (approved):

PR #372, opened 2024-10-04
IR-4351 positional audio upgrades

SamMazerIR:
## Summary
positional audio node editor, helper and descriptions all updated. Gizmo now updates reactively for both inner and outer angle for directional audio rolloff and max distance (in linear mode)

Inverse and exponential will not match the distance of this gizmo but the directionality of the angles will still matter.

updated descriptions in the help hover text for each field for clarity, especially around min/max angle.

removed ref distance as it should basically always be set to 1
updated other default values (angles default to 360 per docs)
made min always smaller or equal to max angle and vice versa

image

IMPORTANT:
- The volume/distance effects cannot be heard in editor play/edit mode!!!!! - you must be in a published location!!! 
- the audio gizmo (turn on gizmos with debug tool or editor menu) will show the cone angles
image
image

- for a better visualization of distance rolloff, please play with this graphing tool [distance rolloff models](https://www.desmos.com/calculator/lzxfqvwoqq)
- audio position is based off the CAMERA position, not the avatar, keep this in mind when testing angles/distance

- within the min angle cone there is no directional reduction in volume
- between the edge of the inner and outer angle cone, the volume will linearly change from 1 at the inner edge to the value set in ConeOuterGain (default 0), after outside the outerAngle it will stay at this value

## References
[https://tsu.atlassian.net/browse/IR-4351](https://tsu.atlassian.net/browse/IR-4351)


HexaField (dismissed):
- Helper being visible should not be dependent on the media element having a source

- The max view that dictates the size of the helper is not affected when changing from linear to another mode, and the helper does not reflect those other modes' distances

HexaField (commented):
btw i think we can also remove `PositionalAudioHelper.ts` it was deprecated a long time ago, and i think it's no longer used

MbfloydIR (approved):

PR #379, opened 2024-10-07
IR-4934: Add Loading Geometry to 3D Models

dinomut1:
https://tsu.atlassian.net/browse/IR-4934

https://tsu.atlassian.net/browse/IR-4948


packages/engine/src/scene/components/ModelComponent.tsx
```diff
@@ -211,6 +231,9 @@ function ModelReactor() {
         animations: scene.animations
       })
     }
+
+    removeLoadingGeo()
```
MichaelEstes, line 236: Super small, but if the file loader errors (404 or some server error), this will never be hit

dinomut1: added a remove function in the error useEffect

MichaelEstes (approved):

PR #381, opened 2024-10-07
Expanded cases where useGet auto-returns an error

barankyle:
## Summary

useGet was only skipping making a request if passed no args. But if passed 'null' or an empty string, it would make the request, which would usually throw an error. This makes those situations also auto-merge an error without making the GET request.

Resolves IR-4951

## References
closes #_insert number here_


HexaField (commented):
my only concern here are cases in which we do in fact have non-normative services that accept an empty string or null as a valid value. if we don't have any, then i see no issues with this.

@hanzlamateen @MoizAdnan @hurairahmateen do we have any services like this?

HexaField (approved):
Actually, a quick search shows none of the sort. This change is fine.

PR #384, opened 2024-10-08
add optional redirectUrl param to loginUserByOAuth function

dtlehrer:
with MT PR: https://github.com/theinfinitereality/irpro-multitenancy/pull/533

to support redirecting to concierge (ecomm) dashboard after onboarding flow

### DIFF
--- a/packages/client-core/src/user/services/AuthService.ts
+++ b/packages/client-core/src/user/services/AuthService.ts
@@ -483,10 +483,10 @@ export const AuthService = {
   /**
    * Logs in the current user based on an OAuth response.
    */
-  async loginUserByOAuth(service: string, location: any) {
+  async loginUserByOAuth(service: string, location: any, redirectUrl?: string) {
     getMutableState(AuthState).merge({ isProcessing: true, error: '' })
     const token = getState(AuthState).authUser.accessToken
-    const path = new URLSearchParams(location.search).get('redirectUrl') || location.pathname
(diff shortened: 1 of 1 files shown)


packages/client-core/src/user/services/AuthService.ts
```diff
@@ -483,10 +483,10 @@ export const AuthService = {
   /**
    * Logs in the current user based on an OAuth response.
    */
-  async loginUserByOAuth(service: string, location: any) {
+  async loginUserByOAuth(service: string, location: any, redirectUrl?: string) {
     getMutableState(AuthState).merge({ isProcessing: true, error: '' })
     const token = getState(AuthState).authUser.accessToken
-    const path = new URLSearchParams(location.search).get('redirectUrl') || location.pathname
+    const path = redirectUrl || new URLSearchParams(location.search).get('redirectUrl') || location.pathname
```
aditya-mitra, line 489: did we initially try adding search params to the url like `url.com?redirectUrl=myurl` ?

aditya-mitra (dismissed):

achen5671 (approved):

packages/client-core/src/user/services/AuthService.ts
dtlehrer: yes, it was redirecting prematurely.. in this case we need to be able to redirect with params other than redirectUrl

PR #394, opened 2024-10-09
refresh directory on file browser actions

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-4499
## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/panels/files/toolbar.tsx
+++ b/packages/editor/src/panels/files/toolbar.tsx
@@ -34,7 +34,7 @@ import Input from '@ir-engine/ui/src/primitives/tailwind/Input'
 import Modal from '@ir-engine/ui/src/primitives/tailwind/Modal'
 import Slider from '@ir-engine/ui/src/primitives/tailwind/Slider'
 import Tooltip from '@ir-engine/ui/src/primitives/tailwind/Tooltip'
-import React, { Fragment } from 'react'
+import React, { Fragment, useEffect } from 'react'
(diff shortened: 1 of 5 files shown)


packages/editor/src/panels/files/toolbar.tsx
```diff
@@ -198,6 +198,10 @@ export default function FilesToolbar() {
 
   const { backDirectory, refreshDirectory, createNewFolder } = useCurrentFiles()
 
+  useEffect(() => {
+    refreshDirectory()
+  }, [filesState.refresh.value])
+
```
aditya-mitra, line 204: how about we directly call refreshDirectory?
(this way we can only do `when update needed -> refreshDirectory()`
with the state method, we are doing `when update needed -> set refresh value -> listen to refresh value -> refreshDirectory()`)

aditya-mitra (changes requested):
we can do this directly instead of using states

(if some more modifications are required to resolve the change request, you can feel free to assign to me)

packages/editor/src/panels/files/toolbar.tsx
MbfloydIR: @aditya-mitra 
The places where the refresh is needed are not in reactive contexts.
But the refreshDirectory method requires a reactive context to be called since it involves hooks.

CITIZENDOT:
I think this PR can be closed? [Bug: IR-4499](https://tsu.atlassian.net/browse/IR-4499) is not reproducible on dev anymore.

PR #403, opened 2024-10-10
IR-4861 avatar nameplates

SamMazerIR:
## Summary
bringing back avatar nameplates, distance based activation, only shows on other player avatars (not your own), displays username, transitions in and out with scale, height and opacity. Camera frustum culling also being used

## References
ticket:
[https://tsu.atlassian.net/browse/IR-4861](https://tsu.atlassian.net/browse/IR-4861)

design:
[https://www.figma.com/design/xG10l6SZBn92pFmPVMWfBG/Viewer-%26-XRUI?node-id=30-6852&t=jm0FDfyqsahXkwjp-1](https://www.figma.com/design/xG10l6SZBn92pFmPVMWfBG/Viewer-%26-XRUI?node-id=30-6852&t=jm0FDfyqsahXkwjp-1)

## QA Steps
join a location with two players (or one normal, one incognito tab) and view avatar nameplate above other player. 
you should see:
- distance threshold is roughly 8 meters
- scaling in/out when within distance threshold
- nameplate rises/lowers in/out
- nameplate fades in/out
- transition time of 0.25 seconds
- basic camera frustum culling (if the center of the nameplate is outside the camera view it will close


packages/client-core/src/systems/AvatarUISystem.tsx
```diff
@@ -94,7 +94,7 @@ export const renderAvatarContextMenu = (userId: UserID, contextMenuEntity: Entit
   contextMenuXRUI.quaternion.copy(cameraTransform.rotation)
 }
 
-const userQuery = defineQuery([AvatarComponent, TransformComponent, NetworkObjectComponent, Not(NetworkObjectOwnedTag)])
+const userQuery = defineQuery([AvatarComponent, TransformComponent, NetworkObjectComponent]) //, Not(NetworkObjectOwnedTag)])
```
HexaField, line 98: the `Not(NetworkObjectOwnedTag)` is to exclude your own avatar from having a nameplate, is this handled somewhere else now?

HexaField: ah I see, in the reactor. the query would make more sense.

packages/client-core/src/systems/AvatarUISystem.tsx
```diff
@@ -296,6 +296,24 @@ const reactor = () => {
       getMutableState(PopupMenuState).menus[AvatarMenus.AvatarContext].set(none)
     }
   }, [])
+
+  return (
+    <>
+      <QueryReactor
+        Components={[AvatarComponent, TransformComponent, NetworkObjectComponent]}
```
HexaField: should add `Not(NetworkOwnerTagComponent)` to the list rather than `if (userId !== Engine.instance.store.userID)`

packages/client-core/src/systems/AvatarUISystem.tsx
```diff
@@ -296,6 +296,24 @@ const reactor = () => {
...
+    </>
+  )
+}
+
+const AvatarInstanceReactor = () => {
+  const avatarEntity = useEntityContext()
+
+  useEffect(() => {
+    const userId = getComponent(avatarEntity, NetworkObjectComponent).ownerId
+    if (userId !== Engine.instance.store.userID) setComponent(avatarEntity, XruiNameplateComponent)
```
HexaField: we need a `return () => removeComponent` here to ensure we clean up

packages/engine/src/camera/functions/cameraFunctions.ts
```diff
@@ -0,0 +1,48 @@
...
+import { TransformComponent } from '@ir-engine/spatial'
+import { EngineState } from '@ir-engine/spatial/src/EngineState'
+import { CameraComponent } from '@ir-engine/spatial/src/camera/components/CameraComponent'
+import { Frustum, Matrix4, Vector3 } from 'three'
+
+const mat4 = new Matrix4()
+const frustum = new Frustum()
+const worldPosVec3 = new Vector3()
+
+export const inFrustum = (entity: Entity): boolean => {
```
HexaField: to make this reusable for other cameras, we should make `viewerEntity` an argument passed in and have `getState(EngineState).viewerEntity` be the default value

HexaField (dismissed):

packages/client-core/src/systems/AvatarUISystem.tsx
SamMazerIR: oh nice I wasn't aware of that one. making the change

MbfloydIR (approved):

PR #404, opened 2024-10-10
Changing node helpers to be visible by default.

Rezmason:
## References
Partially addresses but does not close [IR-4530](https://tsu.atlassian.net/browse/IR-4530).

[IR-4530]: https://tsu.atlassian.net/browse/IR-4530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DanielBelmes (approved):
great work

DanielBelmes (changes requested):
I realized right now that this would enable nodeHelperVisibility always. We're going to want to trigger this state to true on editor load. Maybe the same way we do it for grid state.

HexaField:
We spoke at standup today and have a new approach

SYBIOTE:
i am going to remove the trigger condition for the helpers from the button and tie it to selection instead

HexaField:
> i am going to remove the trigger condition for the helpers from the button and tie it to selection instead

We dont need to revive this branch.

SYBIOTE:
Working on a new fresh branch

PR #407, opened 2024-10-11
Ir 5000 fix particle assets

dinomut1:
Closes https://tsu.atlassian.net/browse/IR-5000

Fixes issue where reactor was crashing when no gltf component was found

Also fixes issue where particle assets not at the origin would render with double offset

### DIFF
--- a/packages/engine/src/scene/components/ParticleSystemComponent.ts
+++ b/packages/engine/src/scene/components/ParticleSystemComponent.ts
@@ -55,6 +55,7 @@ import { Entity, UUIDComponent } from '@ir-engine/ecs'
 import {
   defineComponent,
   getComponent,
+  hasComponent,
   setComponent,
   useComponent,
   useOptionalComponent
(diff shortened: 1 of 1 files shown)


packages/engine/src/scene/components/ParticleSystemComponent.ts
```diff
@@ -886,7 +889,7 @@ export const ParticleSystemComponent = defineComponent({
     const metadata = useHookstate({ textures: {}, geometries: {}, materials: {} } as ParticleSystemMetadata)
     const sceneID = useOptionalComponent(entity, SourceComponent)?.value
     const rootEntity = useHookstate(getMutableState(GLTFSourceState))[sceneID ?? ''].value
-    const sceneLoaded = GLTFComponent.useSceneLoaded(rootEntity)
+    const sceneLoaded = hasComponent(rootEntity, GLTFComponent) ? GLTFComponent.useSceneLoaded(rootEntity) : true
```
HexaField: hooks cannot be conditional. These logic needs to be handled with either a subreactor or something other way

HexaField (dismissed):

MichaelEstes (approved):

PR #409, opened 2024-10-11
IR-4927 Add a hook in the identity provider at the find endpoint and update the server to validate the user’s email before login.

mshafiqmk:
## Summary

JIRA Ticket [Sign-in page is allowing for new membership signups](https://tsu.atlassian.net/browse/IR-4927)
## References
closes #_insert number here_


packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
```diff
@@ -294,6 +295,31 @@ async function createAccessToken(context: HookContext<IdentityProviderService>)
   }
 }
 
+async function checkUserPermissions(context: HookContext): Promise<void> {
+  // Fetch user scopes
+  const userId = context.params.user.id
+  const scopes = (await API.instance.service(scopePath).find({
```
hanzlamateen: Use `context.app.service`. Make use of correct typings

packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
```diff
@@ -294,6 +295,31 @@ async function createAccessToken(context: HookContext<IdentityProviderService>)
...
+    query: {
+      $select: ['type'],
+      userId: userId,
+      paginate: false
+    }
+  })) as any as { type: string }[]
+
+  // Check for editor:write scope and any admin scope
+  const hasEditorWriteScope = scopes.some((scope) => scope.type === 'editor:write')
+  const hasAdminScope = scopes.some((scope) => scope.type.startsWith('admin:'))
```
hanzlamateen: Use common hooks like verifyScope, checkScope, etc

hanzlamateen (approved):

hanzlamateen (changes requested):
Mistakenly approved last time. Need to address the comments that I left.

packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
```diff
@@ -307,7 +309,23 @@ export default {
...
+      // Combined check for external provider and $like usage with scope verification
+      iff(
+        isProvider('external') &&
+          ((context) =>
+            context.params.query?.accountIdentifier?.$like !== undefined ||
+            context.params.query?.email?.$like !== undefined), // Use iffElse to check for admin action
+        iffElse(
+          isAction('admin'),
+          // If the action is admin, check for admin or super scope
+          verifyScope('admin', 'super'),
```
hanzlamateen:
```suggestion
          verifyScope('admin', 'admin'),
```

hanzlamateen (changes requested):

packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
```diff
@@ -311,20 +333,7 @@ export default {
...
-            context.params.query?.email?.$like !== undefined), // Use iffElse to check for admin action
-        iffElse(
-          isAction('admin'),
-          // If the action is admin, check for admin or super scope
-          verifyScope('admin', 'super'),
-          // If not admin action, check for editor:write scope
-          verifyScope('editor', 'write')
-        )
-      )
+      iff(isProvider('external') && validateQueryParameters, verifyScope('admin', 'admin'))
```
hanzlamateen: I dont think we can use `&& validateQueryParameters` like this. I think it need a hook context 

https://github.com/ir-engine/ir-engine/blob/cd401f049ea35a2ee7477db8f17d6ef4c65a331e/packages/server-core/src/networking/instance/instance.hooks.ts#L157

hanzlamateen (changes requested):

packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
mshafiqmk: @hanzlamateen updated.

hanzlamateen: Please also write unit test to verify this hook.

I doubt if it works because if you see my referenced `packages/server-core/src/hooks/is-action.ts` hook it returns a delegate.

Tests will help verify.

hanzlamateen (changes requested):

packages/server-core/src/user/identity-provider/identity-provider.test.ts
```diff
@@ -317,4 +317,100 @@ describe('identity-provider.test', () => {
       }
     )
   })
+
+  it('should not allow search without the required scope - multiple queries', async () => {
```
hanzlamateen: Please add a test case when the user is admin then then he should be able to make the queries.

hanzlamateen (changes requested):

packages/client-core/src/user/services/AuthService.ts
```diff
@@ -679,6 +679,21 @@ export const AuthService = {
     }
   },
 
+  async validateUser(email: string): Promise<boolean> {
+    try {
+      const identityProviders = await API.instance.service(identityProviderPath).find({
+        query: {
+          email: email.toLowerCase()
+        },
+        paginate: false
```
hanzlamateen: On client side paginate: false doesnt work in params.

Why is it required all together?

packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
```diff
@@ -269,6 +269,16 @@ async function createAccessToken(context: HookContext<IdentityProviderService>)
   }
 }
 
+const isSearchQuery = (context: HookContext) => {
+  const { query } = context.params
+  const queryLength = Object.keys(query).length
+  // we only need to allow search based on exact email in the query
+  if (queryLength === 2 && query.email && !query.email.$like && !query.email.$notlike) {
```
hanzlamateen, line 276: Why is query length 2 here? What else is passed beside email?

hanzlamateen (changes requested):

packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
mshafiqmk: 'select' with no values

mshafiqmk: like this 
``` [ 'email', '$select' ] ```

hanzlamateen (approved):
Looks good. @barankyle can you give a 2nd look.

barankyle (approved):

PR #410, opened 2024-10-11
add github action to deploy storybook

CITIZENDOT:
## Summary
Whenever there's a change in `ui` package, this new github action gets triggered and pushes the static storybook files to [storybook-docs](https://github.com/ir-engine/storybook-docs) repo. Then it gets published through github pages.

## References
closes #_insert number here_


.github/workflows/update-storybook.yml
```diff
@@ -0,0 +1,42 @@
...
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+
+          git clone https://x-access-token:${{ secrets.STORYBOOK_DOCS_PAT }}@github.com/your-username/storybook-docs.git
+          cp -R packages/ui/storybook-static/* storybook-docs/
+          
+          cd storybook-docs
+          touch .nojekyll
+          git add .
+          git commit -m "Update Storybook documentation"
```
aditya-mitra, line 44: `commit --amend`

git history will not be necessary for generated files

aditya-mitra (changes requested):

.github/workflows/update-storybook.yml
CITIZENDOT: updated

aditya-mitra (approved):
Awesome 🚀

packages/ui/postcss.config.js
```diff
@@ -25,7 +25,7 @@ Infinite Reality Engine. All Rights Reserved.
 */
 
 
-module.exports = {
+export default {
```
HexaField: need to revert this as we reverted using ESM

HexaField (changes requested):
Can this action be in the storybook repo instead? It can listen for changes to this repository with event webhooks. What are the tradeoffs of having it in this repo vs the other repo?

packages/ui/postcss.config.js
CITIZENDOT: reverted

CITIZENDOT:
> Can this action be in the storybook repo instead?

Yes, that is possible. Instead of creating/using secret token, webhook needs to be created.

HexaField:
Let's do that then.

PR #419, opened 2024-10-14
Add option for font

SYBIOTE:
This pull request adds a new option for font selection. It includes a patch that modifies the `FontMaterial` array to include a new font option called "Standard".


aditya-mitra (approved):
LGTM

PR #421, opened 2024-10-14
Force display precision on UI

SYBIOTE:
This pull request includes a commit that forces the display precision on the UI. The commit modifies the `NumericInput` component to ensure that the displayed value has the specified precision. It introduces a new function `toPrecisionString` that handles the precision formatting. The commit also updates the `onChange` and `handleFocus` functions to use the `toPrecisionString` function for setting the temporary value and focused state respectively.


aditya-mitra (approved):
LGTM

PR #422, opened 2024-10-15
feat(dashboard): hide the iframe to avoid space in small devices [IR-4930]

mikeplascdev:
## Summary

- Changed to display none because on small devices the iframe takes up space on the right
- Add the unsupported device and browser i18n to the common file
- use the new hook in the editor page for the studio

## References
closes #_insert number here_

## Evidence
Screenshot 2024-10-16 at 2 10 07 p m
Screenshot 2024-10-16 at 2 06 59 p m

## Jira
[IR-4930](https://tsu.atlassian.net/browse/IR-4930)

[IR-4930]: https://tsu.atlassian.net/browse/IR-4930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hanzlamateen:
@barankyle hiding iframe will be okay?

barankyle:
I don't think this is viable long-term. It's currently not an issue since everything in MT is on the same root domain, so we don't need to display and use the iframe to capture user input to authorize cross-origin cookie sharing. But when we do support customers bringing their own domains, this iframe will need to be displayed.

Is the video showing what it looks like after it's been set to display: none? Because I don't see the iframe anywhere. A video of the problem this is addressing would be useful.

packages/editor/src/pages/EditorPage.tsx
```diff
@@ -23,26 +23,18 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
 */
 
-import { PopoverState } from '@ir-engine/client-core/src/common/services/PopoverState'
 import '@ir-engine/engine/src/EngineModule'
 import { getMutableState, useHookstate, useImmediateEffect } from '@ir-engine/hyperflux'
 import { loadEngineInjection } from '@ir-engine/projects/loadEngineInjection'
 import { EngineState } from '@ir-engine/spatial/src/EngineState'
-import Button from '@ir-engine/ui/src/primitives/tailwind/Button'
-import Modal from '@ir-engine/ui/src/primitives/tailwind/Modal'
+import { useUnsupported } from '@theinfinitereality/irpro-multitenancy/components/hooks/useUnsupported'
```
jose-galvan: You cannot import from a project into the core. You need to move the implementation to the core

jose-galvan (changes requested):

packages/editor/src/pages/EditorPage.tsx
MoizAdnan: ^

MoizAdnan (dismissed):

packages/editor/src/pages/EditorPage.tsx
mikeplascdev: done 163a8936f67b5aa7e48e947962d4b56d9ea3f56d

packages/client-core/src/components/modals/UnsupportedBrowser.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { getMutableState, useHookstate } from '@ir-engine/hyperflux'
+import Button from '@ir-engine/ui/src/primitives/tailwind/Button'
+import Modal from '@ir-engine/ui/src/primitives/tailwind/Modal'
+import Text from '@ir-engine/ui/src/primitives/tailwind/Text'
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import { HiOutlineExclamationCircle } from 'react-icons/hi2'
+
+const downloadGoogleLink =
+  'https://www.google.com/chrome/dr/download/?brand=CBFU&ds_kid=43700079286123654&gad_source=1&gclid=CjwKCAjwooq3BhB3EiwAYqYoEkgLBNGFDuKclZQTGAA8Lzq66cvirjjOm7ur0ayMgKvn9y3Fd1spThoCXu0QAvD_BwE&gclsrc=aw.ds'
```
hanzlamateen:
```suggestion
  'https://www.google.com/chrome/dr/download/'
```

Should above be enough? Can you check from original author why query params were introduced?

hanzlamateen (changes requested):

packages/client-core/src/components/modals/UnsupportedBrowser.tsx
hanzlamateen: @mikeplascdev

hanzlamateen (changes requested):
I am using Chrome, but still get unsupported browser dialog.

https://github.com/user-attachments/assets/ee36ea37-e6f2-4ccb-a530-8db8db166ba7

hanzlamateen:
> I don't think this is viable long-term. It's currently not an issue since everything in MT is on the same root domain, so we don't need to display and use the iframe to capture user input to authorize cross-origin cookie sharing. But when we do support customers bringing their own domains, this iframe will need to be displayed.
> 
> Is the video showing what it looks like after it's been set to display: none? Because I don't see the iframe anywhere. A video of the problem this is addressing would be useful.

@mikeplascdev is this addressed in recent changes?

packages/client-core/src/components/modals/UnsupportedBrowser.tsx
mikeplascdev: done 838167d638db08dd85b5ec120f799cffbf5057af

mikeplascdev:
@hanzlamateen 
https://github.com/ir-engine/ir-engine/pull/422#pullrequestreview-2397767377
done 5fc0b9c85b57a8eb03571905adb2306f034d09a4

mikeplascdev:
@hanzlamateen 
cc: @barankyle 
for this comment https://github.com/ir-engine/ir-engine/pull/422#issuecomment-2440365895
this is the video of the evidence

https://github.com/user-attachments/assets/2eea2699-965b-449f-b4d5-6e5d91010507

barankyle:
Got a chance to test the iframe behavior and it was indeed not being rendered when it should have, since it had been written to be visible based on the `visibility` property. I pushed a couple of lines to set the iframe's `display` property to set it to `display: block` when it's needed, and otherwise set/leave it as `display: none`. Let me know if that's causing any issues.

hanzlamateen (approved):

kattsushi (approved):
LGTM

jose-galvan (approved):

barankyle (approved):

PR #427, opened 2024-10-15
Prefab objects receive the correct name

MbfloydIR:
updated the add media code to pass the requested name into the the editor control function to create the the scene object, so it can be used instead of "new object"

## Summary
https://tsu.atlassian.net/browse/IR-5033
## References
closes #_insert number here_


packages/editor/src/functions/addMediaNode.ts
```diff
@@ -69,6 +68,12 @@ export async function addMediaNode(
   const contentType = (await getContentType(url)) || ''
   const { hostname } = new URL(url)
 
+  const pathArray = url.split('/')
+  const lastIndex = pathArray.length - 1
+  const fileNameWithExt = pathArray[lastIndex]
+  const fileNameArray = fileNameWithExt.split('.')
+  const name = decodeURI(fileNameArray[0])
```
DanielBelmes: We should wrap the logic here in a try catch as there's a chance the url comes in with invalid characters for decodeURI or somehow the

DanielBelmes (approved):

PR #428, opened 2024-10-15
Ir 5066 fix avatar id existence

DanielBelmes:
## Summary
There is still a chance that a user has no avatar defined which leads to undefined behavior

## References
closes #_insert number here_


barankyle (approved):

HexaField (approved):

packages/server-core/src/user/user-avatar/user-avatar.hooks.ts
```diff
@@ -42,6 +45,20 @@ import {
...
+  for (const item of data) {
+    const entry = await context.service.find({
+      userId: context.params.query.userId
+    })
+    if (entry.total === 0) {
+      await context.service.create({ userId: context.params.query.userId, avatarId: item.avatarId })
+    }
+  }
+  return
+}
```
hanzlamateen, line 60: @DanielBelmes just a suggestion, this can be done in a single find and single create query instead of doing in a for loop. I think following block will always return same result:

```
const entry = await context.service.find({
      userId: context.params.query.userId
    })
```

And create call can access an array I guess.

DanielBelmes: Gotcha. I thought so and was discussing with Josh and we just decided to make it look like the other hooks for consistency. But yea my original version was just the inside of the for loop and checking for data. The for loop iteration does act as a check for data which is neat.

PR #431, opened 2024-10-16
Speed up UI updates upon sending chat message

HexaField:
## Summary

- removes messages unnecessary resolver on channel service
- adds the message to the rendered list as soon as the create call finishes

IR-4941

## References
closes #_insert number here_


aditya-mitra (approved):
Looks good!

PR #432, opened 2024-10-16
fix useEngineCanvas

HexaField:
## Summary

Cherry picked from gltf branch

## References
closes #_insert number here_


packages/client-core/src/hooks/useEngineCanvas.ts
```diff
@@ -59,12 +55,56 @@ export const useEngineCanvas = (ref: React.RefObject<HTMLElement>) => {
...
+    rendererComponent.scenes.merge([originEntity])
+
+    return () => {
+      if (!hasComponent(viewerEntity, RendererComponent)) return
+      const index = rendererComponent.scenes.value.indexOf(originEntity)
+      rendererComponent.scenes[index].set(none)
+    }
+  }, [viewerEntity, originEntity])
+
+  useEffect(() => {
```
MichaelEstes, line 94: I think this useEffect got duplicated somehow

MichaelEstes (approved):

packages/client-core/src/hooks/useEngineCanvas.ts
HexaField: nope, one is local floor one is origin :)

PR #435, opened 2024-10-16
turn off newScene modal

JT00y:
## References
closes #_insert number here_


dinomut1 (commented):
we don't want to remove this functionality, we just want to remove the addon from the world injection

PR #437, opened 2024-10-16
studio: add skybox to default scene

achen5671:
## References
closes #_insert number here_


dinomut1 (approved):

packages/projects/default-project/resources.json
```diff
@@ -66,7 +68,7 @@
     "thumbnailMode": "automatic"
   },
   "assets/avatars/female_01.vrm": {
-    "type": "avatar",
+    "type": "asset",
```
HexaField, line 71: Please do not revert the avatars to assets, this will cause problems.

It's a little broken right now, but will still work OK if we don't commit this to the seed project.

HexaField (changes requested):

PR #438, opened 2024-10-16
turn off gizmo for wizard

JT00y:
## References
closes #_insert number here_


packages/editor/src/systems/EditorControlSystem.ts
```diff
@@ -362,8 +362,14 @@ const execute = () => {
...
     }
   }
+
   if (buttons.PrimaryClick?.up && !buttons.PrimaryClick?.dragging) {
-    if (hasComponent(clickStartEntity, SourceComponent) && !getState(ClickPlacementState).placementEntity) {
+    const urlParams = new URLSearchParams(window.location.search)
+    if (
+      hasComponent(clickStartEntity, SourceComponent) &&
+      !getState(ClickPlacementState).placementEntity &&
+      urlParams.get('wizardEnabled') === 'false'
```
aditya-mitra: `wizardEnabled` comes from the irpro-wizard repo
is there a way we can dynamically do this?
(what if we try to enable it in the irpro-wizard itself by copying some required parts of the EditorControlSystems)

HexaField (changes requested):
we should not rely on url parameters for this, and we should also not be including any reference to the wizard in the core engine

something that would make more sense is a "gizmoEnabled" boolean in a state that can be turned off from the wizard

JT00y:
> we should not rely on url parameters for this, and we should also not be including any reference to the wizard in the core engine
> 
> something that would make more sense is a "gizmoEnabled" boolean in a state that can be turned off from the wizard

is it mean I can add a new state for gizmoEnbaled in core engine? where I can put this state?

HexaField:
> is it mean I can add a new state for gizmoEnbaled in core engine? where I can put this state?

Yes.`EditorHelperState` would make sense

HexaField (approved):

PR #440, opened 2024-10-17
Refactor User Resolvers

HexaField:
## Breaking Changes

- can no longer create new users from the admin panel. this feature never made sense anyway, as the user would not have an identity provider and a user would never bea ble to log in

## References
closes #_insert number here_

## QA Steps

- test user login
- test admin & studio scopes are correct
- test adding & removing scopes on users
- test changing theme mode still works and is persisted to user settings


barankyle:
When I try to go to the admin pages, the useFind to get the user's scopes is resolving as an empty array, so it's always kicking me back to the main page. I think we need to wait until the useFind is resolved? When I print it out, it's only getting as far as the pending state before redirecting.

barankyle:
Login and invites seem to work, though, that's good, and the locations appear to be working.

HexaField:
> When I try to go to the admin pages, the useFind to get the user's scopes is resolving as an empty array, so it's always kicking me back to the main page. I think we need to wait until the useFind is resolved? When I print it out, it's only getting as far as the pending state before redirecting.

oh yeah, the status needs to be considered

barankyle:
Running into an issue with /admin/users after I fixed a bug with addLastLogin. When the page goes to render the results, it crahes, and this error is shown:

```
Warning: React has detected a change in the order of Hooks called by UserTable. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://reactjs.org/link/rules-of-hooks

   Previous render            Next render
   ------------------------------------------------------
1. useContext                 useContext
2. useState                   useState
3. useRef                     useRef
4. useEffect                  useEffect
5. useRef                     useRef
6. useEffect                  useEffect
7. useState                   useState
8. useLayoutEffect            useLayoutEffect
9. useState                   useState
10. useLayoutEffect           useLayoutEffect
11. useState                  useState
12. useLayoutEffect           useLayoutEffect
13. useState                  useState
14. useLayoutEffect           useLayoutEffect
15. useLayoutEffect           useLayoutEffect
16. useMemo                   useMemo
17. useState                  useState
18. useLayoutEffect           useLayoutEffect
19. useState                  useState
20. useLayoutEffect           useLayoutEffect
21. useState                  useState
22. useLayoutEffect           useLayoutEffect
23. useLayoutEffect           useLayoutEffect
24. useMemo                   useMemo
25. useEffect                 useEffect
26. useState                  useState
27. useLayoutEffect           useLayoutEffect
28. useCallback               useCallback
29. useCallback               useCallback
30. useCallback               useCallback
31. useCallback               useCallback
32. useMemo                   useMemo
33. useState                  useState
34. useLayoutEffect           useLayoutEffect
35. undefined                 useState
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    at UserTable (https://localhost:3000/@fs/home/kyle/Development/Forks/ir-engine/packages/client-core/src/admin/components/user/UserTable.tsx:39:3)
    at Users (https://localhost:3000/@fs/home/kyle/Development/Forks/ir-engine/packages/client-core/src/admin/components/user/index.tsx:14:17)
    at RenderedRoute (https://localhost:3000/node_modules/.vite/deps/react-router-dom.js?v=137b8fa4:3098:5)
```

HexaField:
> Running into an issue with /admin/users after I fixed a bug with addLastLogin. When the page goes to render the results, it crahes, and this error is shown:

Fixed

packages/server-core/src/user/user-avatar/user-avatar.resolvers.ts
```diff
@@ -36,7 +37,16 @@ export const userAvatarResolver = resolve<UserAvatarType, HookContext>({
...
 })
 
-export const userAvatarExternalResolver = resolve<UserAvatarType, HookContext>({})
+export const userAvatarExternalResolver = resolve<UserAvatarType, HookContext>({
+  avatar: virtual(async (userAvatar, context) => {
+    if (context.event !== 'removed' && userAvatar.avatarId)
+      try {
+        return await context.app.service(avatarPath).get(userAvatar.avatarId, { query: { skipUser: true } })
+      } catch (err) {
+        return {}
```
hanzlamateen:
```suggestion
        return undefined
```

packages/common/src/schemas/user/user-avatar.schema.ts
```diff
@@ -46,6 +46,7 @@ export const userAvatarSchema = Type.Object(
     avatarId: TypedString<AvatarID>({
       format: 'uuid'
     }),
+    avatar: Type.Ref(avatarDataSchema),
```
hanzlamateen, line 49:
```suggestion
    avatar: Type.Optional(Type.Ref(avatarDataSchema)),
```

packages/server-core/src/user/user/user.hooks.ts
```diff
@@ -314,6 +285,26 @@ const handleUserSearch = async (context: HookContext<UserService>) => {
...
+    const lastLogin = await context.app.service(userLoginPath).find({
+      query: {
+        userId: user.id,
+        $sort: { createdAt: -1 },
+        $limit: 1
+      }
+    })
+    user.lastLogin = lastLogin.data[0]
+  }
+}
```
hanzlamateen: Why is this removed from resolver?

hanzlamateen (changes requested):

packages/server-core/src/user/user/user.hooks.ts
HexaField: Because it should only be resolved for external requests with admin scope

hanzlamateen: This can go in external resolver and there we can put a check for scopes.

In current implementation `iff(isProvider('external'), verifyScope('admin', 'admin'), addLastLogin)` will cause all external find calls to have `admin:admin` scope. Is it really intended? Because `verifyScope` hook will throw exception if that particular scope is missing. For checking, checkScope hook is preferred as it will return true, false. i.e. https://github.com/ir-engine/ir-engine/blob/9895e970b442607a5d77499831cb28ccad293215/packages/server-core/src/setting/project-setting/project-setting.hooks.ts#L85

HexaField: Hmm @barankyle ?

barankyle: I agree with Hanzla, doing this verifyScope on admin:admin would throw errors as-constructed if a non-admin with user:read ran user.FIND. I'm ambivalent on doing checkScope in the external resolver or as part of an iffElse in the hooks.

hanzlamateen (approved):

PR #442, opened 2024-10-17
studio: throw warning for unsupport file

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-4972
https://tsu.atlassian.net/browse/IR-5014

## References
closes #_insert number here_


packages/editor/src/functions/assetFunctions.ts
```diff
@@ -72,18 +73,37 @@ function findMimeType(file): FileType {
...
+    isValid: false,
+    errorMessage: unsupportedFileMessage[mimeType]
+  }
+}
+
+function sanitizeFiles(files) {
+  const newFiles: File[] = []
+  for (const file of files) {
+    const newFile = cleanFileNameFile(file)
+    const { isValid, errorMessage } = isValidFileType(newFile)
```
aditya-mitra, line 99: is possible, we should check if the file is valid before cleaning the file

aditya-mitra (changes requested):
just 1 change

packages/editor/src/functions/assetFunctions.ts
achen5671: yea we can do that but I dont think that's necessary here becaues we're not throwing an error and failing the upload. We are just throwing a alert to the user telling them it's not supported.

achen5671:
Screenshot 2024-10-21 at 11 47 32 AM

DanielBelmes (approved):

DanielBelmes:
Looks good to me @aditya-mitra please review the response. I do agree if we were blocking the upload that we should check is valid first since its less expensive than cleaning. But it looks like the file is still getting added if it's not supported. Though we should ask product if that really is what we want.

aditya-mitra (approved):
looks good

aditya-mitra:
> Looks good to me @aditya-mitra please review the response. I do agree if we were blocking the upload that we should check is valid first since its less expensive than cleaning. But it looks like the file is still getting added if it's not supported. Though we should ask product if that really is what we want.

right Daniel. will ask Ross about this

PR #445, opened 2024-10-17
Fix file browser bugs and mat saving

DanielBelmes:
## Summary

- Fixes projectName being passed down to gltfWriter
- Fixes possible undefined passed to FilesQueryContext
- Fixes possible undefined race condition when deleting files in file view

## References
closes #_insert number here_


packages/editor/src/panels/files/fileitem.tsx
```diff
@@ -114,7 +114,7 @@ function TableView({ file, onClick, onDoubleClick, isSelected, drag, drop, isOve
     staticResourceModifiedDates.set(modifiedDates)
   }, [staticResourceData.status])
 
-  const thumbnailURL = file.thumbnailURL
+  const thumbnailURL = file?.thumbnailURL
```
aditya-mitra, line 117: Hey Daniel, can you explain the case when `file` is undefined?

aditya-mitra (dismissed):

packages/editor/src/panels/files/fileitem.tsx
aditya-mitra: got it. I think some files are undefined because the file browser is not refreshed after a file is deleted.

DanielBelmes: Sorry I totally missed this last week. Do you think the core issue is fixable. It seemed like a race condition most of the time. I couldn't actually reproduce the bug from the ticket. But I know this will fix it.

packages/editor/src/panels/materials/helpers.tsx
```diff
@@ -91,7 +91,7 @@ export async function saveMaterial(sourcePath: string) {
   const gltf = (await exportMaterialsGLTF([UUIDComponent.getEntityByUUID(materialUUID)], {
     binary: false,
     relativePath,
-    projectName: projectName
```
DanielBelmes, line 94: This seemed to have been ported out of this PR since this never got merged. But I kept my change anyway since it's a little cleaner

dinomut1 (approved):

PR #450, opened 2024-10-17
expanded if condition for scene check in thumbnail gen

DanielBelmes:
## References
closes #_insert number here_


aditya-mitra (approved):
LGTM ✅

PR #452, opened 2024-10-17
Allow projects to change prefab category icon

DanielBelmes:
## References
closes #_insert number here_


packages/editor/src/components/prefabs/PrefabEditors.tsx
```diff
@@ -55,6 +56,9 @@ export const PrefabIcon = ({ categoryTitle, isSelected }: { categoryTitle: strin
     Default: <IoCubeOutline className={`h-5 w-5 text-[${color}]`} />
   }
 
+  const IconElement = getState(PrefabShelfState).find((item) => item.category == categoryTitle)?.icon
```
HexaField, line 59: should this be reactive?

DanielBelmes: No it's compile time only because of how we use vite with svgs

HexaField (approved):

PR #456, opened 2024-10-18
speed up vitest on server-core

aditya-mitra:
## References
closes #_insert number here_


aditya-mitra:
automatic sharding is not working on the tests. will need to manual shard the tests.
will also find which file is taking too long

PR #458, opened 2024-10-18
IR-4403 fix for Rapier dynamic rigidbody type not updating

SamMazerIR:
## Summary
fixed issue surrounding Rapier not properly updating the body type without a disable, enable call on the rigidbody as a workaround to the issue

## References
[https://tsu.atlassian.net/browse/IR-4403](https://tsu.atlassian.net/browse/IR-4403)

## QA Steps
1. create an entity (preferably with geometry, either a model or primitive geo)
2. move its Y position up a bit so it isn't inside the platform or any other objects etc.
3. add a collider component
4. add a rigidbody component
5. change rigidbody component type to dynamic
6. watch it fall down instead of just staying locked in position as bugged before


MbfloydIR (approved):

packages/spatial/src/physics/classes/Physics.ts
```diff
@@ -293,6 +293,10 @@ const setRigidBodyType = (world: PhysicsWorld, entity: Entity, type: Body) => {
   }
 
   rigidbody.setBodyType(typeEnum, false)
```
HexaField, line 295: Have you tried waking the body up by changing this line's second parameter to true?

SamMazerIR: yes, didn't work

SamMazerIR: https://github.com/pmndrs/react-three-rapier/issues/314

HexaField (approved):

PR #460, opened 2024-10-18
add loading spinner for lookdev

JT00y:
## References
closes #_insert number here_


HexaField:
is this still necessary after this PR? https://github.com/ir-engine/ir-engine/pull/455

JT00y:
> is this still necessary after this PR? #455

yes, this is for the lookdev asset which not use model component, it just to overwrite the value of certian components, therefore I make reactor to listen snapshot.

HexaField:
> > is this still necessary after this PR? #455
> 
> yes, this is for the lookdev asset which not use model component, it just to overwrite the value of certian components, therefore I make reactor to listen snapshot.

Ok. After #455 is in, we should move this logic there, rather than adding even more complexity to the loading logic. This will help keep functionality together and avoid creating more and more nested reactors.

packages/editor/src/functions/addMediaNode.ts
```diff
@@ -169,10 +194,28 @@ export async function addMediaNode(
...
+          const fileNameWithExt = pathArray[lastIndex]
+          const fileNameArray = fileNameWithExt.split('.')
+          const name = decodeURI(fileNameArray[0])
+          setComponent(entity, NameComponent, name)
+
+          reactor.stop()
+        }, [modelComponent])
+
+        return null
+      })
```
dinomut1: let's remove this from this PR, not in scope

dinomut1 (changes requested):

dinomut1:
> > > is this still necessary after this PR? #455
> > 
> > 
> > yes, this is for the lookdev asset which not use model component, it just to overwrite the value of certian components, therefore I make reactor to listen snapshot.
> 
> Ok. After #455 is in, we should move this logic there, rather than adding even more complexity to the loading logic. This will help keep functionality together and avoid creating more and more nested reactors.

That won't work as the lookdev objects don't use the model component, or any other component that can be queried by a system. Let's leave this as is

packages/editor/src/functions/addMediaNode.ts
JT00y: okay i will do this, there is a fix for load lookdev asset with multiple lookdev components, i will leave tthis in this pr

packages/editor/src/functions/addMediaNode.ts
```diff
@@ -126,25 +121,54 @@ export async function addMediaNode(
...
+        useEffect(() => {
+          if (loadingEntity.value !== UndefinedEntity && snapshots.length !== length) {
+            removeEntityNodeRecursively(loadingEntity.value)
+            loadingEntity.set(UndefinedEntity)
+            if (reactor) reactor.stop()
+          }
+        }, [snapshots.length])
+        return null
+      })
+      return null
```
dinomut1: I don't think we need this reactor. We should just keep this simple, add a loading spinner before the gltfloader.load function, and then remove it in the onLoad callback, and add an onError callback that removes it as well.

dinomut1 (changes requested):

dinomut1 (approved):

PR #466, opened 2024-10-21
add rings to default project

dinomut1:
Adds rings glb to default project, which is being referenced by several starter prefabs


Rezmason (approved):
Straightforward

PR #470, opened 2024-10-21
Physics worlds wait for a scene to load before

DanielBelmes:
## Summary
This is a basic implementation of the concept of Scene active. Where a physics world won't be created if the scene is not fully loaded. We have to finalize this but this is functional and doesn't seem risky as far as I can tell. 

## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/physics/systems/PhysicsSystem.tsx
+++ b/packages/spatial/src/physics/systems/PhysicsSystem.tsx
@@ -112,7 +112,7 @@ const PhysicsSceneReactor = () => {
     return () => {
       Physics.destroyWorld(uuid)
     }
-  }, [uuid, scene.active])
+  }, [uuid, scene.active.value])
   return null
 }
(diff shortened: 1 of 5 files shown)


packages/spatial/src/physics/systems/PhysicsSystem.tsx
```diff
@@ -104,12 +104,15 @@ const execute = () => {
...
+  const scene = useComponent(entity, SceneComponent)
+
   useEffect(() => {
+    if (!scene.active.value) return
     Physics.createWorld(uuid)
     return () => {
       Physics.destroyWorld(uuid)
     }
-  }, [uuid])
+  }, [uuid, scene.active])
```
HexaField: definitely need .value

HexaField (changes requested):

HexaField (approved):

PR #476, opened 2024-10-22
remove close project confirm for wizard

JT00y:
## References
closes #_insert number here_


dinomut1 (changes requested):
we shouldn't be using the gizmo enabled option as a tell for whether the wizard is enabled or not. This requires a change to confirm scene save, which already needs work, as it doesn't let you close a scene without saving.
For now this problem is blocked

dinomut1:
This has already been implemented in the wizard

PR #486, opened 2024-10-23
IR-5157 bugfix - uniform scaling issue with Z resetting to zero

SamMazerIR:
## Summary
adding in Z value for uniform scaling that was defaulting to 0 without being specified

## References
[https://tsu.atlassian.net/browse/IR-5157](https://tsu.atlassian.net/browse/IR-5157)

## QA Steps
1. set a transform's scale uniform lock to locked.
2. drag any of the x/y/z values, all should scale properly (z should not immediately and independently be set to 0)


Rezmason (approved):
Yes.

PR #497, opened 2024-10-24
Add back blob override for node

HexaField:
## References
closes #_insert number here_


DanielBelmes (changes requested):
This is supposed to go into release0.9

DanielBelmes:
reopening based off 0.9

PR #499, opened 2024-10-24
Scene load stuck at 99% percent fix

MichaelEstes:
Update debug texture loading condition, remove resource pending component on unmount, wait for renderer to exist in media component

## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/scene/components/MediaComponent.ts
+++ b/packages/engine/src/scene/components/MediaComponent.ts
@@ -187,10 +187,12 @@ export function MediaReactor() {
   const mediaElement = useOptionalComponent(entity, MediaElementComponent)
   const audioContext = getState(AudioState).audioContext
   const gainNodeMixBuses = getState(AudioState).gainNodeMixBuses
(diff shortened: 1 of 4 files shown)


packages/engine/src/scene/components/MediaComponent.ts
```diff
@@ -187,10 +187,12 @@ export function MediaReactor() {
   const mediaElement = useOptionalComponent(entity, MediaElementComponent)
   const audioContext = getState(AudioState).audioContext
   const gainNodeMixBuses = getState(AudioState).gainNodeMixBuses
+  const rendererAvailable = RendererComponent.useRendererAvailable(Engine.instance.viewerEntity)
```
HexaField: this is unnecessary, in fact, the assumption in the useEffect that we are targeting `Engine.instance.viewerEntity` is part of the problem. we need to use `useRendererEntity` here instead and then pass that to a `useOptionalComponent(rendererEntity, RendererComponent)` and assert that the `rendererComponent.renderer.value` exists in the `useEffect`

HexaField (changes requested):

HexaField (approved):

PR #500, opened 2024-10-24
Define min height for asset panel scrollbar

SYBIOTE:
This pull request adds a minimum height for the asset panel scrollbar. The `asset-panel` class now includes a `min-height` property of 40px.


DanielBelmes (changes requested):
This was requested for 0.9 so if you can base it off and merge into release0.9 that would be perfect.

DanielBelmes:
reopening for new base

PR #503, opened 2024-10-24
various fixes for spectate mode

HexaField:
## References
closes #_insert number here_


DanielBelmes:
Just commenting things we did on discord.

- Spectate Button always on widget bar
- Spectate didn't seem to start and just gives user

HexaField:
> Just commenting things we did on discord.
> 
> * Spectate Button always on widget bar
> * Spectate didn't seem to start and just gives user

fixed. very temperamental reactor.

AidanCaruso (approved):
Definitely an improvement, however forcing scene spectate with and without an entity from scene settings still doesn't seem to work properly. Enabling scene spectate from settings without an optional target entity doesn't serialize anymore.

HexaField:
> Definitely an improvement, however forcing scene spectate with and without an entity from scene settings still doesn't seem to work properly. Enabling scene spectate from settings without an optional target entity doesn't serialize anymore.

We're going to be refactoring this property soon, so I'm not worried about it just yet.

PR #505, opened 2024-10-24
Fix loading spinner not going away when switching variants

dinomut1:
Fixes bug where variants would have an infinite loading spinner when switching
To test, add the variant trigger prefab from ir-starter-assets to a scene and test switching between the two variants on dev + this branch. Observe that the spinner never goes away on dev but does go away on this branch


dinomut1:
re-basing this to rc branch

PR #506, opened 2024-10-24
IR-5143 GroundPlane Component and ColorInput issues fix

SamMazerIR:
## Summary
extra padding in groundPlane node editor, removing a comment in image/SourceProperties, fixing hyperactive reactivity in color picker

## References
[https://tsu.atlassian.net/browse/IR-5143](https://tsu.atlassian.net/browse/IR-5143)

## QA Steps
open/close the color picker in groundPlane component, change colors, click away, should see expected behavior of a color picker

check/uncheck the visibility of the ground plane, should not only work, but show expected behavior

### DIFF
--- a/packages/ui/src/primitives/tailwind/Color/index.tsx
+++ b/packages/ui/src/primitives/tailwind/Color/index.tsx
@@ -24,7 +24,7 @@ Infinite Reality Engine. All Rights Reserved.
 */
 import { ColorResult } from '@uiw/color-convert'
 import SketchPicker from '@uiw/react-color-sketch'
-import React from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import { Color, ColorRepresentation } from 'three'
 
 import { twMerge } from 'tailwind-merge'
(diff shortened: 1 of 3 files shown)


packages/ui/src/primitives/tailwind/Color/index.tsx
```diff
@@ -52,36 +52,68 @@ export function ColorInput({
...
+        !inputRef.current.contains(event.target) &&
+        pickerRef.current &&
+        !pickerRef.current.contains(event.target)
+      ) {
+        setIsPickerOpen(false)
+        if (onRelease) onRelease(color)
+      }
+    }
+
+    document.addEventListener('mousedown', handleClickOutside)
```
SYBIOTE, line 83: can we use the client input system to handle this ?

SamMazerIR: I don't think so, no

DanielBelmes (changes requested):
Looks like this one was supposed to be based off release0.9 but isn't.

DanielBelmes:
Reopening for 0.9 base

PR #512, opened 2024-10-24
fix camera reset studio

HexaField:
## References
closes #_insert number here_


DanielBelmes (changes requested):
Can we get this based off release actually so that it's fixed for that.

DanielBelmes:
Reoppening based off 09

PR #518, opened 2024-10-24
Fix: GTM integration

jose-galvan:
## Summary
Initialize GTM using script snippets instead of using react library. Current integration has an initialization delay caused by the authentication and as a result some pages are not tagged (see alerts generated in GTM dashboard). This change builds add the GTM script at build time instead of initialization after authentication.

## References
closes  https://tsu.atlassian.net/browse/IR-4963

### DIFF
--- a/packages/client/index.html
+++ b/packages/client/index.html
@@ -75,8 +75,20 @@
         rootIframe.style.left = `${left}px`
       })
     </script>
+    <!-- Google Tag Manager -->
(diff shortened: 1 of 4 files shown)


DanielBelmes:
This is probably important for release so I might recreate to merge into release branch

DanielBelmes:
reopened for release candidate https://github.com/ir-engine/ir-engine/pull/524

PR #519, opened 2024-10-24
RC0.9 extra padding in groundPlane node editor, removing a comment in image…

DanielBelmes:
…/SourceProperties, fixing hyperactive reactivity in color picker

## References
closes #_insert number here_

### DIFF
--- a/packages/ui/src/primitives/tailwind/Color/index.tsx
+++ b/packages/ui/src/primitives/tailwind/Color/index.tsx
@@ -24,7 +24,7 @@ Infinite Reality Engine. All Rights Reserved.
 */
 import { ColorResult } from '@uiw/color-convert'
 import SketchPicker from '@uiw/react-color-sketch'
-import React from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import { Color, ColorRepresentation } from 'three'
 
 import { twMerge } from 'tailwind-merge'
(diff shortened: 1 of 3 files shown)


HexaField (dismissed):
I'm not sure if this was intention, but it's not a great experience now using the colour picker, since as soon as you click it it hides the menu. I feel like the menu shouldn't hide unless you click outside of it.

DanielBelmes:
@SamMazerIR

packages/ui/src/primitives/tailwind/Color/index.tsx
```diff
@@ -52,36 +52,68 @@ export function ColorInput({
...
   return (
     <div
+      tabIndex={0} // Make the div focusable
+      ref={inputRef} // Ref to track focus on the input
       className={twMerge(
         'relative flex h-9 items-center gap-1 rounded-lg border-none bg-[#1A1A1A] px-2 text-xs text-[#8B8B8D]',
         disabled && 'cursor-not-allowed',
         className
       )}
+      onClick={handleTogglePicker} //opens the color picker
```
MichaelEstes, line 100: I don't think this should be a toggle, it will also be called when the color picker itself is clicked, so the color picker will be closed every time you drag the color selector and release

SamMazerIR: changed

SamMazerIR:
> I'm not sure if this was intention, but it's not a great experience now using the colour picker, since as soon as you click it it hides the menu. I feel like the menu shouldn't hide unless you click outside of it.

changed

SamMazerIR (approved):

MichaelEstes (approved):

PR #530, opened 2024-10-28
Fix asset search

CITIZENDOT:
## Summary
- Reset the `skip` parameter when search string is changed.
- Merge search results only on scroll events, and not when search string is changed.

## References
closes https://tsu.atlassian.net/browse/IR-5150


aditya-mitra (approved):
Works fine ✅

PR #538, opened 2024-10-29
chore(ir-engine): select component refactor

jonathan-casarrubias:
## Summary
- Select component refactor aiming to introduce more flexible customizations
- Introduced new reverse geocoding service for auto country select
- (NOTE: Moiz's PR improvements should be integrated separated from this PR, to avoid issues)

image
 
## References
https://tsu.atlassian.net/browse/IR-5181

### DIFF
--- a/packages/ui/src/primitives/tailwind/Input/index.tsx
+++ b/packages/ui/src/primitives/tailwind/Input/index.tsx
@@ -45,6 +45,7 @@ export interface InputProps extends React.HTMLAttributes<HTMLInputElement> {
   errorBorder?: boolean
   maxLength?: number
   autoComplete?: string
+  pattern?: string
 }
 
 const variants = {
@@ -71,6 +72,7 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
       labelClassname,
       errorBorder,
       autoComplete,
+      pattern,
       ...props
     },
     ref
--- a/packages/ui/src/primitives/tailwind/Select/index.tsx
+++ b/packages/ui/src/primitives/tailwind/Select/index.tsx
@@ -36,7 +36,13 @@ import Input from '../Input'
 
 export type OptionValueType = string | number
 
-export type SelectOptionsType = { label: string; value: any; disabled?: boolean }
+export type SelectOptionsType = {
+  label: string
+  value: any
+  disabled?: boolean
+  selected?: boolean
+  metadata?: { [key: string]: string }
+}
 
 export interface SelectProps<T extends OptionValueType> {
   label?: string
@@ -57,6 +63,7 @@ export interface SelectProps<T extends OptionValueType> {
   searchDisabled?: boolean
   inputContainerClassName?: string
   endComponent?: JSX.Element
+  customMenuContent?: (SelectOptionsType) => JSX.Element
 }
 
 const Select = <T extends OptionValueType>({
@@ -77,7 +84,8 @@ const Select = <T extends OptionValueType>({
   errorBorder,
   searchDisabled,
   inputContainerClassName,
-  endComponent
+  endComponent,
+  customMenuContent
 }: SelectProps<T>) => {
   const ref = useRef<HTMLDivElement>(null)
   const menuRef = useRef<HTMLDivElement>(null)
@@ -117,11 +125,12 @@ const Select = <T extends OptionValueType>({
     }
   }
 
+  // TODO (JC): LEVERAGE IMPROVEMENTS FROM MOIZ PR
   const handleSearch = (e: React.ChangeEvent<HTMLInputElement>) => {
     if (searchDisabled) {
       return
     }
-    const newOptions = options.filter((item) => item.label.toLowerCase().startsWith(e.target.value.toLowerCase()))
+    const newOptions = options.filter((item) => item.value.toLowerCase().startsWith(e.target.value.toLowerCase()))
     if (newOptions.length > 0) {
       filteredOptions.set(newOptions)
       selectLabel.set(e.target.value)
@@ -181,7 +190,7 @@ const Select = <T extends OptionValueType>({
         ref={menuRef}
       >
         <ul
-          className={twMerge('max-h-40 overflow-auto [&>li]:px-4 [&>li]:py-2', menuClassname)}
+          className={twMerge('max-h-40 overflow-auto bg-theme-surface-main [&>li]:px-4  [&>li]:py-2', menuClassname)}
           data-testid="select-input-list"
         >
           {filteredOptions.value.map((option, index) => (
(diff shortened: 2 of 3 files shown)


packages/ui/src/primitives/tailwind/Input/index.tsx
```diff
@@ -45,6 +45,7 @@ export interface InputProps extends React.HTMLAttributes<HTMLInputElement> {
   errorBorder?: boolean
   maxLength?: number
   autoComplete?: string
+  pattern?: string
```
CITIZENDOT, line 48: Passing this prop is redundant since it's already included in [`InputHTMLAttributes<HTMLInputElement>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern)

packages/ui/src/primitives/tailwind/Input/index.tsx
```diff
@@ -71,6 +72,7 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
       labelClassname,
       errorBorder,
       autoComplete,
+      pattern,
```
CITIZENDOT, line 75: Refer to above comment.

packages/ui/src/primitives/tailwind/Select/index.tsx
```diff
@@ -36,7 +36,12 @@ import Input from '../Input'
 
 export type OptionValueType = string | number
 
-export type SelectOptionsType = { label: string; value: any; disabled?: boolean }
+export type SelectOptionsType = {
+  label: string
+  value: any
+  disabled?: boolean
+  metadata?: { [key: string]: string }
+}
```
CITIZENDOT, line 44: Going forward, we want to use `MenuItem` pattern to render Options. This will allow us to exert greater control over how the options are rendered. Something like this: https://github.com/ir-engine/ir-engine/blob/dev/packages/ui/src/primitives/tailwind/Dropdown/index.tsx.

It can be used as:
```jsx
<Select>
  <MenuItem value={...}>labelContent</MenuItem>
  <MenuItem value={...}>labelContent</MenuItem>
  <MenuItem value={...}>labelContent</MenuItem>
</Select>
```
This is inspired from: https://mui.com/material-ui/react-select/

packages/ui/src/primitives/tailwind/Select/index.tsx
```diff
@@ -77,7 +83,8 @@ const Select = <T extends OptionValueType>({
   errorBorder,
   searchDisabled,
   inputContainerClassName,
-  endComponent
+  endComponent,
+  customMenuContent
```
CITIZENDOT, line 87: Input props are changed:
Few props such as: `inputClassName`, `labelClassName`, `inputContainerClassName` are removed from the Input component.

CITIZENDOT (changes requested):

packages/ui/src/primitives/tailwind/Input/index.tsx
jonathan-casarrubias: Actually you are right, pattern already is part of the InputHTMLAttributes interface, the thing is that we are not extending from that interface, so if I remove the definition it does complain.

I just spent a little bit of time analyzing and found that strangely, you guys have been setting prop types out of that interface, rather than extending, like **type** and **onChange**....

image

Is there a reason for this? shouldn't we simply also extend from InputHTMLAttributes like:

image

Once I do that the missing pattern definition disappears, I'm just trying to understand why you guys decided to use it at prop definition rather than by extending?

Regardless, I do agree that my definition of the property is redundant and incorrect, I just want to understand the logic behind setting types out of InputHTMLAttributes (Like type and onChange) rather than extending the interface.

Cheers!.

packages/ui/src/primitives/tailwind/Select/index.tsx
jonathan-casarrubias: Right, I 100% agree that such is the best approach, by doing that it would avoid the multiple iterations currently existing within the select component, the thing is that this is a 0.9 requirement, I did spent quite a bit of time analyzing the options, in the short and the long term.

I concluded that by making completely disrupting changes would add a lot of risk for the time being, so I decided to implement something that wouldn't break all of the other current implementations of the Select component across all apps while achieving the immediate requirement.

If you analyze my approach, basically I'm by passing the current implementation with customMenu in a way that by implementing it is basically doing what you are mentioning but without breaking other Select implementations that still are using the select component's internal rendering.

I guess the question here is: Is the <Select><MenuItem></MenuItem></Select> approach a 0.9 requirement? depending on the answer comes the decision.

IF the answer is yes then we can simply migrate what's inside the customMenu.
IF the answer is no then we can use this as a temporary solution while the <Select><MenuItem></MenuItem></Select> is available.

In any case, my implementation has always been consider under my view as a low risk immediate solution for 0.9, having a subcomponent like MenuItem it definitely has been my long term solution and I wanted to revisit after 0.9.

packages/ui/src/primitives/tailwind/Select/index.tsx
jonathan-casarrubias: Ok from all of the comments, this actually is the challenging one because these are the only way I found to make styling modifications from the parent component without internally affecting the Select component, as I mentioned I choose the approach that wouldn't break other select component implementations across all apps (admin, dashboard and studio) which is the main goal I aimed here, to avoid the intrinsic risk of breaking stuff.

I guess the question again is, are these UI component refactors expected for 0.9 or for a later release?

What we need to keep in mind is that we can't make breaking changes over components that already are implemented for the given release or we'll end up breaking stuff that already works for such release version, because as I said the intrinsic risk is too high, we'll end up breaking a lot of stuff, the thing is that I have no idea if the refactor of these core ui components are for 0.9 or not.

packages/ui/src/primitives/tailwind/Input/index.tsx
CITIZENDOT: The interface above is not on latest dev (You have to merge `dev` into this branch). Here's the latest interface: https://github.com/ir-engine/ir-engine/blob/af2902ba3af133c446de6e06aba1069d5da086f4/packages/ui/src/primitives/tailwind/Input/index.tsx#L37-L73

The reason className is omitted from the Input props is, we want to stick the `Input` design to Figma design always, without any overrides on its styles.

packages/ui/src/primitives/tailwind/Select/index.tsx
CITIZENDOT: `release0.9` is released yesterday and PRs are no longer made to that branch. Nevertheless all UI refactoring PRs are targeted to dev branch (not release branch).

Reiterating again: to fix the Phone number Input quickly, I think a separate component such as `<PhoneInput />` can be designed and it can be inside multitenancy repo for now. This way, `<Select>` will be left untouched and hence not breaking any other implementations.

packages/ui/src/primitives/tailwind/Select/index.tsx
CITIZENDOT: https://github.com/ir-engine/ir-engine/pull/538#discussion_r1833708693

PR #544, opened 2024-10-29
[IR-5188] Unit Test - Fix our test-coverage tooling after change to vitest

heysokam:
## References
https://tsu.atlassian.net/browse/IR-5188


heysokam:
@HexaField Should be ready to go :+1:

I will implement the workspace setup/config in a separate PR and Jira task,
since this one was mostly to fix the tooling not working at all :writing_hand:

.vscode/launch.json
```diff
@@ -96,37 +96,37 @@
             "type": "node-terminal",
         },
         {
-            "command": "cd packages/editor && npm run test",
+            "command": "cd packages/editor && npm run test -- --maxConcurrency 1 --no-file-parallelism --bail 1 --reporter=basic",
```
HexaField, line 99: why are we declaring separate parameters here? can we include these in the config fies?

HexaField (changes requested):

.vscode/launch.json
heysokam: They restore the behavior expected when debugging with vscode. Otherwise stepping through the code is very slow and results are unpredictable and non-reproducible due to concurrency, the code not stopping on errors, and the terminal output being overwritten very aggressively by `--reporter=default`

These options only make sense in the context of vscode debugging, that's why they are here
The goal eventually is to change this to `--project=spatial.vscode` and manage these options from the workspace config instead, but this is just to get by in the meantime

HexaField:
Failing to generate coverage unfortunately :(

Screenshot from 2024-11-10 16-50-17

HexaField:
Ahh, just had to match the vitest coverage versions to vitest

HexaField (approved):

PR #553, opened 2024-10-29
RC0.9: IR 5191: fix variant crashing

dinomut1:
https://tsu.atlassian.net/browse/IR-5191

### DIFF
--- a/packages/engine/src/scene/components/ModelComponent.tsx
+++ b/packages/engine/src/scene/components/ModelComponent.tsx
@@ -105,14 +105,6 @@ function ModelReactor() {
   const [gltf, error] = useGLTF(modelComponent.src.value, entity)
 
   useEffect(() => {
-    if (modelComponent.src.value) return
-    addError(entity, ModelComponent, 'INVALID_SOURCE', 'No source provided')
-    return () => {
-      removeError(entity, ModelComponent, 'INVALID_SOURCE')
-    }
-  }, [modelComponent.src])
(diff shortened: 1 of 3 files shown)


DanielBelmes (approved):

packages/engine/src/scene/components/ModelComponent.tsx
```diff
@@ -104,14 +104,6 @@ function ModelReactor() {
 
   const [gltf, error] = useGLTF(modelComponent.src.value, entity)
 
-  useEffect(() => {
```
MichaelEstes, line 107: This was added to prevent an infinite loading screen if a model component existed in the scene without a source value (since the scene is never loaded), if we don't want to add an error for that case we may need to create a dummy scene if there's no source value

dinomut1: So would that just be returning an empty scene when passing a blank URL into useGLTF?

dinomut1: We already generate a dud Group if the gltf is blank

MichaelEstes: The GLTFComponent requires that all ModelComponent instances either have a valid ModelComponent.scene value or an error before marking a GLTF as loaded, I think this was to prevent models falling through colliders initially

the-reality-engineer (approved):

PR #561, opened 2024-10-30
Enhance asset panel scrolling and navigation

SYBIOTE:
Introduce improvements to the asset panel's scrollbar, making it relative and defining a minimum height. Enhance the navigation experience with a sticky mini navbar for better resource management.


packages/editor/src/panels/assets/assetPanel.css
aditya-mitra, line 1: we are discouraging introduction of new css files since we are shifting to tailwind
we should put a new class in tailwind `components.css`

aditya-mitra (dismissed):

packages/editor/src/panels/assets/assetPanel.css
SYBIOTE: cool

SYBIOTE: done

CITIZENDOT:
https://github.com/user-attachments/assets/3941f7de-3bd8-4329-8eea-b313c48d45d3

Hovering on the space right side to the minimap, is causing this glitch.

packages/editor/src/panels/assets/resources.tsx
```diff
@@ -205,47 +204,185 @@ function ResourceFile({ resource }: { resource: StaticResourceType }) {
...
+    switch (index) {
+      case hoveredIndex:
+        return 'white'
+      case hoveredIndex - 1:
+      case hoveredIndex + 1:
+        return 'gray-700'
+      default:
+        return 'gray-400'
+    }
+  }
```
CITIZENDOT: We should not build  tailwind classes dynamically.

> [Dynamic class names](https://tailwindcss.com/docs/content-configuration#dynamic-class-names)
The most important implication of how Tailwind extracts class names is that it will only find classes that exist as complete unbroken strings in your source files.
> If you use string interpolation or concatenate partial class names together, Tailwind will not find them and therefore will not generate the corresponding CSS:

- https://stackoverflow.com/q/69687530/12347371
- https://stackoverflow.com/q/71818458/12347371

packages/editor/src/panels/assets/resources.tsx
```diff
@@ -205,47 +204,185 @@ function ResourceFile({ resource }: { resource: StaticResourceType }) {
...
+            onClick={() => handleScrollToPage(i)}
           >
-            {resources.map((resource) => (
-              <ResourceFile key={resource.id} resource={resource as StaticResourceType} />
-            ))}
+            <span
+              className={twMerge(
+                'nav-handle duration-250 h-[1px] w-3 transition-all',
+                hoveredIndex === i ? 'w-10' : '',
+                `bg-${assignNavColor(i)}`
```
CITIZENDOT: 1. It's always preferable to choose what properties are transitionable. Because `transition-all` is slow.
2. `bg-${assignNavColor(i)`, this has to be built statically.

packages/editor/src/panels/assets/resources.tsx
```diff
@@ -205,47 +204,185 @@ function ResourceFile({ resource }: { resource: StaticResourceType }) {
...
+              )}
+            ></span>
+            <span
+              className={twMerge(
+                'nav-id w-[1em] transition-opacity duration-500 ',
+                !navBarActivated && 'opacity-0',
+                `text-${assignNavColor(i)}`
```
CITIZENDOT: `text-${assignNavColor(i)}`: this has to be built statically.

CITIZENDOT (dismissed):

packages/editor/src/panels/assets/resources.tsx
CITIZENDOT: resolved

packages/editor/src/panels/assets/resources.tsx
CITIZENDOT: resolved

packages/editor/src/panels/assets/resources.tsx
CITIZENDOT: resolved

MbfloydIR (approved):

PR #565, opened 2024-10-30
IR-5186 positional audio in studio mode

SamMazerIR:
## Summary
adding positional audio functionality in studio mode (edit and play mode), fixing some missing tooltips in MediaComponent and SpatialAudioComponent and adding a new preview element for the MediaComponent that allows selecting the media to preview as well as playing it non-spatialized in the properties node editor of MediaComponent

## References
[https://tsu.atlassian.net/browse/IR-5186](https://tsu.atlassian.net/browse/IR-5186)

### DIFF
--- a/packages/engine/src/scene/components/MediaComponent.ts
+++ b/packages/engine/src/scene/components/MediaComponent.ts
@@ -137,7 +137,7 @@ export const MediaComponent = defineComponent({
   schema: S.Object({
     controls: S.Bool(false),
     synchronize: S.Bool(true),
-    autoplay: S.Bool(true),
+    autoplay: S.Bool(false), //false = personal preference, this is super annoying when it just starts playing once added to a scene while editing
     uiOffset: S.Vec3(),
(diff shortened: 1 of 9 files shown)


packages/engine/src/scene/components/MediaComponent.ts
```diff
@@ -137,7 +137,7 @@ export const MediaComponent = defineComponent({
   schema: S.Object({
     controls: S.Bool(false),
     synchronize: S.Bool(true),
-    autoplay: S.Bool(true),
+    autoplay: S.Bool(false), //false = personal preference, this is super annoying when it just starts playing once added to a scene while editing
```
HexaField, line 140: ideally we should play but have it muted with an option to unmute all scene audio in the studio. ill bring it up with the design team

HexaField (approved):

HexaField:
Need some better design iteration on this, but it's at least more usable than before. We can iterate on this another time.

PR #566, opened 2024-10-30
[IR-5240] Unit Test - Replace `npm/assert` with `vitest/assert`

heysokam:
## Breaking Changes
#Problem1:
 DeepEqual has different behavior
`assert.deepEqual( Vector3(x,y,z), {x,y,z} )` used to work fine with `npm/assert`  
But now it fails to match its containing properties with `vitest/assert`

#Problem2:
`assert.ok` is not a type assertion, and does not remove `possibly undefined`

#Problem3:
`assert.deepEqual(null, undefined)` used to work, but now it doesn't

#Problem4:
`assert.rejects` and `assert.doesNotReject` do not exist in `vitest/assert`.  
Need to find a replacement for their behavior with the new api (or with the `vitest/expect` tools for promises)

## References
https://tsu.atlassian.net/browse/IR-5240


heysokam:
The time investment required to refactor existing test code to match the new api is not worth the effort.

Its best to recommend the new tools for newly written tests,
and keep existing code as is.

PR #572, opened 2024-10-31
Set label prop to optional

CITIZENDOT:
## References
closes #_insert number here_


jose-galvan:
I think this PR should be open against the  [release0.9](https://github.com/ir-engine/ir-engine/tree/release0.9) branch, @DanielBelmes can you confirm?

jonathan-casarrubias (approved):

PR #576, opened 2024-10-31
Ground plane becomes visible when primitive geometry is added to the scene

MichaelEstes:
## Summary
Make ground plane mesh on separate entity
Remove unneeded component functions 
## References
closes #_insert number here_


HexaField (changes requested):
this essentially removes the distinction between the visible component and the visible flag. the visible flag allows the ground plane to receive shadows (necessary for AR and some scene floor colliders). ill discuss how we should tackle this on jira

HexaField:
Maybe we just need to ensure the color is black when it's a shadow material? Or it's just a bug in the reactor logic

packages/engine/src/scene/components/GroundPlaneComponent.ts
```diff
@@ -48,45 +57,38 @@ export const GroundPlaneComponent = defineComponent({
...
-  },
-
   reactor: function () {
     const entity = useEntityContext()
 
     const component = useComponent(entity, GroundPlaneComponent)
 
-    const getMaterial = (): MeshLambertMaterial | ShadowMaterial => {
-      return component.visible.value ? new MeshLambertMaterial() : new ShadowMaterial({ opacity: 0.5 })
+    const meshEntity = useHookstate(() => {
```
HexaField, line 65: are we able to keep the mesh entity on the current entity?

HexaField (changes requested):

packages/engine/src/scene/components/GroundPlaneComponent.ts
MichaelEstes: QA was adding the ground plane to a entity that already had a MeshComponent on it which is what was causing the main issues, there's nothing yet to restrict adding a component if uses components that are already on the entity.

HexaField: Yeah ok, I think we should hold off on trying to fix that category of problem since we're going to refactoring that stuff soon, and need a general solution.

PR #577, opened 2024-10-31
IR-5194 mount point fixes

SamMazerIR:
## Summary
fixing issues with mountpoint when exiting play mode while sitting (cleanup of MountPointState when entities are deleted)
also fixing issue with mount point interactable xrui positioning if there is no bounding box

## References
[https://tsu.atlassian.net/browse/IR-5194](https://tsu.atlassian.net/browse/IR-5194)

## QA Steps
Drag and drop a chair
Add entity → empty
Drag the new entity into the asset in the hierarchy
On this child object, add component → Mount Point
Enter preview and press E at the mount point’s location (could be 0,0)
Observe character enters seated position
Exit preview
Use the translate tool to move the child entity to the model
Reenter preview, walk up the chair and press E
Continue moving this mount point component and observe how it now still functions


AidanCaruso (dismissed):
This fixes the issue, but causes issues with the isOccupied check based on the state map. See ` if (getState(MountPointState)[mountPointUUID] || hasComponent(avatarEntity, SittingComponent)) return` in the `mountEntity` function. MountPointState no longer directly returns the record object so you'll have to change this to access the relevant map property instead.

HexaField (approved):

PR #584, opened 2024-11-01
IR-5272 Update user created prefab file extension

MichaelEstes:
## Summary
Change user created prefabs to use the same file extension as "Add Entity" prefabs, this makes the user created prefabs go through the flow where they're automatically dereferenced when added to a scene which prevents multiple user created prefabs from being added to the scene with the same UUID.

## References
closes #_insert number here_


dinomut1:
I don't think we should do this. User-created prefabs should not be dereferenced by default

PR #588, opened 2024-11-04
Add projectId population logic to locationPatchResolver

MoizAdnan:


barankyle:
One of the location tests is failing, probably because of the change in this PR. Not sure if that indicates an unintended problem or if the test needs to be changed to match this.

barankyle (approved):

hanzlamateen (approved):

PR #590, opened 2024-11-04
IR-5204 Stop overwriting non-authority rigidbodies

AidanCaruso:
## References
closes #_insert number here_


packages/spatial/src/physics/classes/Physics.ts
```diff
@@ -382,6 +383,7 @@ function updateRigidbodyPose(entities: Entity[]) {
     if (!world) continue
     const body = world.Rigidbodies.get(entity)
     if (!body) continue
+    if (hasComponent(entity, NetworkObjectComponent) && !hasComponent(entity, NetworkObjectAuthorityTag)) continue
```
heysokam, line 386: Can we add a test case for this around here? 
https://github.com/ir-engine/ir-engine/blob/139a060ca21b7055f6cf96bc3aab0eff56dab0bb/packages/spatial/src/physics/classes/Physics.test.tsx#L1191

> should not do anything for an entity when the entity has a NetworkObjectComponent and it does not have a NetworkObjectAuthorityTag

AidanCaruso: Good point

packages/spatial/src/physics/classes/Physics.test.tsx
```diff
@@ -1215,7 +1216,24 @@ describe('Physics : Rapier->ECS API', () => {
         removeEntity(testEntity)
         return destroyEngine()
       })
-
+      it('should not do anything for an entity that has a NetworkObjectComponent but no NetworkAuthorityComponent', () => {
```
HexaField: Let's be clearer with the language here, rather than "should not do anything" let's say "should not update pose and velocity"

HexaField (changes requested):

HexaField (approved):

PR #600, opened 2024-11-05
Fix spawn points not always working

HexaField:
## Summary

Add transform component as a dependency of spawning avatar to ensure pose is correct

## References
closes #_insert number here_


DanielBelmes (approved):

DanielBelmes:
Gotta reopen for tests to run annoyingly

PR #609, opened 2024-11-05
locked the drag and drop option behind a feature flag

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-5192
## References
closes #_insert number here_


HexaField (changes requested):
we don't need a new feature flag for this, we can reuse the point and click flag since the button is a toggle between the two settings, and drag and drop is the default / desirable state

MbfloydIR:
> we don't need a new feature flag for this, we can reuse the point and click flag since the button is a toggle between the two settings, and drag and drop is the default / desirable state

implemented

aditya-mitra (approved):
Looks good!

HexaField (approved):

PR #620, opened 2024-11-06
Fix gizmo rotation synchronization with UI

SYBIOTE:
Ensure that the gizmo rotation correctly syncs with the UI by updating the angle state based on the quaternion input.


packages/ui/src/components/editor/input/Euler/index.tsx
```diff
@@ -51,13 +51,20 @@ const getBoundedRoundedAngle = (angle: number) => {
...
+const tempQuat = new Quaternion()
 export const EulerInput = (props: EulerInputProps) => {
+  tempQuat.copy(props.quaternion)
   tempEuler.setFromQuaternion(props.quaternion, 'YXZ')
   const angle = useHookstate({
     x: getBoundedRoundedAngle(tempEuler.x * RAD2DEG),
     y: getBoundedRoundedAngle(tempEuler.y * RAD2DEG),
     z: getBoundedRoundedAngle(tempEuler.z * RAD2DEG)
   })
+  angle.set({
```
HexaField: it doesn't make sense to have a hookstate that is immediately set. we can avoid it entirely and use the values directly

HexaField (changes requested):

packages/ui/src/components/editor/input/Euler/index.tsx
SYBIOTE: sure, ill test if that works

SYBIOTE: it works , also fixed an issue, material entities do not have a transform component associated with them but do have an entitytree, therefore errors popup when we compute transforms for the material entities parent but fail for the material entity itsef as it doesn not have a tranform component, so i made sure transforms are only computed for children entities which have a transform component

packages/ui/src/components/editor/input/Euler/index.tsx
```diff
@@ -53,48 +52,42 @@ const getBoundedRoundedAngle = (angle: number) => {
 const tempEuler = new Euler() // we need the persistance, the hookstate doesnt register the dynamically allocated euler and quat value otherwise, thus we cannot assign new variable to the same
 export const EulerInput = (props: EulerInputProps) => {
   tempEuler.setFromQuaternion(props.quaternion, 'YXZ')
-  const angle = useHookstate({
+  const angle = {
```
HexaField, line 55: this change doesnt make any sense, why is should this no longer be reactive?

HexaField (changes requested):

packages/ui/src/components/editor/input/Euler/index.tsx
SYBIOTE: Setting the values directly wasnt actually being applied in the hookstate, setting them explicitly seemed to work, i checked if we needed the hookstate at all and after some testing it didnt seem necessary

SYBIOTE: the component, itself is using the quat as prop, there are two cases
1. we use the gizmo to change quat
2. we use the scrubber to change euler

in case one 
the component gets rerendered rapidly as the prop changes, so we always have the fresh angle value and thus appropriate values in the input fields 
in case two 
the callback changes the quat causing a rerender, which in turn gives us a fresh angle value

in either case, the angle value itself does not need to be reactive   
@HexaField

HexaField (approved):

PR #621, opened 2024-11-06
updated logic to validate the time and check if it is before the star…

MbfloydIR:
…t time, so it can be adjusted

## Summary
https://tsu.atlassian.net/browse/IR-5096
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/scene/components/MediaComponent.ts
+++ b/packages/engine/src/scene/components/MediaComponent.ts
@@ -193,11 +193,10 @@ export function MediaReactor() {
   if (!isClient) return null
 
   function validateTime() {
-    const mediaElementState = getMutableComponent(entity, MediaElementComponent)
-    const element = mediaElementState.element.value as HTMLMediaElement
-    console.log('element.currentTime = ' + element.currentTime)
+    const mediaElementComponent = getMutableComponent(entity, MediaElementComponent)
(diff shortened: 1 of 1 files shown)


packages/engine/src/scene/components/MediaComponent.ts
```diff
@@ -192,6 +192,15 @@ export function MediaReactor() {
 
   if (!isClient) return null
 
+  function validateTime() {
+    const mediaElementState = getMutableComponent(entity, MediaElementComponent)
```
SamMazerIR: i'd rename mediaElementState to mediaElementComponent here

packages/engine/src/scene/components/MediaComponent.ts
```diff
@@ -192,6 +192,15 @@ export function MediaReactor() {
...
+    const element = mediaElementState.element.value as HTMLMediaElement
+    console.log('element.currentTime = ' + element.currentTime)
```
SamMazerIR: remove console.log

MbfloydIR: implemented

packages/engine/src/scene/components/MediaComponent.ts
MbfloydIR: implemented

SamMazerIR (approved):

PR #624, opened 2024-11-07
fix for default material metalness value

HexaField:
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/gltf/GLTFState.tsx
+++ b/packages/engine/src/gltf/GLTFState.tsx
@@ -30,6 +30,7 @@ import {
   AnimationMixer,
   Bone,
   BufferGeometry,
+  FrontSide,
   Group,
   LoaderUtils,
   MathUtils,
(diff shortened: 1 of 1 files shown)


packages/engine/src/gltf/GLTFState.tsx
```diff
@@ -977,7 +978,16 @@ const CameraReactor = (props: { nodeIndex: number; documentID: string; entity: E
   return null
 }
 
-const defaultMaterial = () => new MeshStandardMaterial()
+const defaultMaterial = () =>
+  new MeshStandardMaterial({
+    color: 0xffffff,
+    emissive: 0x000000,
+    metalness: 1,
```
dinomut1, line 985: metalness should be 0 instead of 1 i think

dinomut1 (dismissed):

packages/engine/src/gltf/GLTFState.tsx
HexaField: GLTF specifies it should be 1 by default. same thing exists in the threejs GLTF loader. This is why certain things were loading with different colours than before.

HexaField: https://github.com/ir-engine/ir-engine/blob/fix-for-default-material-metalness-value/packages/engine/src/assets/loaders/gltf/GLTFLoaderFunctions.ts#L440

DanielBelmes: https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-material-pbrmetallicroughness I believe he is right yea

DanielBelmes (approved):
Approved on my front. Does seem to match default material of GLTF.

dinomut1:
not resolved. metalness should never be 1 by default

PR #625, opened 2024-11-07
gltf assign final material parity

HexaField:
## References
closes #_insert number here_


packages/engine/src/gltf/GLTFState.tsx
```diff
@@ -1136,6 +1140,25 @@ const MaterialInstanceReactor = (props: {
     else materialInstance.uuid.set([materialUUID])
   }, [materialEntity, primitive.material])
 
+  const material = useOptionalComponent(materialEntity, MaterialStateComponent)?.material
+  const useDerivativeTangents = primitive.attributes.tangent === undefined
+  const useVertexColors = primitive.attributes.color !== undefined
+  const useFlatShading = primitive.attributes.normal === undefined
```
AidanCaruso: These will always evaluate to undefined because the primitive attribute properties are in all caps.
image

HexaField: the caps are once they are translated to threejs, in this form they are in the gltf format, which is lower case.

HexaField: hmm ok you may be right :/

HexaField: ahh, in the threejs loader, it was looking for parameters on the geometry. this is now on the gltf document. my mistake. good catch.

HexaField: Screenshot from 2024-11-15 09-55-33
and look at that, vertex colors working!

AidanCaruso (approved):

PR #628, opened 2024-11-07
IR-5377: Fix User Table showing single account identifier for all rows

MoizAdnan:
https://tsu.atlassian.net/browse/IR-5377


MoizAdnan (commented):
is `action : admin` needed?

hanzlamateen (approved):

PR #634, opened 2024-11-07
Import lint rules

DanielBelmes:
## Summary
Starting to implement some eslint rules that can help detect import cycles

## References
closes #_insert number here_


DanielBelmes:
One issue so far is that it's not really understanding the difference between export and export type which is weird since I am using a typescript resolver so it should understand.

PR #635, opened 2024-11-07
set metalness to 1

HexaField:
## References
closes #_insert number here_


DanielBelmes (changes requested):
Just putting requested changes until we can finalize discussion. We've had some confusion on how this will effect artists. Most likely this won't as it's the fallback material if no material is defined.

PR #636, opened 2024-11-07
IR-5415 Fix animation loading and clean up loop animation component logic

AidanCaruso:
## References
closes #_insert number here_


MbfloydIR (approved):

HexaField (commented):
Let's definitely get some basic regression tests written for this

AidanCaruso:
> Let's definitely get some basic regression tests written for this

Added two new test cases for animation packs, and single animation clip gltfs with tracks that bind to non-skeletal child nodes, which are the two cases impacted/fixed by this PR

packages/engine/src/avatar/functions/moveAvatar.test.tsx
```diff
@@ -87,16 +86,16 @@ describe('moveAvatar function tests', () => {
         parentUUID: getComponent(sceneEntity, UUIDComponent),
         position: new Vector3(),
         rotation: new Quaternion(),
-        entityUUID: Engine.instance.userID as string as EntityUUID,
+        entityUUID: Engine.instance.store.userID as string as EntityUUID,
```
HexaField: i think we need to append `_avatar` to the end of the entityUUID for it to be aligned with how avatars exist at runtime, then we can use `AvatarComponent.getUserAvatarEntity()` correctly

packages/engine/src/avatar/functions/moveAvatar.test.tsx
```diff
@@ -128,16 +127,16 @@ describe('moveAvatar function tests', () => {
...
```
HexaField: same here

packages/engine/src/avatar/functions/moveAvatar.test.tsx
```diff
@@ -128,16 +127,16 @@ describe('moveAvatar function tests', () => {
...
         name: ''
       })
     )
 
     applyIncomingActions()
 
-    spawnAvatarReceptor(Engine.instance.userID as string as EntityUUID)
-    const entity = AvatarComponent.getUserAvatarEntity(Engine.instance.userID)
+    const entity = UUIDComponent.getEntityByUUID(Engine.instance.store.userID as string as EntityUUID)
+    spawnAvatarReceptor(Engine.instance.store.userID as string as EntityUUID)
```
HexaField: i wonder if instead of running spawnAvatarReceptor here we should ensure the event sourcing state reactors work properly (which call it at runtime)

HexaField (changes requested):

packages/engine/src/avatar/functions/moveAvatar.test.tsx
AidanCaruso: The useEffect that runs this post event sourcing is locked behind an isClient check. Going to try removing it and see what happens. Otherwise will need to stick with manually calling spawnAvatarReceptor

AidanCaruso: Seems okay with the check removed

HexaField: Aha, yeah should be fine to remove that check.

HexaField (approved):

PR #638, opened 2024-11-08
Merge GroupComponent and ObjectComponent

HexaField:
## References
closes #_insert number here_


DanielBelmes:
Imma convert to draft while these test are failing. They definitely seem related

packages/spatial/src/input/functions/ClientInputHeuristics.ts
```diff
@@ -168,8 +168,8 @@ export function findEditor(intersectionData: Set<IntersectionData>, caster: Rayc
   const inputObj = inputObjectsQuery().concat(cameraGizmo)
 
   const objects = (pickerObj.length > 0 ? allGizmos : inputObj) // gizmo heuristic
-    .map((eid) => getComponent(eid, GroupComponent))
-    .flat()
+    .map((eid) => getComponent(eid, ObjectComponent))
+  console.log({ objects })
```
DanielBelmes: console.log

DanielBelmes (dismissed):
- Transforms no longer working from the transform properties
- Transform Gizmo doesn't appear
- Asset loading circle stays in screen(might be related to gizmo)

That's it so far without being able to continue with transform

HexaField:
setting to draft, still some problems with gizmos, but they are now completely entity driven at least

packages/spatial/src/common/debug/LightHelperComponent.ts
```diff
@@ -1,90 +0,0 @@
...
-    if (matchesColor.test(json.color)) component.color.set(json.color)
-  },
-
-  reactor: function () {
-    const entity = useEntityContext()
-    const component = useComponent(entity, LightHelperComponent)
-    const light = component.light.value as Light<LightShadow<Camera>>
-    const [helper] = useDisposable(getLightHelperType(light), entity, light, component.size.value)
-    useHelperEntity(entity, component, helper)
-    helper.update()
```
MichaelEstes, line 71: I don't know if the update functions not being called anymore is causing this, but moving objects after helpers have been enabled is causing incorrect transformations on the helper objects

packages/engine/src/scene/components/ParticleSystemComponent.ts
```diff
@@ -73,7 +73,7 @@ import {
   useHookstate
 } from '@ir-engine/hyperflux'
 import { NameComponent } from '@ir-engine/spatial/src/common/NameComponent'
-import { addObjectToGroup, removeObjectFromGroup } from '@ir-engine/spatial/src/renderer/components/GroupComponent'
+import { addObjectToGroup, removeObjectFromGroup } from '@ir-engine/spatial/src/renderer/components/ObjectComponent'
```
MichaelEstes: We'll probably need a different solution for the particle system component since the particle system library relies on the three hierarchy, so particle systems are not showing up on this branch

DanielBelmes:
Not able to reproduce consistently but I asked @MichaelEstes and he also saw this. Sometimes transform gizmo just stops working and you can't transform things. Still smoke testing though.

packages/engine/src/scene/components/ParticleSystemComponent.ts
HexaField: dang, we can do the same thing as XRUI where we manually update the hierarchy in a system

HexaField: particles dont seem to be working on dev, so i'm not sure what the correct fix is here

HexaField:
A few things were busted from proxifying all objects. XRUI and particles still use three hierarchies, so now there is a `preserveChildren` option that can be assigned to an object to avoid proxifying children & parent fields

AidanCaruso (changes requested):
Overall looks good, however I'm noticing two avatar related issues. First, tpose enforcement is now calculated incorrectly for non-vrm rigs. 
image

Other issue is that everything unloads whenever you try to switch avatars.

AidanCaruso:
Fixed avatar bone rotations in the commit above, switching avatars unloading everything is still a problem though

AidanCaruso (approved):

PR #640, opened 2024-11-08
prevent circular dependency from crashing /admin pages

dtlehrer:
## Summary
quick fix for this error that appears when trying to load /admin routes:
image
image

## References
closes #_insert number here_


dtlehrer:
it sounds like defining `groupQuery` inside of `execute` could be problematic, so closing this PR and waiting for wider architectural solution to be implemented.. will continue working from this branch until then

PR #650, opened 2024-11-12
feat(studio): create a project history after update the static resource [IR-5106]

mikeplascdev:
## Summary

- Add to the project history when the user adds a new asset thumbnail

## References
closes #_insert number here_
https://github.com/ir-engine/ir-engine/blob/87adccb126a55f1eb31296bad8f81849ad9bcc14/packages/client-core/src/common/services/FileThumbnailJobState.tsx#L171

## QA Steps

1. Open Studio
2. Open a scene
3. Navigate to the Files panel
4. Upload a file from the local machine
5. Observe thumbnail creation for an uploaded file
6. Navigate to Dashboard
7. Click History on the corresponding project
8. Check logs including file uploading action
9. Check logs including thumbnail creation action for uploaded file

## Jira
[IR-5106](https://tsu.atlassian.net/browse/IR-5106)

[IR-5106]: https://tsu.atlassian.net/browse/IR-5106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/server-core/src/media/static-resource/static-resource.hooks.ts
```diff
@@ -292,6 +292,36 @@ const addDeleteLog = async (context: HookContext<StaticResourceService>) => {
   }
 }
 
+const addPatchLog = async (context: HookContext<StaticResourceService>) => {
```
jose-galvan: How is this different from addDeleteLog? Is there a reason the same code can't be reused?

jose-galvan (changes requested):

packages/server-core/src/media/static-resource/static-resource.hooks.ts
mikeplascdev: only the resource action, I can update the delete log to add the action in the params

hanzlamateen (commented):
Agree with what Jose suggested. Otherwise it looks good.

packages/server-core/src/media/static-resource/static-resource.hooks.ts
mikeplascdev: done c228f29661e37949471c1b41e2b79e38a1b0b18c

hanzlamateen (approved):

PR #654, opened 2024-11-12
ensure search on collapsed nodes

SYBIOTE:
## References
closes #_insert number here_


packages/editor/src/panels/hierarchy/helpers.ts
```diff
@@ -128,25 +139,32 @@ function buildHierarchyTree(
...
     isLeaf: !(node.children && node.children.length > 0),
-    lastChild: lastChild
+    lastChild: lastChild,
+    isRendered: isRendered
   }
   array.push(item)
 
-  if (hasComponent(entity, GLTFComponent) && showModelChildren) {
-    const scene = GLTFComponent.getInstanceID(entity)
+  if (hasComponent(entity, ModelComponent) && showModelChildren) {
```
HexaField: please revert. we no longer use the ModelComponent, it is deprecated and only exists for comparison.

HexaField (changes requested):

packages/editor/src/panels/hierarchy/helpers.ts
SYBIOTE: roger that

SYBIOTE: done

packages/editor/src/panels/hierarchy/helpers.ts
```diff
@@ -128,25 +139,32 @@ function buildHierarchyTree(
...
     isLeaf: !(node.children && node.children.length > 0),
-    lastChild: lastChild
+    lastChild: lastChild,
+    isRendered: isRendered
   }
   array.push(item)
 
   if (hasComponent(entity, GLTFComponent) && showModelChildren) {
-    const scene = GLTFComponent.getInstanceID(entity)
+    const modelSceneID = getModelSceneID(entity)
```
HexaField: need to also revert the usage of getModelSceneID, it is also not used now.

HexaField (changes requested):

packages/editor/src/panels/hierarchy/helpers.ts
SYBIOTE: updating

HexaField (approved):

PR #655, opened 2024-11-12
IR-5464: Feature flag names as unique ids

Rezmason:
Feature flags were getting duplicated in rare circumstances (related to how quickly a user changed a feature flag in the admin panel). We now assign flag IDs to the id field of the table, guaranteeing uniqueness.

https://tsu.atlassian.net/browse/IR-5464

### DIFF
--- a/packages/server-core/src/setting/feature-flag-setting/migrations/20241105001200_primary_flagNames.ts
+++ b/packages/server-core/src/setting/feature-flag-setting/migrations/20241105001200_primary_flagNames.ts
@@ -1,0 +1,109 @@
+/*
+CPAL-1.0 License
+
+The contents of this file are subject to the Common Public Attribution License
+Version 1.0. (the "License"); you may not use this file except in compliance
+with the License. You may obtain a copy of the License at
+https://github.com/ir-engine/ir-engine/blob/dev/LICENSE.
+The License is based on the Mozilla Public License Version 1.1, but Sections 14
+and 15 have been added to cover use of software over a computer network and 
+provide for limited attribution for the Original Developer. In addition, 
+Exhibit A has been modified to be consistent with Exhibit B.
+
+Software distributed under the License is distributed on an "AS IS" basis,
+WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for the
+specific language governing rights and limitations under the License.
+
+The Original Code is Infinite Reality Engine.
+
+The Original Developer is the Initial Developer. The Initial Developer of the
+Original Code is the Infinite Reality Engine team.
+
+All portions of the code written by the Infinite Reality Engine team are Copyright © 2021-2023 
+Infinite Reality Engine. All Rights Reserved.
+*/
+
+import {
+  FeatureFlagSettingType,
+  featureFlagSettingPath
+} from '@ir-engine/common/src/schemas/setting/feature-flag-setting.schema'
+import type { Knex } from 'knex'
+import { v4 as uuidv4 } from 'uuid'
+
+type OldFeatureFlagSettingType = FeatureFlagSettingType & {
+  flagName: string
+}
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function up(knex: Knex): Promise<void> {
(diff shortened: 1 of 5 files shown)


Rezmason:
Aw dang! This fails to rollback. Something about foreign keys? 😢

```
alter table `feature-flag-setting` drop foreign key `feature_flag_setting_userid_foreign` - 
ER_CANT_DROP_FIELD_OR_KEY: Can't DROP FOREIGN KEY `feature_flag_setting_userid_foreign`;
check that it exists
```

Do I need to do something like this in my migration?
```JS
table.foreign('userId').references('id').inTable('user').onDelete('SET NULL').onUpdate('CASCADE')
```

barankyle:
> Aw dang! This fails to rollback. Something about foreign keys? 😢
> 
> ```
> alter table `feature-flag-setting` drop foreign key `feature_flag_setting_userid_foreign` - 
> ER_CANT_DROP_FIELD_OR_KEY: Can't DROP FOREIGN KEY `feature_flag_setting_userid_foreign`;
> check that it exists
> ```
> 
> Do I need to do something like this in my migration?
> 
> ```js
> table.foreign('userId').references('id').inTable('user').onDelete('SET NULL').onUpdate('CASCADE')
> ```

This might be avoided by adding this line as the first line of the `down` function:

`await knex.raw('SET FOREIGN_KEY_CHECKS=0')`

and then as the last line of that function insert

`await knex.raw('SET FOREIGN_KEY_CHECKS=1')`

Rezmason:
> This might be avoided by adding this line as the first line of the `down` function:
> 
> `await knex.raw('SET FOREIGN_KEY_CHECKS=0')`
> 
> and then as the last line of that function insert
> 
> `await knex.raw('SET FOREIGN_KEY_CHECKS=1')`

Yes, and,

I was also declaring the `uuid` field improperly. I should have done this:
```TS
table.uuid('userId', 36).collate('utf8mb4_bin').nullable().index()
```

but because it wouldn't compile, I instead did this:
```TS
table.uuid('userId'.nullable()
```
—thinking that it was due to the difference between `alterTable` and `createTable`. I was wrong! And to get the first one to compile, I just had to prefix it with `//@ts-ignore` like in the earlier migration.

`dev-reinit`, `migrate:rollback` and `migrate` all run happily now. Just dev tested this and one other fix. Ready for review, for real!

packages/server-core/src/setting/feature-flag-setting/migrations/20241105001200_primary_flagNames.ts
```diff
@@ -0,0 +1,144 @@
...
+  await knex.raw('SET FOREIGN_KEY_CHECKS=0')
+
+  migrateUp: {
+    const flagNameColumnExists = await knex.schema.hasColumn(featureFlagSettingPath, 'flagName')
+
+    if (!flagNameColumnExists) {
+      break migrateUp
+    }
+
+    const oldRows = await knex.select<OldFeatureFlagSettingType[]>().from(featureFlagSettingPath)
```
HexaField, line 53: rather than persisting the data in memory and recreating the table, we can just migrate the columns in question. @hanzlamateen do you have any opinions on this?

HexaField (changes requested):

packages/server-core/src/setting/feature-flag-setting/migrations/20241105001200_primary_flagNames.ts
hanzlamateen: Yes @Rezmason, I think we can use alter table. And then makes the updates. renameColumn and dropColumn.

Reference:
https://github.com/theinfinitereality/irpro-multitenancy/blob/97dfa0194905d9e87e3a354d17517a274632b243/services/account/account-slug-generator/migrations/20240919180000_rename_customerId.ts#L14

https://github.com/ir-engine/ir-engine/blob/b1995ec4135d23901f4f3cda15dddb55d67fa6ee/packages/server-core/src/matchmaking/match-instance/migrations/20230425043432_match-instance.ts#L43

hanzlamateen (changes requested):

PR #658, opened 2024-11-13
Run npm install before projectEventHooks call

barankyle:
## Summary

New projects may use dependencies that are not present in the core engine for their onInstall logic. We therefore need to run npm install before that so that those project-specific dependencies are installed on the pod running updateProject.

## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/projects/project/project-helper.ts
+++ b/packages/server-core/src/projects/project/project-helper.ts
@@ -83,6 +83,7 @@ import { getFileKeysRecursive } from '../../media/storageprovider/storageProvide
 import { createStaticResourceHash } from '../../media/upload-asset/upload-asset.service'
 import logger from '../../ServerLogger'
 import { ServerState } from '../../ServerState'
+import { execPromise } from '../../util/execPromise'
 import { getContentType } from '../../util/fileUtils'
 import { getGitConfigData, getGitHeadData, getGitOrigHeadData } from '../../util/getGitData'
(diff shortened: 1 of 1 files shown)


packages/server-core/src/projects/project/project-helper.ts
```diff
@@ -1517,6 +1518,7 @@ export const updateProject = async (
     )
   }
   // run project install script
+  await execPromise(`npm install`, { cwd: appRootPath.path })
```
HexaField, line 1521: i wonder if instead of this, we should only do this after (or as part of) a successful build, since if the project hook relies on services it specifies, it would throw errors (that could result in worse problems)

barankyle: I may have slightly misidentified the problem in the description. The problem we're encountering is that projectEventHooks in the MT project imports a file that imports some AWS SDKs that are only in the MT project, not in the core engine. onInstall doesn't use them, nor is it calling any of the services that are part of the project, but the import itself fails because of those downstream missing dependencies. I didn't see that problem when I ran this branch on QAT.

HexaField: Yeah okay, I am identifying a whole class of issues we may run into in the future. This solution is a good stopgap for now.

HexaField (approved):

hanzlamateen (approved):

PR #660, opened 2024-11-13
IR-4723-Validate user during SSO sign-in process 

mshafiqmk:
## Summary
Added a message prompting users to use the signup page for GitHub, Google and Apple SSO authentication.

This prevents users from attempting to sign in through GitHub on the sign-in page, directing them to the appropriate signup process instead.

## References

[Jira Ticket 4723 ](https://tsu.atlassian.net/browse/IR-4723)
[MT-Project-PR 610](https://github.com/theinfinitereality/irpro-multitenancy/pull/610)


barankyle (approved):
I fixed a bug I'd found in Google login when I was testing.

I saw that validateSignInUser was checking against the email associated with the OAuth signin. I'm not 100% sure if every OAuth signin will have an email, or if there's any way for e.g. multiple GitHub accounts to have the same email. Neither is likely to be a problem, but I thought it was safer to instead search on the identity-provider's `token`, assembling that using getEntityQuery. It's guaranteed to return something, and it will definitely match one and only one existing identity-provider if it's there.

hanzlamateen (approved):

PR #661, opened 2024-11-13
[IR-5278, 5658] respect default import path and refresh files on import file

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-5278
https://tsu.atlassian.net/browse/IR-5658

## References
closes #_insert number here_


packages/editor/src/panels/assets/topbar.tsx
```diff
@@ -77,9 +78,11 @@ const ViewModeSettings = () => {
 
 const uploadFiles = () => {
   const projectName = getState(EditorState).projectName
+  const importFolder = getState(ImportSettingsState).importFolder || '/assets/'
+
   return inputFileWithAddToScene({
     projectName: projectName as string,
-    directoryPath: `projects/${projectName}/assets/`
+    directoryPath: `projects/${projectName}${importFolder}`
```
aditya-mitra, line 92: This will allow importing to any folder inside the project. We should allow uploading only inside `public` and `assets` folders.
@HexaField what are your thoughts?

HexaField: Yeah we need to limit to just these folders

achen5671: fixed. We're only allowed to upload to public and assets folder. We also needed to check whether the path ends with '/'

kattsushi (approved):
LGTM

PR #662, opened 2024-11-13
Michael/ir 5285 scene preview thumbnail

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-5285
## References
closes #_insert number here_


packages/ui/src/components/editor/properties/scene/previewCamera/index.tsx
```diff
@@ -58,18 +64,58 @@ const scene = new Scene()
...
   const { t } = useTranslation()
   const [bufferUrl, setBufferUrl] = useState<string>('')
+  const [blob, setBlob] = useState<Blob>()
   const transformComponent = useComponent(Engine.instance.cameraEntity, TransformComponent)
 
   const onSetFromViewport = () => {
     const { position, rotation } = getComponent(Engine.instance.cameraEntity, TransformComponent)
     const transform = getComponent(props.entity, TransformComponent)
+    const camera = getComponent(props.entity, ScenePreviewCameraComponent).camera
+    camera.position.copy(position)
```
DanielBelmes: @HexaField So this is a piece I had to help with. It seems like commitTransformSave wasn't updating the scene preview camera fast enough. So I asked Michael to update the three object directly. Are there alternative ways to get this done?

DanielBelmes: For context there's a button to set scene preview to the viewport and then update the screenshot. It sends a transform update but the screenshot happens instantly after which I think is just happening before SnapshotSystem, TransformSystem, etc....

DanielBelmes: Maybe a useEffect is the answer detecting if the ScenePreviewCamera has moved.

HexaField: ah this makes sense. yeah just update the position and then call `computeTransformMatrix`

HexaField: sorry just getting back to this now

HexaField (approved):
i merged dev in, which includes rahul's fixes, and this works really well now :)

PR #664, opened 2024-11-14
Trimming job names to make sure they do not exceed k8s limits

barankyle:
## Summary

Resolves IR-5113

## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/projects/project/project-helper.ts
+++ b/packages/server-core/src/projects/project/project-helper.ts
@@ -76,7 +76,7 @@ import { BUILDER_CHART_REGEX } from '@ir-engine/common/src/regex'
 import { Application } from '../../../declarations'
 import config from '../../appconfig'
 import { getPodsData } from '../../cluster/pods/pods-helper'
-import { getJobBody } from '../../k8s-job-helper'
+import { getJobBody, getValidPodName } from '../../k8s-job-helper'
 import { getStats, regenerateProjectResourcesJson } from '../../media/static-resource/static-resource-helper'
 import { getStorageProvider } from '../../media/storageprovider/storageprovider'
 import { getFileKeysRecursive } from '../../media/storageprovider/storageProviderUtils'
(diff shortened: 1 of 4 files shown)


packages/server-core/src/projects/project/project-helper.ts
```diff
@@ -1170,7 +1170,7 @@ export const createOrUpdateProjectUpdateJob = async (app: Application, projectNa
   if (k8BatchClient) {
     try {
       await k8BatchClient.patchNamespacedCronJob(
```
hanzlamateen, line 1172: @barankyle I am just wandering what if, inside `patchNamespacedCronJob` & `deleteNamespacedCronJob`, etc we trim the name parameter instead of doing it here? Just like we are doing in `getJobBody` method.

Or else, maybe introduce a method in k8s-job-helper `getValidPodName` and then call it where we are passing names instead of manually trimming them based on 52 length at various places?

hanzlamateen (changes requested):

packages/server-core/src/projects/project/project-helper.ts
barankyle: patch/deleteNamespacedronJob are functions of the k8s SDK, not ones we wrote, so we can't change their behavior. But you're right about making the trimming a single function, which I did.

barankyle: Slicing it to 52 characters is from [k8's documentation on cronjob names](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#:~:text=Even%20when%20the%20name%20is,no%20more%20than%2063%20characters.). It's technically only a limit for cronjob names, not job names, but I figured it couldn't hurt to just use that everywhere. 

The names are mostly a formality, so that someone looking in Lens can see what the job was doing. If we crop a few letters off a project job name unnecessarily, it's not going to affect anything.

hanzlamateen (approved):

PR #666, opened 2024-11-14
Fix Project Modal style

CITIZENDOT:
## Summary
Fix regressions caused by Input component update.

## References
closes https://tsu.atlassian.net/browse/IR-5537


DanielBelmes:
Before:
image
After:
image

Was smoke testing and it doesn't seem like all the inputs there were fixed. Or does this match the figma?

hanzlamateen (approved):

hanzlamateen:
> Before: image After: image
> 
> Was smoke testing and it doesn't seem like all the inputs there were fixed. Or does this match the figma?

@DanielBelmes I think this PR addresses the project dialog instead of location one. I think we will need a separate ticket for Location dialog.

DanielBelmes:
Ah I am blind. Thank you Hanzla

PR #671, opened 2024-11-14
IR-5479 Models should be hidden until they are fully loaded

MichaelEstes:
## References
closes #_insert number here_


packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -199,14 +200,19 @@ export const GLTFComponent = defineComponent({
   reactor: () => {
     const entity = useEntityContext()
     const gltfComponent = useComponent(entity, GLTFComponent)
+    const loaded = GLTFComponent.useSceneLoaded(entity)
+
+    useEffect(() => {
+      setVisibleComponent(entity, loaded)
```
HexaField: this will cause GLTFs invisible/hidden in the studio to become visible upon load

packages/spatial/src/transform/components/EntityTree.tsx
```diff
@@ -494,17 +494,23 @@ export function useChildWithComponents(rootEntity: Entity, components: Component
...
 
-export function useChildrenWithComponents(rootEntity: Entity, components: ComponentType<any>[]): Entity[] {
+export function useChildrenWithComponents(
+  rootEntity: Entity,
+  components: ComponentType<any>[],
+  exclude: ComponentType<any>[] = []
+): Entity[] {
   const children = useHookstate([] as Entity[])
   const componentsString = components.map((component) => component.name).join()
+  const excludeString = exclude.length ? exclude.map((component) => component.name).join() : ''
```
HexaField: `[].join()` evaluates to `''`  so the conditional here is redundant

packages/spatial/src/transform/components/EntityTree.tsx
```diff
@@ -494,17 +494,23 @@ export function useChildWithComponents(rootEntity: Entity, components: Component
...
   useLayoutEffect(() => {
     let unmounted = false
     const ChildSubReactor = (props: { entity: Entity }) => {
       const tree = useOptionalComponent(props.entity, EntityTreeComponent)
       const matchesQuery = _useHasAllComponents(props.entity, components)
+      const matchesExludeQuery = _useHasAllComponents(props.entity, exclude) && exclude.length
 
       useLayoutEffect(() => {
-        if (!matchesQuery) return
+        if (!matchesQuery || matchesExludeQuery) return
```
HexaField, line 514: we need to include `matchesExludeQuery` in the deps of this effect

HexaField (dismissed):

packages/engine/src/gltf/GLTFComponent.tsx
MichaelEstes: Removed for now, functionality will be added back implicitly with the loader refactor

packages/spatial/src/transform/components/EntityTree.tsx
```diff
@@ -494,17 +494,23 @@ export function useChildWithComponents(rootEntity: Entity, components: Component
...
+): Entity[] {
   const children = useHookstate([] as Entity[])
   const componentsString = components.map((component) => component.name).join()
+  const excludeString = exclude.map((component) => component.name).join()
   useLayoutEffect(() => {
     let unmounted = false
     const ChildSubReactor = (props: { entity: Entity }) => {
       const tree = useOptionalComponent(props.entity, EntityTreeComponent)
       const matchesQuery = _useHasAllComponents(props.entity, components)
+      const matchesExludeQuery = _useHasAllComponents(props.entity, exclude) && exclude.length
```
HexaField: i think this would have to be a `_useHasAnyComponents` which would OR them all, not AND, otherwise [Component1, Component2] would only exclude if both were there, not if just one or none was on the entity

HexaField (changes requested):

HexaField (approved):

PR #673, opened 2024-11-15
[IR-5597] Unit Test - Move `webxr-emulator` from @ir-engine/ir-bot into `packages/spatial`

heysokam:
## References
https://tsu.atlassian.net/browse/IR-5597


packages/spatial/tests/webxr/emulator.ts
```diff
@@ -0,0 +1,107 @@
...
+export function getXRSessionData(session: XRSession | null) {
+  if (session === null) return null
+  return session[XRSESSION_SYMBOL]
+}
+
+/**
+ * @description Returns the data of the last session found at the `@param session` passed.
+ * @why Shorthand for getting the last session data in an ergonomic way.
+ * */
+export function getLastXRSessionData(session: XRSession | null) {
```
HexaField, line 55: is this and `getXRSessionData` required? doesn't seem to be used anywhere, we only care about our APIs and the WebXR API, and should avoid getting any data from the emulator directly.

HexaField (changes requested):
only one small question, but please also make a PR to remove the emulator from the ir-bot package too, and we can merge these together to avoid having the code across multiple places

packages/spatial/tests/webxr/emulator.ts
heysokam: It is used in the XR unit tests branch to check that the state of `XRSession` is correctly managed by our functions.
https://github.com/ir-engine/ir-engine/blob/d270a09a840061981d97acdd394f866f73b93e1e/packages/spatial/src/xr/XRSessionFunctions.test.ts#L37

heysokam: @HexaField The `ir-bot` PR is also ready
https://github.com/ir-engine/ir-bot/pull/5

HexaField (approved):
great stuff

PR #675, opened 2024-11-15
Expose name plate parameters to user graphic settings

SYBIOTE:
Introduce user-configurable settings for name plate visibility and trigger distance, enhancing customization options for users.


packages/common/src/constants/UserConstants.ts
```diff
@@ -24,3 +24,4 @@ Infinite Reality Engine. All Rights Reserved.
 */
 
 export const USERNAME_MAX_LENGTH = 70
+export const DEFAULT_USERNAME_TAG_VISIBILITY_RADIUS = 8
```
HexaField: looks like this is unused?

HexaField (changes requested):

packages/common/src/constants/UserConstants.ts
SYBIOTE: yeah, best to do away with it

HexaField (approved):

PR #678, opened 2024-11-16
Test for MountPointComponent.ts 

AmadeusSlabu:
**Test structure**
-
### Describe `MountPointComponent.ts` ( file level ).
- **Describe** `MountPointComponent` ( defining the component ).
  - Name check.
  - JsonID check.
  - Check initial data being set.
  - **Describe** `Reactor` ( functions inside component ).
    - Check to see if the `mountEntity` function was set as a callback to the entity.
    - Check for UI update of Interacteable component if its mounted.
    - Check for `arrow helper component` if debug enabled.
- **Describe** `mountEntity` function
  - Check for early return if no `avatarEntity`.
  - Check for early return if no `mountPointEntity`.
  - Check for early return if `avatarEntity` not seated and mount point occupied by another `avatarEntity2`.
  - Check for early return if `avatarEntity` already seated.
  - Check for `SittingComponent` set on `avatarEntity` and `mountPointEntity` set on `SittingComponent`.
  - Check for movement capture of `AvatarControllerComponent` of `avatarEntity`
  - Check for the `AvatarControllerComponent` `animation state` being set to sitting
  - Check for `MountPointActions` receiving a `mountInteraction` action to set the `mountEntity` and the target mount.
- **Describe** `unmountEntity`.
  - Check for early return if `avatarEntity` has no `SittingComponent`
  - Check for the looping of the seated animation removal.
  - Check for `AvatarControllerComponen` of `avatarEntity` movement capture release.
  - Check for entity unmount.
  - Check so that `avatarEntity` is not dismounted if `forceDismountPosition` is not true and the raycast does not hit anything bellow the `avatarEntity`.
  - Check that `avatarentity` is dismounted if `forceDismountPosition` is true.
  - Check that `avatarEntity` is dismounted if `forceDismountPosition` is false but the raycast hit a collider under avatar.
  - Check that `SittingComponent` is removed from `avatarEntity`.


HexaField (approved):
Amazing work, thank you for the contribution!

PR #687, opened 2024-11-19
Scene settings component.test.ts

AmadeusSlabu:
- Name check
- JsonID check
- Initial data set check
- Proper update on data set change check


HexaField (approved):
Thank you!

PR #697, opened 2024-11-19
ECS Layers & Authoring

dinomut1:
https://tsu.atlassian.net/browse/IR-5411

https://tsu.atlassian.net/browse/IR-5592


packages/spatial/tsconfig.json
```diff
@@ -26,5 +26,5 @@
     "declaration": false,
     "types": ["@types/node"]
   },
-  "include": ["../../__global.d.ts", "./src", "./tests", "./index.ts"]
+  "include": ["../../__global.d.ts", "./src", "./tests", "./index.ts", "../engine/src/scene/components/SimulationLayerTagComponent.ts"]
```
HexaField: vscode being funny

packages/engine/src/avatar/systems/AvatarAnimationSystem.tsx
```diff
@@ -98,9 +99,19 @@ export const AvatarAnimationState = defineState({
...
 })
 
-const avatarAnimationQuery = defineQuery([AnimationComponent, AvatarAnimationComponent, AvatarRigComponent])
-const avatarComponentQuery = defineQuery([AvatarComponent, RigidBodyComponent, AvatarAnimationComponent])
-const avatarRigQuery = defineQuery([AvatarRigComponent])
+const avatarAnimationQuery = defineQuery([
+  AnimationComponent,
+  AvatarAnimationComponent,
+  AvatarRigComponent,
+  SimulationLayerTagComponent
```
HexaField: i thought we were going to be adding this to all queries with an opt out, rather than it be opt in? @speigg

packages/engine/tsconfig.json
```diff
@@ -26,5 +26,5 @@
     "declaration": false,
     "types": ["@types/node"]
   },
-  "include": ["../../__global.d.ts", "./src", "./index.ts"]
+  "include": ["../../__global.d.ts", "./src", "./index.ts", "../ecs/src/SimulationLayerTagComponent.ts"]
```
HexaField: vscode being funny

packages/ecs/src/UUIDComponent.ts
the-reality-engineer, line 1: This looks good

packages/ecs/src/createEntity.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (relationType === 'propagate') {
+      const layerEntity = createEntity(dstLayerID as LayerID)
+      linkedEntityObject[dstLayerID] = layerEntity
+      layerState.linkedEntities[layerEntity].set(linkedEntityObject)
+    }
+  }
+  if (layerID === 'simulation') {
+    setComponent(result, SimulationLayerTagComponent)
+  }
+  if (layerID === 'simulation') {
```
the-reality-engineer: duplicated code?

packages/editor/src/functions/EditorControlFunctions.ts
the-reality-engineer, line 1: Great cleanup here!

the-reality-engineer, line 1: Great cleanup here!

packages/engine/src/avatar/systems/AvatarAnimationSystem.tsx
the-reality-engineer: yes, our queries should probably include `SimulationLayerTagComponent` automatically, with a `simulationLayerOnly` parameter defaulted to true. 

Let's think about this:
- It's clear that there are systems/reactors that should ONLY run in the simulation layer (e.g. physics, game logic, rendering)
- It's clear that there are systems/reactors that should run in BOTH layers (e.g., asset loading, transforms)
- I can't think of a use-case for systems/reactors that should only ONLY run in the authoring layer. 

That said, it seems that the semantics we want to offer are:
A) query entities in all layers
B) query entities in simulation layer

If the default is all layers (A), then the danger is that users may accidentally create systems/reactors that run in the authoring layer without realizing it — is this actually a problem? It could mean that their "game logic", for example, runs in authoring layer by defualt, which I'd say is problematic. 

If the default is the simulation layer (B), I can't think of any downsides, aside from having to go find the systems/reactors we want to run in both layers, and updating those queries to include the authoring layer entities (either by excluding the simulation tag, or by adding an authoring tag). 

So, while we want to support both (A) and (B) semantics, I think our cleanest api design is to make (B) the default, since it appears to represent the pit of success. If so, our queries should indeed include the simulation tag by default, with an option for (A) semantics. Or as Josh calls it, an option to opt-out of (B) semantics.

HexaField: The problem would be that component reactors currently define helpers, so we'd potentially be duplicating those across both layers. Maybe we restrict those to just the authoring layer.

I also think this will be simpler once we get rid of component reactors altogether - at that point we will have systems with lots and lots of QueryReactors, which could have a separate argument for layer that defaults to simulation.

My expectation is the majority of cases will be simulation only, with some being both (transforms, asset loading) and a few being authoring only (helpers).

the-reality-engineer: No, helpers exist entirely within, are rendered by and interacted with from the simulation layer. They do not exist in the authoring layer at all. Only the scene content exists in the authoring layer.

dinomut1:
@speigg @HexaField Yep I agree we need to put SimulationLayerTagComponent under the hood in querying. For now just plowing through for a working prototype

packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -1227,19 +1303,307 @@ const _createCubicSplineTrackInterpolant = (track: KeyframeTrack) => {
...
+
+    if (node.mesh !== undefined) {
+      if (!document.meshes) throw new Error('GLTFLoader: Referencing undefined mesh array')
+
+      if (!hasComponent(nodeEntity, ColliderComponent)) setComponent(nodeEntity, VisibleComponent)
+
+      const meshDef = document.meshes[node.mesh]
+
+      const isSinglePrimitive = meshDef.primitives.length === 1
+      const meshGeometry = await GLTFLoaderFunctions.useLoadPrimitives(options, node.mesh)
```
HexaField: we should do this in parallel

HexaField (changes requested):

packages/engine/src/avatar/components/AnimationComponent.test.tsx
```diff
@@ -103,21 +97,19 @@ describe('AnimationComponent', () => {
...
       createEngine()
+      startEngineReactor()
     })
 
     afterEach(() => {
       return destroyEngine()
     })
 
-    it('should bind animation tracks to entities based on node id sourced from entity UUIDs', async () => {
+    it.only('should bind animation tracks to entities based on node id sourced from entity UUIDs', async () => {
```
heysokam: Just noticed the `.only` here. Noting in case it was left by accident.

HexaField: thanks yeah i reverted it

HexaField (approved):
Oh boy, here we go...

AidanCaruso (approved):
Obligatory approval!

DanielBelmes (commented):
Approved

packages/client-core/src/hooks/useRender3DPanelSystem.tsx
```diff
@@ -78,6 +80,8 @@ export function useRender3DPanelSystem(canvas: React.MutableRefObject<HTMLCanvas
     return () => {
       // cleanup entities and state associated with this 3d panel
       removeEntity(cameraEntity)
+      /**@todo why is this necessary to prevent gltf system stopping? */
```
HexaField, line 83: this is probably fixed now since we are using a suspense boundary and delayed entity recycling

PR #698, opened 2024-11-19
added the "my assets" category to the assets panel, this category sho…

MbfloydIR:
…ws the user's uploaded assets

updated the sort to use the name property, and updated the static resource hooks to update the name property hen the record is being created, updated, or patched

## Summary
https://tsu.atlassian.net/browse/IR-5611
## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/media/static-resource/static-resource.hooks.ts
+++ b/packages/server-core/src/media/static-resource/static-resource.hooks.ts
@@ -103,6 +103,20 @@ const createHashIfNeeded = async (context: HookContext<StaticResourceService>) =
   }
 }
 
+const updateName = async (context: HookContext<StaticResourceService>) => {
+  if (!context.data || !(context.method === 'create' || context.method === 'update' || context.method === 'patch')) {
+    throw new BadRequest(`${context.path} service only works for data in ${context.method}`)
+  }
+
+  if (Array.isArray(context.data)) throw new BadRequest('Batch create is not supported')
+
+  const data = context.data
+  if (!data.key) return
+
+  const [_, directory, file] = /(.*)\/([^\\\/]+$)/.exec(data.key)!
+  context.data.name = file
+}
+
 const updateResourcesJson = async (context: HookContext<StaticResourceService>) => {
   if (!context.method || !(context.method === 'create' || context.method === 'update' || context.method === 'patch'))
     throw new BadRequest('[updateResourcesJson] Only create, update, patch methods are supported')
@@ -362,7 +376,8 @@ export default {
       // schemaHooks.validateData(staticResourceDataValidator),
       discardQuery('projectId'),
       schemaHooks.resolveData(staticResourceDataResolver),
-      createHashIfNeeded
+      createHashIfNeeded,
+      updateName
     ],
     update: [
       ensureProject,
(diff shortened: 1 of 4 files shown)


packages/server-core/src/media/static-resource/static-resource.hooks.ts
```diff
@@ -103,6 +103,20 @@ const createHashIfNeeded = async (context: HookContext<StaticResourceService>) =
...
+    throw new BadRequest(`${context.path} service only works for data in ${context.method}`)
+  }
+
+  if (Array.isArray(context.data)) throw new BadRequest('Batch create is not supported')
+
+  const data = context.data
+  if (!data.key) return
+
+  const [_, directory, file] = /(.*)\/([^\\\/]+$)/.exec(data.key)!
+  context.data.name = file
```
HexaField: we need to add some more logic here

- always set the name on create method
- if update or patch method, and a static resource entry already exists, check to see if the name matches the name part of  key
  - update the name derived from the key, otherwise do not update the name (this will preserve custom asset names)

HexaField (changes requested):

packages/server-core/src/media/static-resource/static-resource.hooks.ts
MbfloydIR: implemented

packages/server-core/src/media/static-resource/static-resource.hooks.ts
```diff
@@ -103,6 +103,39 @@ const createHashIfNeeded = async (context: HookContext<StaticResourceService>) =
...
+    throw new BadRequest(`${context.path} service only works for data in ${context.method}`)
+  }
+
+  if (Array.isArray(context.data)) throw new BadRequest('Batch create is not supported')
+
+  const id = context.id
+  const data = context.data
+  if (!data.key || !id) return
+
+  const [_, directory, file] = /(.*)\/([^\\\/]+$)/.exec(data.key)!
```
HexaField: this regex can be moved inside the `doUpdate` condition, since it's only used there

packages/server-core/src/media/static-resource/static-resource.hooks.ts
```diff
@@ -103,6 +103,39 @@ const createHashIfNeeded = async (context: HookContext<StaticResourceService>) =
...
+  let doUpdate = false
```
HexaField: let's call this `shouldUpdateName` for clarity

HexaField (changes requested):

packages/server-core/src/media/static-resource/static-resource.hooks.ts
MbfloydIR: implemented

packages/server-core/src/media/static-resource/static-resource.hooks.ts
MbfloydIR: implemented

HexaField (changes requested):
great, i think the only thing left is to write a few unit tests for the various situations for the static resource backend hook

MbfloydIR:
> great, i think the only thing left is to write a few unit tests for the various situations for the static resource backend hook

implemented

HexaField (approved):
test looks great!

PR #700, opened 2024-11-19
Fix <Input /> regressions

CITIZENDOT:
## Summary
Fixes regressions caused by recent Input component update
- Mostly changes are setting fullWidth where it is needed.

## References
closes #_insert number here_

### DIFF
--- a/packages/ui/src/primitives/tailwind/Input/index.tsx
+++ b/packages/ui/src/primitives/tailwind/Input/index.tsx
@@ -23,7 +23,7 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
 Infinite Reality Engine. All Rights Reserved.
 */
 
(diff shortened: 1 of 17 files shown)


packages/ui/src/primitives/tailwind/Input/index.tsx
```diff
@@ -110,6 +110,12 @@ const Input = (
     }
   }, [labelProps])
 
+  useEffect(() => {
+    if (labelProps?.text === 'Site Url') {
+      console.log('fullWidth: ', fullWidth, props)
+    }
+  }, [labelProps])
```
aditya-mitra: yes, we can have it removed

CITIZENDOT: removed

aditya-mitra (approved):

PR #703, opened 2024-11-20
IR-4998 collider component overhaul

SamMazerIR:
## Summary
collider component is overhauled. It now supports the use case of a single collider component at or below the level of a rigidbody, and will automatically assign colliders to itself and it's children (ignores any child meshComponent entities that also have a collider on it or between)

Also overhauled our calculation of collider shapes and placement. Notably the center position will respect the mesh if present, meaning that meshes with origin points that are not the mesh center will fit snugly inside the assigned collider types (mesh collider not applicable)

## References
[https://tsu.atlassian.net/browse/IR-4998](https://tsu.atlassian.net/browse/IR-4998)


packages/spatial/src/physics/components/ColliderComponent.tsx
```diff
@@ -23,20 +23,43 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
 import { TransformComponent } from '../../transform/components/TransformComponent'
+import { computeTransformMatrix } from '../../transform/systems/TransformSystem.ts'
 import { Physics } from '../classes/Physics'
 import { CollisionGroups, DefaultCollisionMask } from '../enums/CollisionGroups'
-import { ShapeSchema, Shapes } from '../types/PhysicsTypes'
+import { Shapes, ShapeSchema } from '../types/PhysicsTypes'
 import { RigidBodyComponent } from './RigidBodyComponent'
 import { TriggerComponent } from './TriggerComponent'
 
+export const NestedCollidersState = defineState({
```
HexaField: a separate state like this is a bad idea. the collider component should contain all the data relevant for itself. this can be moved to the collider schema.

packages/spatial/src/physics/classes/Physics.ts
```diff
@@ -302,11 +307,22 @@ const setRigidBodyType = (world: PhysicsWorld, entity: Entity, type: Body) => {
...
   /** @todo turns out this is a react/rapier bug when it comes to changing the rigidbody type. This is the best workaround I could find*/
   rigidbody.setEnabled(false)
   rigidbody.setEnabled(true)
+
+  const childColliderEntities = getChildrenWithComponents(entity, [ColliderComponent])
+  const colliderEntities = hasComponent(entity, ColliderComponent)
+    ? [entity, ...childColliderEntities]
+    : childColliderEntities
+
+  const nestedCollidersState = getMutableState(NestedCollidersState)
```
HexaField: it makes me nervous to rely on state like this in the physics abstraction. This means that the physics API and the component itself is handling data that feeds back on itself. we need a uni-directional flow of responsibilities here.

The Physics API exists purely to translate component data into the physics world, not to manipulate the state of the ECS.

packages/spatial/src/physics/components/ColliderComponent.tsx
```diff
@@ -60,22 +83,104 @@ export const ColliderComponent = defineComponent({
     const physicsWorld = Physics.useWorld(entity)
     const triggerComponent = useOptionalComponent(entity, TriggerComponent)
     const hasCollider = useState(false)
+    const uuid = useComponent(entity, UUIDComponent)
+    const childMeshEntities = useChildrenWithComponents(entity, [MeshComponent])
+
+    const nestedCollidersState = getMutableState(NestedCollidersState)
```
HexaField: i think we need separating the user story here from the technical implementation.

Does the ColliderComponent specify a single collider, or potentially a hierarchy of colliders? I think it should be the former, and we can have a higher-level collider that handles which colliders exist/are enabled.

packages/spatial/src/transform/components/TransformComponent.ts
```diff
@@ -130,7 +130,21 @@ export const TransformComponent = defineComponent({
   getMatrixRelativeToEntity: (entity: Entity, relativeEntity: Entity, outMatrix: Matrix4) => {
     const transform = getComponent(entity, TransformComponent)
     const relativeTransform = getComponent(relativeEntity, TransformComponent)
-    return outMatrix.copy(relativeTransform.matrixWorld).invert().multiply(transform.matrixWorld)
+
+    if (!transform || !relativeTransform) {
```
HexaField: because this is a hot code path, we should assume that transforms exist on both entities and handle this wherever we are unsure that they arent. otherwise a few extra ifs and a try catch could cause performance problems in the transform and physics systems.

we also want to return the outMatrix as before, such that we can use it in-line in various places

packages/ui/src/components/editor/properties/collider/index.tsx
```diff
@@ -48,13 +51,18 @@ export const ColliderComponentEditor: EditorComponentType = (props) => {
...
+  const childMeshEntities = useChildrenWithComponents(props.entity, [MeshComponent])
+
   return (
     <NodeEditor
       {...props}
       name={t('editor:properties.collider.name')}
       description={t('editor:properties.collider.description')}
       Icon={ColliderComponentEditor.iconComponent}
     >
+      {childMeshEntities.length == 0 && ( //<ErrorPopup className="ml-5 text-red-400" visibleDuration={10000}  >{ t('editor:properties.collider.warn-missing-mesh') }  </ErrorPopup>
```
HexaField: i think this limitation will break expectations. we dont need a mesh (that is then set to invisible) just to add a collider - we should be able to just add a collider to an entity.

HexaField (changes requested):

packages/spatial/src/physics/components/ColliderComponent.tsx
SamMazerIR: I could see a new "collider manager" component working nicely, I personally feel like there was more conceptual iteration needed prior to starting this feature work...which I tried to kickstart

packages/ui/src/components/editor/properties/collider/index.tsx
SamMazerIR: this is a good point, maybe we only show this in mesh mode. This discussion does however make me feel like we should include an option in the node editor to attempt to base the collider off of a mesh or not

packages/spatial/src/transform/components/TransformComponent.ts
SamMazerIR: sounds good, i'll revert this change

packages/spatial/src/physics/components/ColliderComponent.tsx
```diff
@@ -51,6 +52,7 @@ export const ColliderComponent = defineComponent({
     collisionMask: S.Number(DefaultCollisionMask),
 
     //shape specific parameters
+    matchMesh: S.Bool(false),
```
HexaField: is this option necessary? we should try to keep the collider component adhering the to proposed GLTF spec

HexaField (dismissed):
can we please move all the collider functionality to a hook in a separate file, so we can reason about the GLTFComponent's core functionality separately from this new colldier logic?

packages/spatial/src/physics/components/ColliderComponent.tsx
SamMazerIR: this is being used to drive the collider based on mesh data rather than manual input, and disables the related fields in the node editor. It is used by the GLTFComponent for applying colliders

packages/spatial/src/physics/components/ColliderComponent.tsx
SamMazerIR: removed

heysokam:
Remember to add tests for all the new logic.
These files were already fully covered, so this script should help you find what your code is missing:
https://github.com/ir-engine/ir-engine/blob/689d61f0430ac6762f4e90dc81bbf53c0bf5b237/packages/spatial/package.json#L12

packages/spatial/src/physics/components/ColliderComponent.tsx
```diff
@@ -75,7 +88,25 @@ export const ColliderComponent = defineComponent({
...
+      transform.scale,
+      component.centerOffset,
+      component.boxSize,
+      component.radius,
+      component.height,
+      meshComponent
+    ])
+
+    useEffect(() => {
+      if (!physicsWorld) return
```
HexaField: this should be inside the return function

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -248,13 +256,23 @@ const ResourceReactor = (props: { documentID: string; entity: Entity }) => {
...
   useEffect(() => {
     if (getComponent(props.entity, GLTFComponent).progress === 100) return
     if (!getState(GLTFDocumentState)[props.documentID]) return
     const entities = resourceQuery.filter((e) => getComponent(e, SourceComponent) === props.documentID)
     if (!entities.length) {
-      if (dependenciesLoaded) getMutableComponent(props.entity, GLTFComponent).progress.set(100)
+      if (dependenciesLoaded) {
+        getMutableComponent(props.entity, GLTFComponent).progress.set(100)
+
+        //@todo add the collider state update here
```
HexaField: is this still todo?

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -248,13 +256,23 @@ const ResourceReactor = (props: { documentID: string; entity: Entity }) => {
   const resourceQuery = useQuery([SourceComponent, ResourcePendingComponent])
   const gltfDocumentState = useMutableState(GLTFDocumentState)
   const sourceEntities = useHookstate(SourceComponent.entitiesBySourceState[props.documentID])
+  const childMeshEntities = useChildrenWithComponents(props.entity, [MeshComponent])
```
HexaField: are all these hooks still needed?

packages/engine/src/gltf/GLTFComponent.tsx
SamMazerIR: removing

packages/spatial/src/physics/components/ColliderComponent.tsx
SamMazerIR: fixing

packages/engine/src/gltf/GLTFComponent.tsx
SamMazerIR: nope, missed removing them when moving to external file

HexaField (changes requested):
sky station has broken :(

HexaField (dismissed):
As we discussed, the use of BOTH a transform at the collider description level and the collider instance transform is causing problems.

SamMazerIR:
fixed sky station, opted for vertical default orientation of capsules/cylinders like all other engines as the alternative requires us "trying to know best" and add counter rotation to compensate (capsules/cylinders always get created vertically)

SamMazerIR:
> As we discussed, the use of BOTH a transform at the collider description level and the collider instance transform is causing problems.

issue surrounded when we did rotations etc as. you saw yesterday, this is fixed

the-reality-engineer (commented):
I like the matchMesh implementation. It looks useful.

DanielBelmes (approved):

PR #705, opened 2024-11-20
ECS animation tracks

AidanCaruso:
## Summary
The ticket [IR-3595](https://tsu.atlassian.net/browse/IR-3595) has been repeatedly reopened as the threejs name based animation binding was failing on different kinds of gltfs after we refactored the loader. Rather than fixing by performing the threejs loader's destructive renaming operation on our nodes to sanitize them of delimiting characters, this refactor avoids that issue by overriding threejs functions to use either the VRM schema directly for rigged root entities, or ECS UUIDs to find nodes, and then choosing child entities of a root node by their node ID, to ensure reusibility of the same loaded clips between identical GLTF hierarchies.

Despite changing to use ECS for node binding, we still uses threejs' object node animation mixer. This requires an ECS-sourced reference to the threejs object node for each entity, typically accomplished by getting the relevant object node component (BoneComponent, Object3DComponent etc). In the case of VRM rigs, we bind animations to the "normalized" rig, hence the addition of the NormalizedBoneComponent on each bone entity of a rig. This allows the normalized bones to be animated directly through the ecs hierarchy without need for creating a second entity rig hierarchy.

## References
closes https://tsu.atlassian.net/browse/IR-3595 once and for all

## QA Steps

[IR-3595]: https://tsu.atlassian.net/browse/IR-3595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/engine/src/avatar/components/NormalizedBoneComponent.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { defineComponent, S } from '@ir-engine/ecs'
+import { Bone } from 'three'
+
+export const NormalizedBoneComponent = defineComponent({
+  name: 'NormalizedBoneComponent',
+
+  schema: S.Required(S.Type<Bone>()),
+
+  onSet: (entity, component, mesh: Bone) => {
```
HexaField: i think this is redundant with schemas?

HexaField (changes requested):

packages/engine/src/avatar/components/NormalizedBoneComponent.ts
AidanCaruso: Yes that's right, left this in by mistake when copy/pasting BoneComponent.

packages/engine/src/avatar/components/AnimationComponent.ts
```diff
@@ -97,85 +99,54 @@ export const useLoadAnimationFromGLTF = (url: string, keepEntity = false) => {
...
+  return results
+}
 
-      return null
-    }
+export const getTrackId = (entity: Entity) =>
+  getComponent(entity, UUIDComponent).replace(getComponent(entity, SourceComponent) + '-', '')
 
-    const subTreeNode = searchEntitySubtree(children)
+PropertyBinding.findNode = (root: Object3D, nodeName) => {
```
HexaField: please add string type to `nodeName`

packages/engine/src/avatar/systems/AvatarAnimationSystem.tsx
```diff
@@ -338,9 +348,23 @@ const Reactor = () => {
...
+/**
+ * @todo replace this with a retargeting utility to retarget the source animation assets rather than every time on load,
+ * and introduce a loader function that only loads the necessary data to avoid cleanup of the ecs armature
+ */
+export const setupMixamoAnimation = (entity: Entity) => {
+  normalizeAnimationClips(entity)
+  setComponent(entity, AvatarRigComponent)
+  traverseEntityNode(entity, (child) => {
+    const name = getComponent(child, NameComponent).replace(':', '')
+    if (mixamoVRMRigMap[name]) AvatarRigComponent.setBone(entity, child, mixamoVRMRigMap[name])
```
HexaField, line 360: ah, i see this is where we are using setBone. is there a reason why we have one place setting this data internally and one externally?

packages/engine/src/mocap/MotionCaptureSystem.ts
```diff
@@ -130,8 +131,8 @@ const execute = () => {
   for (const entity of motionCaptureQuery()) {
     const peers = Object.keys(network.peers).find((peerID: PeerID) => timeSeriesMocapData.has(peerID))
     const rigComponent = getComponent(entity, AvatarRigComponent)
-    if (!rigComponent.normalizedRig) continue
-    const worldHipsParent = rigComponent.normalizedRig.hips.node.parent
+    if (!rigComponent.bonesToEntities) continue
```
HexaField: this will always be truthy, maybe we should be checking for `rigComponent.bonesToEntities.hips` instead?

packages/engine/src/mocap/poseToInput.ts
```diff
@@ -41,7 +42,7 @@ const minSeatedAngle = 1.25, //radians
 let poseHoldTimer = 0
 
 export const evaluatePose = (entity: Entity) => {
-  const rig = getComponent(entity, AvatarRigComponent).normalizedRig
+  const rig = getComponent(entity, AvatarRigComponent).bonesToEntities
```
HexaField, line 45: rig will always be truthy here too

HexaField (changes requested):
we have a few references to the VRM bones i think we should refactor to use the new entity <=> bone maps anywhere we are accessing `vrm.humanoid`

packages/engine/src/avatar/systems/AvatarAnimationSystem.tsx
AidanCaruso: Typically the vrm bone map has its values set on VRM creation, which is the only time it happens for any actual avatar asset. 

The external function here handles imported animations, which are a special case. We have this avatar rig that needs to be mapped to VRM for retargeting, but there's no reason to create a full VRM object from the rig. To prevent the extra logic being run, I opted to break out the only necessary processing that needs to happen into a separate function called any time an animation file is loaded.

If we wanted to only call setBone logic from one place by breaking it out entirely to its own loop after calling the VRM creation functions, that would require an extra traversal of the rig's entities outside of the vrm creation function, which already does this once.

packages/engine/src/mocap/MotionCaptureSystem.ts
AidanCaruso: Yea good point, missed these

HexaField (approved):

the-reality-engineer:
Exciting!

PR #708, opened 2024-11-20
IR-5369 A variable with a NonSerialized schema will still load values from the scene if it was previously saved

MichaelEstes:
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/gltf/GLTFState.tsx
+++ b/packages/engine/src/gltf/GLTFState.tsx
@@ -73,7 +73,6 @@ import {
   State,
   Topic,
   useHookstate,
-  useImmediateEffect,
   useMutableState
 } from '@ir-engine/hyperflux'
 import { CameraComponent } from '@ir-engine/spatial/src/camera/components/CameraComponent'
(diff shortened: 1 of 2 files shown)


packages/engine/src/gltf/GLTFState.tsx
```diff
@@ -659,7 +658,10 @@ const NodeReactor = (props: { nodeIndex: number; childIndex: number; parentUUID:
       for (const extension in node.extensions.value) {
         const Component = ComponentJSONIDMap.get(extension)
         if (!Component) continue
-        setComponent(entity, Component, node.extensions[extension].get(NO_PROXY_STEALTH))
+
+        const serialized = Component.toJSON(node.extensions[extension].get(NO_PROXY_STEALTH))
```
HexaField: is this going to work for things like colors and anything with a function call?

I think we should lean on explicit schema validation. Maybe that can be implemented in toJSON instead?

HexaField (changes requested):

packages/engine/src/gltf/GLTFState.tsx
MichaelEstes: Shouldn't this extension data only be what is serializable in the json? (No classes or functions)

toJSON is currently a wrapper for SerializeSchema which is explicit schema validation, I'd prefer the schema functions be exposed in as few places as possible, but we could change this to use the validation function directly.

HexaField (approved):

PR #709, opened 2024-11-20
Fix screenshareTarget duplicate, Fix FollowCameraComponent warning

DanielBelmes:
## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/camera/components/FollowCameraComponent.ts
+++ b/packages/spatial/src/camera/components/FollowCameraComponent.ts
@@ -213,7 +213,7 @@ const computeCameraFollow = (cameraEntity: Entity, referenceEntity: Entity) => {
   const follow = getComponent(cameraEntity, FollowCameraComponent)
   const followState = getMutableComponent(cameraEntity, FollowCameraComponent)
   const cameraTransform = getComponent(cameraEntity, TransformComponent)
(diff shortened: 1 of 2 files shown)


packages/spatial/src/camera/components/FollowCameraComponent.ts
```diff
@@ -213,7 +213,7 @@ const computeCameraFollow = (cameraEntity: Entity, referenceEntity: Entity) => {
   const follow = getComponent(cameraEntity, FollowCameraComponent)
   const followState = getMutableComponent(cameraEntity, FollowCameraComponent)
   const cameraTransform = getComponent(cameraEntity, TransformComponent)
-  const targetTransform = getComponent(referenceEntity, TransformComponent)
+  const targetTransform = getOptionalComponent(referenceEntity, TransformComponent)
```
HexaField, line 216: what was this breaking for?

HexaField (approved):

packages/spatial/src/camera/components/FollowCameraComponent.ts
DanielBelmes: 1000s of warning messages in FollowCameraComponent from the Transform being computed. Not sure where it came from. Because this code hasn't changed.

PR #718, opened 2024-11-21
Revert site manifest changes I made

DanielBelmes:
## Summary
Awhile ago I made these changes and there's been an error ever since. Turns out I didn't understand we have a vite plugin that configures the manifest file for us so I don't need to include it with ejs.

I probably didn't need to include it in client settings as well.

## References
closes #_insert number here_


DanielBelmes:
I would like to remove the data from client settings as well

HexaField:
@DanielBelmes is this ready?

DanielBelmes:
@HexaField I think so. it's been so long.

DanielBelmes:
There is still refrences to siteManifest within client-setting and a migration I wrote. But I removed it from vite config and the index at least. Maybe I gotta look into it one more time to make sure.

PR #719, opened 2024-11-21
IR-4606 IR-4625 Schema validate function implementation, component schema cleanup

MichaelEstes:
## References
closes #_insert number here_


HexaField (commented):
Looks good, just noting that we spoke and decided to move the threejs schema helpers to spatial package

HexaField (changes requested):
seems like spectator mode is being force, cant see any other regressions yet

HexaField (approved):

the-reality-engineer:
I'm concerned about the growing number of single-letter namespaces here... why `T`?? 
:)

HexaField:
> I'm concerned about the growing number of single-letter namespaces here... why `T`?? :)

Yeah me too, I'm not sure what we should do about it. Some concepts simply do not belong in the ecs package (like threejs and other spatial concepts)

the-reality-engineer:
Also, why are we moving a lot of core schemas out of the ECS package? Entity and Color, for example.

the-reality-engineer:
Seems arbitrary that Booleans are considered worthy to be in the ecs package, but not Vector3 or Entity or Color

the-reality-engineer:
Can we just put all of our schema primitives in the common package instead of splitting them up across packages like this?

HexaField:
The common package is common for the server and client,not the engine. Any engine packages cannot import the common package. They need to be in the packages with the appropriate concepts (entities in ecs, color/vector3/quaternion in spatial etc...)

the-reality-engineer:
@HexaField The `editor`, `ecs`, and `spatial` packages all contain imports from the `common` package.

HexaField:
> @HexaField The `editor`, `ecs`, and `spatial` packages all contain imports from the `common` package. 

And we are trying to get rid of that. Editor is studio UI, ecs has one unused import of common and I don't know about spatial.

PR #729, opened 2024-11-23
Ensure thumbnail generation and loading screen only in editor

SYBIOTE:
Limit thumbnail and loading screen functionalities to the editor environment, preventing unnecessary operations when not in use.


packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
```diff
@@ -97,7 +98,12 @@ export default function AddEditLocationModal(props: {
   const screenSharingEnabled = useHookstate<boolean>(location?.locationSetting.screenSharingEnabled || true)
   const locationType = useHookstate(location?.locationSetting.locationType || 'public')
   const sceneThumbnailState = useHookstate(getMutableState(SceneThumbnailState))
-  const sceneSettingsEntities = useQuery([SceneSettingsComponent])
+  // only use query if engine is running
+  let sceneSettingsEntities: Entity[] = []
+  if (getState(EngineState).viewerEntity !== UndefinedEntity) {
+    // no viewerEntity if not in Editor, running the engine
+    sceneSettingsEntities = useQuery([SceneSettingsComponent])
```
HexaField: We should put the conditional logic after this, react requires the same number of hooks across renders

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
```diff
@@ -97,7 +98,12 @@ export default function AddEditLocationModal(props: {
   const screenSharingEnabled = useHookstate<boolean>(location?.locationSetting.screenSharingEnabled || true)
   const locationType = useHookstate(location?.locationSetting.locationType || 'public')
   const sceneThumbnailState = useHookstate(getMutableState(SceneThumbnailState))
-  const sceneSettingsEntities = useQuery([SceneSettingsComponent])
+  // only use query if engine is running
+  let sceneSettingsEntities: Entity[] = []
+  if (getState(EngineState).viewerEntity !== UndefinedEntity) {
```
HexaField: This needs to be reactive

HexaField (changes requested):

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
```diff
@@ -98,12 +98,17 @@ export default function AddEditLocationModal(props: {
...
+
   // only use query if engine is running
-  let sceneSettingsEntities: Entity[] = []
-  if (getState(EngineState).viewerEntity !== UndefinedEntity) {
-    // no viewerEntity if not in Editor, running the engine
-    sceneSettingsEntities = useQuery([SceneSettingsComponent])
-  }
+  useEffect(() => {
+    if (engineState.viewerEntity.value === UndefinedEntity) return
+    sceneSettingsEntities.set(useQuery([SceneSettingsComponent]))
```
HexaField: not quite what i meant. it is as simple as

```ts
const sceneSettingsEntities = useQuery([SceneSettingsComponent])
const sceneSettingsEntity = engineState.viewerEntity.value ? sceneSettingsEntities[0] : UndefinedEntity
```

And I noticed it also looks like `SceneThumbnailState.uploadThumbnail` only takes one entity, not an array of entities, hence the change above. Please add an argument type to `SceneThumbnailState.uploadThumbnail`

HexaField (changes requested):

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
SYBIOTE: const sceneSettingsEntities = useQuery([SceneSettingsComponent])
this line causes issues in the multitenancy page, we just cant use the query at all , as we dont have an engine running the background , when on the MT dashboard

HexaField: in that case, we should not have this logic here at all. the multitenancy page must not even import any of the engine packages to ensure it loads fast. we need to figure out a different solution

HexaField (changes requested):

HexaField:
@aditya-mitra @CITIZENDOT we may need two copies of this modal in this case unfortunately

SYBIOTE:
the efficiency freak in me cries, i understand the need for separation, but the modals to exactly the same thing

HexaField:
Maybe we can lazily load a part of the UI based on whether we're in the studio or not

SYBIOTE:
> Maybe we can lazily load a part of the UI based on whether we're in the studio or not

i intend to do just that, i was using a simple conditional check, ill look into lazy loading

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
HexaField: this needs to be fixed. useQuery in this context will throw an error

SYBIOTE:
after looking it up, looks like my current implementation is already utilizing lazy load, although im using a hookstate to implement it, should i transfer it to react state? i believe the best practice is to use hookstate over react state?
@HexaField

SYBIOTE:
using a react state to implement the lazy loading 
@HexaField

SYBIOTE:
implemented

HexaField:
No, lazy loading uses dynamic imports with `const Component = React.lazy(() => import('./whatever'))`. The engine code should be lazily loaded, such that it is only imported when it is needed.

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
```diff
@@ -317,56 +318,11 @@ export default function AddEditLocationModal(props: {
...
-                    SceneThumbnailState.uploadLoadingScreen(sceneSettingsEntities)
-                  }}
-                  disabled={!sceneThumbnailState.loadingScreenImageData.value}
-                  className="w-full"
-                >
-                  {t('editor:properties.sceneSettings.save')}
-                </Button>
-              </div>
-            </div>
+            {engineState.viewerEntity.value !== UndefinedEntity && (
```
HexaField: rather than using engineState here, we should pass in a boolean argument to the react component that specifies whether or not we should render this. we want to avoid using any ecs/spatial/engine imports at all

HexaField (changes requested):

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
SYBIOTE: gotcha

HexaField (approved):

PR #730, opened 2024-11-23
IR-5433 Scene complexity grows infinitely as you repeatedly enter and exit preview mode

MichaelEstes:
## Summary
Use object id a resource key, update resource tracking on GLTF source change

## References
closes #_insert number here_


HexaField (commented):
Still getting textures leaking when using the dev test suite with the `scene_resource_tracking` example

- add two textures, then remove both, texture count has increased by 1
- add flight helmet, remove, add, remove - shadows/lighting will be messed up


Edit: looks like each type of threejs resource has it's own id counter (Node, BufferGeometry, Object3D, UniformGroup, Material, Source, Texture)

HexaField (approved):

PR #745, opened 2024-11-27
Revive Networked Scene Entities and Change Network ID Rules

HexaField:
## References
closes #_insert number here_


packages/engine/src/grabbable/GrabbableComponent.ts
```diff
@@ -0,0 +1,170 @@
...
+        entityUUID: getComponent(grabbableEntity, UUIDComponent),
+        grabberEntityUUID: getComponent(grabberEntity, UUIDComponent),
+        grabbed: false
+      })
+    )
+  }
+})
+
+const grabCallback = (grabbableEntity: Entity) => {
+  const nonCapturedInputSources = InputSourceComponent.nonCapturedInputSources()
```
the-reality-engineer, line 118: Grabbing all non-captured input sources here probably isn't the behavior we need. I think we want to know which input source triggered the grab callback, and use that to determine which hand the grabbable should be attached to. We can pass the input source down from the InteractComponent's callback.

the-reality-engineer (changes requested):

packages/engine/src/grabbable/GrabbableComponent.ts
HexaField: that information does not come from InteractableComponent currently. how do you suggest we do this? might be out of scope for the context of this PR

the-reality-engineer (approved):

PR #753, opened 2024-12-03
Prevent avatars reloading on change

AidanCaruso:
## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/src/networking/AvatarSpawnSystem.tsx
+++ b/packages/client-core/src/networking/AvatarSpawnSystem.tsx
@@ -127,7 +127,7 @@ export const AvatarSpawnReactor = (props: { sceneEntity: Entity }) => {
         dispatchAction(WorldNetworkAction.destroyEntity({ entityUUID: getComponent(selfAvatarEntity, UUIDComponent) }))
       }
     }
-  }, [spawnAvatar.value, userAvatar, !!userAvatarQuery.data.length])
+  }, [spawnAvatar.value, userAvatar])
 
   const selfAvatarEntity = AvatarComponent.useSelfAvatarEntity()
   const errorWithAvatar = !!useOptionalComponent(selfAvatarEntity, ErrorComponent)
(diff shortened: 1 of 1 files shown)


packages/client-core/src/networking/AvatarSpawnSystem.tsx
```diff
@@ -127,7 +127,7 @@ export const AvatarSpawnReactor = (props: { sceneEntity: Entity }) => {
         dispatchAction(WorldNetworkAction.destroyEntity({ entityUUID: getComponent(selfAvatarEntity, UUIDComponent) }))
       }
     }
-  }, [spawnAvatar.value, !!userAvatar])
+  }, [spawnAvatar.value, userAvatar, !!userAvatarQuery.data.length])
```
HexaField: the deps here should just be `[spawnAvatar.value, !!userAvatar.data[0]]`

HexaField (changes requested):

packages/client-core/src/networking/AvatarSpawnSystem.tsx
```diff
@@ -127,7 +127,7 @@ export const AvatarSpawnReactor = (props: { sceneEntity: Entity }) => {
         dispatchAction(WorldNetworkAction.destroyEntity({ entityUUID: getComponent(selfAvatarEntity, UUIDComponent) }))
       }
     }
-  }, [spawnAvatar.value, !!userAvatar])
+  }, [spawnAvatar.value, userAvatar])
```
HexaField: needs to be boolean cast `!!userAvatar`

HexaField (changes requested):

packages/client-core/src/networking/AvatarSpawnSystem.tsx
AidanCaruso: my bad, multitasking

HexaField (approved):

PR #757, opened 2024-12-03
UI improvements

CITIZENDOT:
## Summary
- Improve Tooltip component
  - Added arrow (Previously, arrow wasn't visible/working)
  - Removed dependency on `reactjs-popup` (Now handling the logic natively)
- Moved Tooltip & Select components (and their imports) to UI root
- **Input:** Renamed `variantSize` to `size`
- **Select:** Renamed `inputVariantSize` to `inputSize`

## References
closes #_insert number here_


packages/ui/src/primitives/tailwind/Select/index.tsx
```diff
@@ -45,7 +45,7 @@ export interface OptionType {
 export interface SelectProps<T = string | number> {
   options: OptionType[]
   width?: 'sm' | 'md' | 'lg' | 'full'
-  inputSizeVariant?: InputProps['variantSize']
+  inputSize?: InputProps['size']
```
aditya-mitra: `inputSize` is just for the height right?
how about we rename it to `height`?

aditya-mitra (approved):
awesome
works good locally

packages/ui/src/primitives/tailwind/Select/index.tsx
CITIZENDOT: updated

PR #761, opened 2024-12-04
fix offline avatar loading

HexaField:
## Summary

Dev test suite can't spawn avatars in most examples without this

## References
closes #_insert number here_


packages/client-core/src/components/World/LoadLocationScene.tsx
```diff
@@ -88,11 +87,21 @@ export const useLoadLocation = (props: { locationName: string }) => {
...
-    if (!props.sceneName || !props.projectName) return
-    const key = `${props.projectName}/${props.sceneName}`
-    const url = getState(DomainConfigState).cloudDomain + `/projects/${key}`
-    getMutableState(LocationState).currentLocation.location.sceneId.set(key)
-    return GLTFAssetState.loadScene(url, key)
-  }, [])
+    if (!resourceQuery.data.length) return
+    const resource = resourceQuery.data[0]
+    getMutableState(LocationState).currentLocation.location.sceneId.set(resource.id)
+    const unload = GLTFAssetState.loadScene(resource.url, resource.id)
```
AidanCaruso: Changing this to use the resource url includes the hash delimiter which seems to be causing issues with scene loading in my testing

AidanCaruso (changes requested):
Nothing seems to load in the test suite when testing with this branch. Testing after removing everything after the hash delimiter (mentioned below) and with the avatar pack installed, I'm getting loads of possibly unrelated errors on the avatar examples, with none successfully loading.

HexaField:
Urg yeah I think something else is missing. I picked this out of the p2p branch.

HexaField:
hmm it does seem to be working for me. most examples don't load an avatar, the WebXR ones do.

AidanCaruso (approved):
Seems fine, but still for some reason without removing the hash delimiter none of the gltf assets load in the example route. Maybe it's a problem to be addressed with the examples route specifically. I'm not seeing any issues anywhere else.

PR #767, opened 2024-12-05
Align default text component with prefab settings

SYBIOTE:
Update the default properties of the text component to match the prefab configuration, ensuring consistency in text rendering.


packages/engine/src/scene/components/TextComponent.tsx
```diff
@@ -246,12 +247,17 @@ export const TextComponent = defineComponent({
 
     useEffect(() => {
       const troikaMesh = text.troikaMesh.value! as TextMesh
+      console.log('DEBUG in text', text.troikaMesh.value, troikaMesh)
```
HexaField: remove console log

SYBIOTE: done

HexaField (approved):

PR #769, opened 2024-12-06
Removed address and port from prod instanceserver websocket request

barankyle:
## Summary

As part of IR-5570, which is mostly addressed [here](https://github.com/ir-engine/ir-engine-ops/pull/1), the address and port of the instanceserver that a websocket connection is being made to are no longer sourced from query params on the WS address. Instead, the nginx server now looks up the instance record of the already-provided instanceID and proxies to the address in the record, if the record exists.

As the query params are no longer used, this removes them from the connection. In developments settings, they are still used as the direct IP:port of the connection since there normally is no nginx server to act as an intermediary.

## References
closes #_insert number here_


barankyle:
Please do not merge this in after approving, I will do so once I update the helm deployments on (mt-)dev once the linked PR is approved and merged. Just merging this will break deployments without the helm deployment updates.

HexaField:
@barankyle this seems to have been lost to time but still relevant?

HexaField (approved):

barankyle:
It's sort of not relevant any more. This isn't used with P2P I believe. On GCP, we're no longer using nginx, so the additional nginx code I'd written in the ir-engine-ops repo to get the IP/port from the database instead of whatever it was passed on these query params isn't really relevant any more, either.

HexaField:
but if we still don't need the search params in the primus request then we should just remove them right?

barankyle:
It would technically break any AWS deployments, which we don't care about any more, but someone else might. If we merge in the linked PR as well then I think it's fine to merge this too.

PR #771, opened 2024-12-06
Skip scale lock in tab index

SYBIOTE:
Adjust the tab index to skip over the scale lock element, improving keyboard navigation.


HexaField (approved):
nice

PR #773, opened 2024-12-06
Make scaling change uniform across Vector3Input component

SYBIOTE:
Refactor the scaling logic in the Vector3Input component to ensure uniformity in value changes. This improves the handling of vector scaling by calculating the change based on the current value.


aditya-mitra (approved):
LGTM

PR #777, opened 2024-12-06
Ir 5663 mobile warnings for release0.9

DanielBelmes:
## Summary
Add device warnings

## References
closes #_insert number here_


packages/editor/src/hooks/useBrowserCheck.ts
```diff
@@ -30,3 +36,31 @@ export const isSupportedBrowser = () => {
 
   return isGoogleChrome || isSafari
 }
+
+export const useBrowserCheck = () => {
```
aditya-mitra, line 40: we can use the [`useUnsupported`](https://github.com/ir-engine/ir-engine/blob/9c03ef85affc2d566330720cbbb54dacb91c2365/packages/client-core/src/hooks/useUnsupported.tsx/#L43) hook. The advantage is it will show the popover for which we already have ready-made design.

This hook just needs to be added to the location page similar to how it is done in the editor page.

DanielBelmes: This PR is already in Dev actually. This is just a port to release0.9 if you feel like we should change it we can make a new ticket for dev.

HexaField: I've refactored this to be simpler in a draft PR on dev, this is fine for our release

barankyle (approved):

PR #780, opened 2024-12-06
IR-6060: Fix Build Failures in npm run local Command

the-reality-engineer:
This PR fixes two issues causing failures in the `npm run local` command:

1. Environment Variables Processing
- Updated Vite's define configuration to properly handle both VITE_ prefixed and other environment variables
- Created separate process.env object in define configuration
- Added specific handling for VITE_ prefixed variables
- Preserved npm_package_bin variables by nesting them under process.env

2. Tailwind CSS Configuration
- Updated content paths to target only relevant source directories:
  - ./packages/client/src/**/*.{ts,tsx}
  - ./packages/ui/src/**/*.{ts,tsx}
  - ./packages/editor/src/**/*.{ts,tsx}
  - ./packages/projects/**/src/**/*.{ts,tsx}
- Prevented build failures from processing non-existent files
- Improved build performance by limiting scope

Testing:
- [x] Verified successful execution of `npm run local`
- [x] Confirmed environment variables are properly accessible
- [x] Verified CSS styles are correctly applied

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -28,9 +28,11 @@ module.exports = {
   mode: 'jit',
   content: [
     './packages/client/src/**/*.{ts,tsx}',
-    './packages/ui/src/**/*.{ts,tsx}',
+    './packages/client-core/src/**/*.{ts,tsx}',
+    './packages/engine/src/**/*.{ts,tsx}',
     './packages/editor/src/**/*.{ts,tsx}',
-    './packages/projects/**/src/**/*.{ts,tsx}'
+    './packages/projects/**/src/**/*.{ts,tsx}',
+    './packages/ui/src/**/*.{ts,tsx}'
   ],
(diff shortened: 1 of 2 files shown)


barankyle (approved):

tailwind.config.js
```diff
@@ -26,7 +26,12 @@ Infinite Reality Engine. All Rights Reserved.
 /** @type {import('tailwindcss').Config} */
 module.exports = {
   mode: 'jit',
-  content: ['../**/*.{ts,tsx}'],
+  content: [
+    './packages/client/src/**/*.{ts,tsx}',
```
HexaField, line 30: also need client-core probably? and maybe engine if it has XRUIs

HexaField (changes requested):

HexaField (approved):

PR #783, opened 2024-12-07
Fix spawn point helpers

HexaField:
## References
closes #_insert number here_


SYBIOTE (approved):
just a basic white bounding box added to the debug gltf, looks good to me

PR #784, opened 2024-12-07
Fix WSL Port Proxies to LAN

AidanCaruso:
## Summary
Adds a host origin override to auth service, used with local dev on WSL to avoid using the VITE_HOST variable (which must be set to 0.0.0.0 in order to be bound to the Windows ipv4). This restores LAN/remote debugging support for WSL using Windows port proxies.

## References
closes #_insert number here_


HexaField:
Looks good, will let @barankyle give final review. Though maybe we should instead add a config to disable the root cookie since it's not required for local / single tenant in most cases.

barankyle (approved):

PR #790, opened 2024-12-09
Use project name correctly and fix extension logic

CITIZENDOT:
## Summary

Fixes Image Compression. The bug is at two areas:
1. Fetch projectName of the resource instead of `path.split('/')`
2. Fix extension extraction logic. Previously, `default.thumbnail.jpg` used to return `thumbnail` as the extension. Now it correctly returns `jpg`.

## References
closes https://tsu.atlassian.net/browse/IR-6067

### DIFF
--- a/packages/editor/src/components/assets/ImageCompressionPanel.tsx
+++ b/packages/editor/src/components/assets/ImageCompressionPanel.tsx
@@ -32,7 +32,7 @@ import {
   KTX2EncodeArguments,
   KTX2EncodeDefaultArguments
 } from '@ir-engine/engine/src/assets/constants/CompressionParms'
-import { ImmutableArray, useHookstate } from '@ir-engine/hyperflux'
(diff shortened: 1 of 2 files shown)


barankyle:
With the changes to ImageCompressionPanel, what happens if you're trying to compress an image from a different project? Several weeks ago I put in #559 that fixed a bug where patching a file from a different project was incorrectly swapping that file's project's name with the project that was currently open, even if it wasn't the project of that file. At first glance it seems like something similar might happen with this change.

CITIZENDOT:
Thank you for catching this case. This'll indeed throw an error when compressing an image from different project. (or try to place the new compressed file in the scene's project, which is also unintended).

Is there a reliable way of extracting project paths? (considering slashes in project names)

CITIZENDOT:
We can always fetch the static resource details from the table and fetch the projectName, but I was wondering if that info is already with us locally.

For now, I'll fetch it from the static-resource service.

barankyle (approved):

PR #792, opened 2024-12-09
updated 8thwall app key ref to use the config ref

MbfloydIR:
## References
closes #_insert number here_


HexaField (changes requested):
This is incorrect. The spatial package cannot import the common package. The way it was previously should work fine.

HexaField:
Closed, see the jira ticket.

PR #798, opened 2024-12-10
IR-6111: Fixed crash caused by Redis Pub/Sub output buffer limits

jose-galvan:
## Summary
This PR fixes the crash caused by Redis Pub/Sub output buffer limits. 

Files in params were included in requests to static resources which were being emitted 
in Redis Pub Sub by feathers-sync causing the buffer to reach its limits and forcing the connection to close. 
When Redis forces a connection to close the client will emit an error but if there are not listeners 
the process will terminate due to an unhandled exception

See: https://redis.io/docs/latest/develop/reference/clients/#output-buffer-limits

## References
closes #_insert number here_

## QA Steps
- Upload an avatar exported from RPM

### DIFF
--- a/packages/server-core/src/media/upload-asset/upload-asset.service.ts
+++ b/packages/server-core/src/media/upload-asset/upload-asset.service.ts
@@ -118,7 +118,9 @@ const uploadAssets = (app: Application) => async (data: AssetUploadType, params:
       ...(data.args as AvatarUploadArgsType)
     } as any
     if (data.path) callData.path = data.path
-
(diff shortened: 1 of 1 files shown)


barankyle (approved):

AidanCaruso (approved):

DanielBelmes (approved):
@HexaField This is the bug fix for that odd error we were seeing when uploading avatars. So odd this changed randomly.

AidanCaruso:
@DanielBelmes there were some changes to uploadAvatarStaticResource last week (which is just downstream). Perhaps that was part of it. Otherwise I'm also unsure

PR #807, opened 2024-12-10
Fixed some security issues with identity-provider service

barankyle:
## Summary

Create method was able to set fields on new guests that it shouldn't. It now only preserves id, token, type, userId, createdAt, and updatedAt. Also made sure that guest tokens are UUIDs, as one was able to put any value for guest tokens. Now, if that value isn't a UUID, it is replaced with a UUID.

Disabled external patching of identity-providers. This was not being done externally, but it was allowing patching of fields that shouldn't be patched. In general, this is not something that users should be doing, hence the disabling.

Added some identity-provider tests and organized them by method.

Resolves IR-5581

## References
closes #_insert number here_


HexaField (approved):
Looks good to me.

PR #808, opened 2024-12-11
IR-5665 fixing spam for no entityUUID component being found

SamMazerIR:
## Summary
fixing spam for no entityUUID component being found, which apparently also fixes audio volume adjustment issue

## References
[https://tsu.atlassian.net/browse/IR-5665](https://tsu.atlassian.net/browse/IR-5665)

## QA Steps
Open Studio

Add Audio Entity

Add an audio asset to the media component source path

Add a Trigger component

Add target object options within trigger component

Navigate to volume setting within media component

Attempt to adjust volume with the slider or text input

### DIFF
--- a/packages/ui/src/components/editor/properties/trigger/index.tsx
+++ b/packages/ui/src/components/editor/properties/trigger/index.tsx
@@ -23,7 +23,15 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
 Infinite Reality Engine. All Rights Reserved.
 */
 
-import { EntityUUID, UUIDComponent, getComponent, hasComponent, useComponent, useQuery } from '@ir-engine/ecs'
+import {
+  EntityUUID,
+  UUIDComponent,
+  getComponent,
+  getOptionalComponent,
+  hasComponent,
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/editor/properties/trigger/index.tsx
```diff
@@ -78,9 +86,14 @@ const TriggerProperties: EditorComponentType = (props) => {
     for (const entity of callbackQuery) {
       if (!hasComponent(entity, EntityTreeComponent)) continue
       const callbacks = getComponent(entity, CallbackComponent)
+      const label = getOptionalComponent(entity, NameComponent)
```
HexaField: why not just include these in the query?

HexaField (changes requested):

packages/ui/src/components/editor/properties/trigger/index.tsx
SamMazerIR: good point, just did a spot-fix

HexaField (approved):

PR #810, opened 2024-12-11
Move input heuristics into appropriate places

HexaField:
## References
closes #_insert number here_


packages/spatial/src/input/functions/ClientInputHeuristics.test.tsx
```diff
@@ -301,13 +137,10 @@ describe('ClientInputHeuristics', () => {
...
         const rayDirection = new Vector3(2, 2, 2)
-        const ray = new Ray(rayOrigin, rayDirection)
-        const hitTarget = new Vector3()
         const data = new Set<IntersectionData>()
-        assert.equal(data.size, 0)
 
-        ClientInputHeuristics.findBBoxes(data, ray, hitTarget)
-        assertVec.approxEq(boxMin, hitTarget, 3)
+        boundingBoxHeuristic(data, rayOrigin, rayDirection)
+        assertFloat.approxEq(1, Array.from(data)[0].distance, 3)
```
heysokam: Probably meant:
```ts
assertFloat.approxEq(1, Array.from(data)[0].distance, 0.001)
```
but `0.001` is the default epsilon for float approxEq, iirc. So it could be omitted

the-reality-engineer (approved):

packages/spatial/src/input/functions/ClientInputHeuristics.test.tsx
HexaField: Thanks, that was actually hiding a failing test!

PR #812, opened 2024-12-11
Basic Materials

AidanCaruso:
## Summary
Basic materials exist as our solution to the default PBR threejs shaders' performance toll on lower spec devices (standalone xr/mobile), the implementation of which has been broken since the GLTF refactor. This PR restores functionality by refactoring the basic materials logic into the material system, and adds tests.

Due to the current asynchronous loading of materials, basic materials enabled on load are subject to occasional race conditions. This will be addressed after the GLTF loader functions are refactored synchronously, see https://github.com/ir-engine/ir-engine/pull/697
## References
closes https://tsu.atlassian.net/browse/IR-5404


packages/engine/src/scene/materials/systems/MaterialLibrarySystem.tsx
```diff
@@ -45,12 +65,97 @@ const reactor = () => {
...
+    setComponent(fallbackMaterialEntity, MaterialStateComponent, {
+      material: fallbackMaterial,
+      instances: [UndefinedEntity]
+    })
+    setComponent(fallbackMaterialEntity, UUIDComponent, MaterialStateComponent.fallbackMaterial)
+    setComponent(fallbackMaterialEntity, NameComponent, 'Fallback Material')
   }, [])
 
+  const rendererState = useMutableState(RendererState)
+  useEffect(() => {
```
HexaField, line 78: rather than mutating the state here, we should be using `qualityLevel === 0 || forceBasicMaterials === true ` in any effects that depend on `forceMasicMaterials`. When testing locally, the I would turn on basic materials from the debug menu, and (likely due to the few seconds of lag experienced as materials are switched) it would turn back off (because the qualityLevel is changing and triggering the useEffect)

Edit: In fact, it looks like `tieredSettings` is in the performance scaler is doing this too, which is also causing problems with the setting locally. The material changes themselves look fine, so maybe we can address this in a followup.

packages/spatial/src/renderer/materials/MaterialComponent.tsx
```diff
@@ -191,6 +192,8 @@ const MaterialInstanceSubReactor = (props: { array: boolean; uuid: EntityUUID; e
     } else {
       meshComponent.material.set(material)
     }
+
+    materialStateComponent.instances.set([...materialStateComponent.instances.get(NO_PROXY), entity])
```
HexaField: for less verbosity we can do `materialStateComponent.instances.merge([entity])`

HexaField (changes requested):

packages/engine/src/scene/materials/systems/MaterialLibrarySystem.tsx
AidanCaruso: @HexaField Yeah originally my solution was to do just this, but to keep consistency with that preexisting functionality I added this useEffect. Weirdly the tieredSettings config seems to only be properly applied the first time you set quality level to 0 in the settings menu but not when it's 0 on load. We might want to fix this in a follow up

HexaField (approved):

PR #813, opened 2024-12-11
IR-5665 trigger and audiovolume issue

SamMazerIR:
## Summary
found the real culprit this time, infinite retriggering of a useeffect. also fixing spam in default.gltf

## References
[https://tsu.atlassian.net/browse/IR-5665](https://tsu.atlassian.net/browse/IR-5665)

## QA Steps
Open Studio

Add Audio Entity

Add an audio asset to the media component source path

Add a Trigger component

Add target object options within trigger component

Navigate to volume setting within media component

Attempt to adjust volume with the slider or text input

### DIFF
--- a/packages/ui/src/components/editor/properties/trigger/index.tsx
+++ b/packages/ui/src/components/editor/properties/trigger/index.tsx
@@ -84,7 +84,7 @@ const TriggerProperties: EditorComponentType = (props) => {
       })
     }
     targets.set(options)
-  }, [callbackQuery])
+  }, [JSON.stringify(callbackQuery)])
 
   return (
     <NodeEditor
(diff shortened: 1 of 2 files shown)


the-reality-engineer (approved):

the-reality-engineer (approved):

packages/ui/src/components/editor/properties/trigger/index.tsx
```diff
@@ -85,7 +84,7 @@ const TriggerProperties: EditorComponentType = (props) => {
       })
     }
     targets.set(options)
-  }, [callbackQuery])
+  }, [JSON.stringify(callbackQuery)])
```
HexaField, line 87: This stringing shouldn't be necessary, since there is not something that would change the query inside the useEffect. Either useQuery is broken or something else is causing a feedback here.

PR #815, opened 2024-12-11
Improved tooltip

CITIZENDOT:
## Summary

1. Improved tooltip logic:
  - When hovered over the target element, a new div is placed under the root div, that contains the tooltip.
  - When the user moves away from the target element, tooltip div is unmounted from the root div.
- Tooltip's show/hide status can be controlled from the parent (uses `useImperativeHandle` internally).

2. This implementation is designed to accurately reflect Toolbar inputs. ⬇️
- image

3. The dropdowns are required to display tooltips at the bottom as per the design. ⬇️
- image

But also, when the dropdown is clicked/expanded, tooltip needs to disappear => Hence it should be controlled by the parent div. This PR implements this feature and also eliminates the tooltip clipping (due to overflow).

## References
closes #_insert number here_


packages/ui/src/primitives/tailwind/Tooltip/index.tsx
```diff
@@ -23,96 +23,185 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
+}
+
+export type TooltipProps = BaseTooltipProps & (ControlledProps | UncontrolledProps)
+
+export interface TooltipRef {
+  showTooltip: () => void
+  hideTooltip: () => void
+}
+
+function Tooltip(
```
aditya-mitra, line 57: lets add a jsdoc to explain the `ref` provides those 2 functions

```suggestion
/**
 * Ref:

 * Provides an imperative handle with showTooltip and hideTooltip methods to manually control the tooltip visibility.
*/
```

CITIZENDOT: done

packages/ui/src/primitives/tailwind/Select/index.tsx
```diff
@@ -47,6 +47,7 @@ export interface SelectProps<T = string | number> {
   width?: 'sm' | 'md' | 'lg' | 'full'
   inputHeight?: InputProps['height']
   onChange: (value: T) => void
+  onOpenChange?: (isOpen: boolean) => void
```
aditya-mitra: lets rename this to `onOpen` since it indicates `on` indicates there is a handler associated

aditya-mitra:
image

Transform Space Dropdown will be below

CITIZENDOT:
> Transform Space Dropdown will be below

That particular tooltip is positioned at the right because, `bottom` would clip out some portion of the tooltip.

packages/ui/src/primitives/tailwind/Tooltip/index.tsx
```diff
@@ -23,96 +23,188 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
+export interface BaseTooltipProps {
   title?: string
   content: ReactNode
   children: ReactNode
   position?: 'auto' | 'top' | 'bottom' | 'left' | 'right'
 }
 
-function Tooltip({ title, content, children, position = 'auto' }: TooltipProps) {
+export interface ControlledProps {
+  isControlled: true
```
aditya-mitra, line 38:
```suggestion
  isControlled: boolean
```

packages/ui/src/primitives/tailwind/Tooltip/index.tsx
```diff
@@ -23,96 +23,188 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
+  onMouseEnter: () => boolean
+  onMouseLeave: () => boolean
+}
+
+export interface UncontrolledProps {
+  isControlled?: false
```
aditya-mitra, line 44:
```suggestion
  isControlled?: boolean
```

aditya-mitra (approved):
Works good!

packages/ui/src/primitives/tailwind/Tooltip/index.tsx
CITIZENDOT: This is deliberately done like this, so user gets better type hints. When `isControlled` is set to `true`: Tooltip expects `onMouseEnter` and `onMouseLeave` functions as required props. Otherwise it wouldn't expect it.

packages/ui/src/primitives/tailwind/Tooltip/index.tsx
CITIZENDOT: https://github.com/ir-engine/ir-engine/pull/815#discussion_r1883303466

PR #827, opened 2024-12-12
IR-5238: verify user has required project permissions to save node changes

jose-galvan:
## Summary

This change verifies user has the appropriate project permissions to save node changes
This will prevent users to modify prefabs in ir-starter-assets if they don't have the required permissions. 

## References
closes #_insert number here_


packages/editor/src/panels/hierarchy/hierarchynode.tsx
```diff
@@ -336,6 +341,42 @@ export default function HierarchyTreeNode(props: ListChildComponentProps<undefin
...
+    const userID = getState(EngineState).userID
+    const { data } = await API.instance.service(projectPermissionPath).find({
+      query: {
+        project: fullProjectName,
+        userId: userID
+      }
+    })
+    const [permission] = data
+    if (!permission) {
+      return canSaveChanges.set(false)
```
aditya-mitra:
```suggestion
      canSaveChanges.set(false)
```

packages/editor/src/panels/hierarchy/hierarchynode.tsx
```diff
@@ -336,6 +341,42 @@ export default function HierarchyTreeNode(props: ListChildComponentProps<undefin
...
+    const gltfComponent = getComponent(node.entity, GLTFComponent)
+    const [, orgName, projectName] = STATIC_ASSET_REGEX.exec(gltfComponent.src)!
+    const fullProjectName = `${orgName}/${projectName}`
+
+    const { projectName: stateProjectName } = getState(EditorState)
+
+    if (stateProjectName === fullProjectName) {
+      canSaveChanges.set(true)
+      return
+    }
```
aditya-mitra: i am not understanding this part. can you explain it to me? do we allow all editors to make node changes if the project name remains unchanged?

packages/editor/src/panels/hierarchy/hierarchynode.tsx
```diff
@@ -126,6 +129,8 @@ export default function HierarchyTreeNode(props: ListChildComponentProps<undefin
   const { setMenu } = useHierarchyTreeContextMenu()
   const renameRef = useRef<HTMLDivElement>(null)
   let isRenameOpen = useState(false)
+  const canSaveChanges = useState(false)
+  const permissionVerified = useState(false)
```
aditya-mitra: i believe we can do this skipping the usage of useState.

we can use the `useFind` feathers hook to get the project permission and then use calculate the boolean value using `userHasProject Permission`

packages/editor/src/panels/hierarchy/hierarchynode.tsx
jose-galvan: Users can add prefabs to the scene, but some prefabs may belong to public projects they don’t have permission to modify, such as [ir-starter-assets](https://github.com/ir-engine/ir-starter-assets).

Once a prefab is added to the scene, users can modify its properties and save changes but we need to verify whether they have permission to modify assets from public projects.

The if statement compares the projectName in the state (the project the scene belongs to) with the projectName of the modified node.

If they are the same, the user has permission to modify the node and should be allowed to save changes.
If they are different, it means the user added a prefab from a public project, and they may not have permission to modify it, in this case an api call will be performed to fetch project permissions and verify user's access.

packages/editor/src/panels/hierarchy/hierarchynode.tsx
jose-galvan: I decided to not use that hook and use `await API.instance.service(projectPermissionPath).find` instead because we only check project permissions for prefabs/nodes that belong to a different project than the scene, not for all nodes. This should reduce the number of API calls, making it more cost-effective

packages/editor/src/panels/hierarchy/hierarchynode.tsx
aditya-mitra: okay. now i understand.
can you rename these variables to a more a meaningful name as you discussed?

packages/editor/src/panels/hierarchy/hierarchynode.tsx
aditya-mitra: okay.

packages/editor/src/panels/hierarchy/hierarchynode.tsx
jose-galvan: I renamed them, let me know what you think!

aditya-mitra (approved):
Looks good!

DanielBelmes:
Sadly got some conflicts @jose-galvan

PR #828, opened 2024-12-12
Fixed issues with static client build process

barankyle:
## Summary

On some occasions, the static client build process was removing files it had just pushed to S3, resulting in the deployment not functioning because those static files were missing.

The suspected reason was that, after a build had stripped out its current files and pushed that list to S3FilesToRemove.json, but before it pruned away those files after the deployment had been applied, another build process was overwriting S3FilesToRemove.json

The build process now uses an Initial and Final version of S3FilesToRemove.json The former is written by get-deletable-client-files.ts. push-client-to-s3.ts now reads from that, strips out any matching files it just built, and pushes the results to the Final version. delete-old-s3-files.ts uses the Final version to perform the prune.

envmap versions of client files were not getting pruned, nor were client files with dashes or underscores. This was solved by adding to the regex used to match files, and having it look at the key of the file rather than the name. The former problem came from .map files' name being e.g. 'test-abcd1234.js', which was not matching the old regex.

Only macthing files in client/assets, as that is where all of the vite-build files go. There were occasional false matches with e.g. root-cookie-accessor.html simply because 'accessor' has 8 characters like the vite-built suffixes do.

Resolves IR-6178

## References
closes #_insert number here_

### DIFF
--- a/packages/common/src/regex/index.ts
+++ b/packages/common/src/regex/index.ts
@@ -133,3 +133,5 @@ export const VALID_PROJECT_NAME = /^(?!\s)[\w\-\s]+$/
 
 export const MAIN_CHART_REGEX = /ir-engine-([0-9]+\.[0-9]+\.[0-9]+)/g
 export const BUILDER_CHART_REGEX = /ir-engine-builder-([0-9]+\.[0-9]+\.[0-9]+)/g
+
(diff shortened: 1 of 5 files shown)


HexaField (changes requested):
Please add regression tests for the regex and any other relevant changes. If this is urgent, then please make a note in jira that this needs coverage.

HexaField (approved):

PR #831, opened 2024-12-12
changed the order of execution to get the highlight effect to work again

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-6062
## References
closes #_insert number here_


HexaField (changes requested):
Whilst this may fix the issue, we need to make sure the useEffect has the correct dependencies. Let's add `!!rendererComponent.effectComposer.value` as a dependency of the useEffect for the highlight effect, and ensure that it exists. Without this, it could still break (possibly in tests)

HexaField (dismissed):
I'm not seeing the highlight effect in the studio still

DanielBelmes:
I did changes to make the useEffects work as initially expected but the reorder still saves some extranous useeffect runs. It's minuscule but an improvements an improvement. @HexaField

DanielBelmes:
tests fixed

DanielBelmes:
test might be fixed. Didn't test locally because it's 8pm but fingers crossed.

SamMazerIR (approved):

PR #839, opened 2024-12-13
refactor entity layers around ECS and UUIDComponent

HexaField:
## References
closes #_insert number here_


packages/ecs/src/ComponentFunctions.ts
```diff
@@ -561,61 +563,78 @@ export const setComponent = <C extends Component>(
   component.onSet(entity, component.stateMap[entity]!, args)
 
   const propagate = () => {
```
dinomut1, line 560: this function isn't actually getting called so no layer logic is running in setComponent right now

packages/ecs/src/ComponentFunctions.ts
```diff
@@ -561,61 +563,78 @@ export const setComponent = <C extends Component>(
...
-              }
-            }
+    if (!component.jsonID) return
+    if (component === LayerComponent || LayerComponents.includes(component as any)) return
+    if (LayerComponent.get(entity) !== Layers.Authoring) return
+    if (component === UUIDComponent) return
+
+    const uuidComponent = getOptionalComponent(entity, UUIDComponent)
+    if (!uuidComponent) return
+    const linkedEntity = UUIDComponent.getEntityByUUID(uuidComponent, Layers.Simulation)
```
dinomut1: Need to re-implement the enforcement of arbitrary layer relations instead of hard-coding propagation as authoring -> simulation

packages/ecs/src/ComponentFunctions.ts
```diff
@@ -561,61 +563,78 @@ export const setComponent = <C extends Component>(
...
+    setComponent(linkedEntity, component, argsClone)
+    // }
+    // }
   }
 
-  propagate()
+  // propagate()
 
-  if (!componentExists && !component.reactorMap.has(entity)) {
+  if (!componentExists && !component.reactorMap.has(entity) && hasComponent(entity, SimulationLayerComponent)) {
```
dinomut1: Can't disable creating reactors for all entities that are not simulation layer. We want to be able to have reactive logic in other layers too

packages/engine/src/gltf/GLTFState.tsx
```diff
@@ -153,6 +158,59 @@ export const GLTFSourceState = defineState({
...
+      <>
+        {/* The authoring layer entities will have their entities propagated to the simulation layer */}
+        {gltfSimulationEntities.map((entity) => (
+          <GLTFComponentReactor key={'simulation-' + entity} entity={entity} />
+        ))}
+        {gltfAuthoringEntities.map((entity) => (
+          <GLTFAuthoringComponentReactor key={'authoring-simulation-' + entity} entity={entity} />
+        ))}
+        {gltfAuthoringEntities.map((entity) => (
+          <GLTFComponentReactor key={'authoring-' + entity} entity={entity} />
```
dinomut1: authoring layer and simulation layer gltf component entities should not both have reactive logic running. This will cause the same data to be loaded twice instead of loaded once and then propagated according to layer relations. In the original implementation this is done in the gltf component reactor where it checks if an upstream entity exists and defers load responsibility upstream if so

packages/spatial/src/renderer/components/MeshComponent.ts
```diff
@@ -43,7 +43,7 @@ import { addObjectToGroup, removeObjectFromGroup } from './GroupComponent'
 
 export const MeshComponent = defineComponent({
   name: 'MeshComponent',
-  jsonID: 'EE_mesh',
```
dinomut1, line 46: Why are we removing json id for meshes? Don't we want to be able to serialize meshes into a scene?

packages/ecs/src/EntityFunctions.tsx
```diff
@@ -31,24 +31,13 @@ import { HyperFlux } from '@ir-engine/hyperflux'
...
-
-  const layer = EntityLayerState.getEntityLayer(entity)
-  if (!layer) return
-  for (const layerID of Object.keys(layer.relations)) {
-    const relation = layer.relations[layerID as LayerID]
-    if (relation === 'propagate') {
-      const linkedEntity = EntityLayerState.getLinkedEntity(entity, layerID as LayerID)
-      removeEntity(linkedEntity)
-    }
-  }
```
dinomut1, line 51: still need to propagate entity deletion

packages/engine/src/avatar/components/AnimationComponent.ts
```diff
@@ -73,7 +72,6 @@ export const useLoadAnimationFromBatchGLTF = (urls: string[], keepEntities = fal
 
 export const useLoadAnimationFromGLTF = (url: string, keepEntity = false) => {
   const assetEntity = useHookstate(UndefinedEntity)
-  const simEntity = EntityLayerState.useLinkedEntity(assetEntity.value, 'simulation' as LayerID)
```
dinomut1, line 76: This logic will break again once we remove having two concurrent component reactors for both simulation and authoring layer gltf component entities

packages/engine/src/avatar/components/AvatarAnimationComponent.ts
```diff
@@ -158,7 +155,7 @@ export function createVRM(rootEntity: Entity) {
       : formatHumanBones(vrmExtensionDefinition.humanoid!.humanBones as any)
     const bones = humanBonesArray.reduce((bones, bone) => {
       const nodeID = `${documentID}-${bone.node}` as EntityUUID
-      const entity = UUIDComponent.getEntityByUUID(nodeID, layerID)
```
dinomut1, line 161: Will need to check layer ID with deferred upstream gltf loading, otherwise authoring layer vrm rigs get bound to simulation layer bone entities

dinomut1 (changes requested):

packages/ecs/src/ComponentFunctions.ts
```diff
@@ -864,3 +833,167 @@ export const getAllComponentsOfType = <C extends Component>(component: C): Compo
...
+}
+
+export type LayerType = keyof typeof Layers
+export type LayerID = (typeof Layers)[keyof typeof Layers]
+
+export const LayerRelationTypes = {
+  Propagate: 'propagate'
+}
+
+export const LayerRelations = {
```
dinomut1, line 902: This needs to be reactive, so that relations can be defined procedurally

dinomut1 (changes requested):

packages/ecs/src/ComponentFunctions.ts
HexaField: Yeah I disabled it for now, just trying to get loading into both layers working properly

packages/ecs/src/ComponentFunctions.ts
HexaField: I disagree with this.

packages/spatial/src/renderer/components/MeshComponent.ts
HexaField: Meshes are threejs objects, how would they get serialized?

dinomut1 (approved):
ok i think we're good now to sync everything into one branch

packages/ecs/src/ComponentFunctions.ts
dinomut1: resolved

PR #840, opened 2024-12-13
Fix entire scene unloading every time a user disconnects in client/server networking

AidanCaruso:
## Summary
Fixes the entire scene unloading in non-p2p mode every time a user disconnects, by adding a check before removing the networked entity.

## References
Closes https://tsu.atlassian.net/browse/IR-6211


HexaField (commented):
I don't think this is quite right. The error likely comes from something upstream in the mediasoup transport code around handling peers.

HexaField:
addressed here https://github.com/ir-engine/ir-engine/pull/844

PR #841, opened 2024-12-13
Improve Add Project modal

CITIZENDOT:
### UI package changes
- Introduced Searchable dropdowns
  - Search can be of three modes: `prefix`, `substring`, `fuzzy`.
 
### Modal changes
- Copy button is aligned to the end of the input (without that teeny tiny gap)
- Branch and Commit Dropdowns are **searchable**, and hence text cursor appears now.
  - Branch name uses **fuzzy** search
  - Commits uses **substring** seach
- Max Height of branch and commit dropdowns is set to 200px, thus preventing the dropdown overflowing the actual Modal
- Increased z-index of the dropdowns
- Improved dropdown label structure/format
- Added clear button for the dropdowns in this modal

## References
closes https://tsu.atlassian.net/browse/IR-5537


barankyle (changes requested):
I know this issue technically only covered the project add/update modal, but can you also apply those fixes to UpdateEngineModal as well? It has similar problems, and it's causing some minor headaches for the QA team. I was temporarily able to make it render locally by changing ProjectTopMenu.tsx:118 from `config.client.localBuildOrDev` to `!config.client.localBuildOrDev`, and then in project-helper.ts, at the end of findBuilderTags, just returning an array of results that matched the form of the response manually.

The "Destination URL valid and accessible", "Source URL Valid and accessible", and "Source matches Destination or is empty" indicators are still not being rendered when the project modal is in update mode.

I like the text-search capability on the branch and commit. Only nitpick is that when you open it in update mode, when you start typing, it preserves all of the value that was already there, which can lead to some overlapping results. But maybe it's not an issue; most people are probably just going to click to open the dropdown menu, click an option, and finish.

CITIZENDOT:
> but can you also apply those fixes to UpdateEngineModal as well? It has similar problems

Done: Set the width of dropdown to 100%, and updated it's dropdown option's format.



> The "Destination URL valid and accessible", "Source URL Valid and accessible", and "Source matches Destination or is empty" indicators are still not being rendered when the project modal is in update mode.

Fixed it.

> Only nitpick is that when you open it in update mode, when you start typing, it preserves all of the value that was already there, which can lead to some overlapping results

I don't quite understand this. Should we clear input on click?

https://github.com/user-attachments/assets/231a9a41-e5f5-49e7-bb84-4243e31aa94a

CITIZENDOT:
Added clear button on the right
image

barankyle:
The Update Engine selector is still disappearing if there are a lot of options to choose from. In this example, I had the API server return about 20 entries, but the first 7 are cut off.

Screenshot from 2024-12-16 13-56-49

CITIZENDOT:
> The Update Engine selector is still disappearing if there are a lot of options to choose from. In this example, I had the API server return about 20 entries, but the first 7 are cut off.

Fixed

barankyle (approved):

PR #842, opened 2024-12-13
IR-6241 - fix: example routes mounting and state update issues

the-reality-engineer:
1. Fixes several React warnings and renderer component / viewerEntity lifecycle issues that were affecting the ability to switch between examples in the dev-test-suite 

2. Adding API call logging for debugging

Paired with https://github.com/ir-engine/ir-development-test-suite/pull/22


packages/common/src/utils/FeathersHooks.tsx
```diff
@@ -174,7 +174,7 @@ export const useService = <S extends keyof ServiceTypes, M extends Methods>(
   }
 
   // use immediate effect to get the stack trace of the react context, then add it to the state
-  useImmediateEffect(() => {
+  useEffect(() => {
```
HexaField: this means that our stacktraces are useless. what does this fix?

HexaField:
Can you try again with the latest of the dev branch on the dev test suite? i merged some changes in yesterday that may solve the same problems

packages/common/src/utils/FeathersHooks.tsx
the-reality-engineer: I was trying to fix react warnings from modifying state in a component that hasn't mounted yet, but upon further investigation I don't think those warnings were the culprit for the issues I was seeing (just no-op react warnings), I reverted this change.

packages/spatial/src/input/functions/ClientInputHooks.test.tsx
```diff
@@ -53,16 +53,16 @@ import { InputState } from '../state/InputState'
 import ClientInputHooks from './ClientInputHooks'
 
 const createMockHTMLCanvasElement = (ev: MockEventListener) => {
-  return {
-    addEventListener: ev.addEventListener,
-    removeEventListener: ev.removeEventListener,
-    getDrawingBufferSize: () => 0,
-    getContext: () => {},
-    parentElement: {
+  return new (class {
```
HexaField, line 55: with https://github.com/ir-engine/ir-engine/pull/854 we can retire this custom mock function and use a standardized mock

packages/hyperflux/src/hooks/useImmediateEffect.ts
```diff
@@ -34,24 +34,42 @@ function depsDiff(deps1, deps2) {
...
     if (cleanupRef.current) {
       cleanupRef.current()
     }
 
+    // run effect
     cleanupRef.current = effect()
   }
 
+  // make sure deps are hooked
+  useEffect(() => {}, deps)
```
HexaField: can we put this arrow function in module scope to reduce garbage?

packages/editor/src/panels/viewport/index.tsx
```diff
@@ -168,11 +168,9 @@ function ViewportContainer() {
         {sceneName.value ? <SelectionBox viewportRef={ref} toolbarRef={toolbarRef} /> : null}
         {sceneName.value ? <TransformGizmoTool /> : null}
         {sceneName.value ? <CameraGizmoTool viewportRef={ref} toolbarRef={toolbarRef} /> : null}
+        <div id="engine-renderer-canvas-container" ref={ref} className="absolute h-full w-full" />
```
HexaField, line 171: does this still show the logo correctly when no scene is loaded?

packages/common/src/utils/FeathersHooks.tsx
```diff
@@ -158,6 +158,7 @@ export const useService = <S extends keyof ServiceTypes, M extends Methods>(
     // prettier-ignore
     return API.instance.service(serviceName)[method](...args)
       .then((res) => {
+        console.log(`API: ${serviceName}.${method}`, ...args, res)
```
HexaField: pleaser remove logging

HexaField (changes requested):

packages/editor/src/panels/viewport/index.tsx
the-reality-engineer: yes

packages/common/src/utils/FeathersHooks.tsx
the-reality-engineer: I thought it was useful, but sure.

packages/spatial/src/input/functions/ClientInputHooks.test.tsx
the-reality-engineer: The problem with this is that the standard event listener API does not allow you to check if the event exists in the event listener.

the-reality-engineer:
There are still some occasional issues when switching between example scenes (loading state is sometimes stuck at 0%), but it's working better than before.

packages/spatial/src/input/functions/ClientInputHooks.tsx
```diff
@@ -348,7 +351,7 @@ export const CanvasInputReactor = () => {
       canvas.removeEventListener('click', onClick)
       canvas.removeEventListener('wheel', onWheelEvent)
     }
-  }, [xrState.session])
+  }, [xrState.session, rendererComponent])
```
HexaField: we probably want this to be rendererComponent.canvas, rather than potentially reacting to anything changing in the renderer component

HexaField (changes requested):

packages/spatial/src/input/functions/ClientInputHooks.tsx
the-reality-engineer: Good point

HexaField (approved):

PR #845, opened 2024-12-14
[IR-5703] Unit Test - Spatial XR coverage: Component & System Reactors

heysokam:
## References
https://tsu.atlassian.net/browse/IR-5703


packages/spatial/src/xr/XRDepthOcclusion.tsx
```diff
@@ -283,12 +283,11 @@ const reactor = () => {
   const xrState = useMutableState(XRState)
 
   useEffect(() => {
-    if (!xrState.sessionActive.value) {
+    return () => {
```
HexaField: we still need a session active check here, otherwise this will run when a session is made active too

HexaField (changes requested):

packages/spatial/src/xr/XRDepthOcclusion.tsx
heysokam: Oh, good catch! I was meaning to revert this yesterday after asking you guys about it, but totally forgot.
Fixed with 640ac97

HexaField (approved):

PR #851, opened 2024-12-16
Ir 1916 dulicate scene publish

JT00y:
## References
closes #_insert number here_


dinomut1 (changes requested):
Getting an access error on creating the publish directory: 
image
Also the folder path needs to be "public/publish"

PR #852, opened 2024-12-17
Fix dynamic bounding boxes in viewport

SYBIOTE:
Correct the handling of dynamic bounding boxes by ensuring the bounding box updates appropriately when the transform changes. This improves the accuracy of the bounding box representation in the viewport.


the-reality-engineer (dismissed):
It looks like we are updating the bounding box when the `debugEnabled` flag is set to true. Shouldn't the behavior of the bounding box be decoupled from whether or not the debug mode enabled? The `debugEnabled` flag should only control whether or not the bounding box is visible, not when it is updated.

HexaField:
> It looks like we are updating the bounding box when the `debugEnabled` flag is set to true. Shouldn't the behavior of the bounding box be decoupled from whether or not the debug mode enabled? The `debugEnabled` flag should only control whether or not the bounding box is visible, not when it is updated.

`updateBoundingBox` is called in the transform system too, so it shouldn't be necessary here at all

SYBIOTE:
> It looks like we are updating the bounding box when the `debugEnabled` flag is set to true. Shouldn't the behavior of the bounding box be decoupled from whether or not the debug mode enabled? The `debugEnabled` flag should only control whether or not the bounding box is visible, not when it is updated.

hmm, ok, then ill move the code around , but as @HexaField said, its already updated when we move the object

SYBIOTE:
> > It looks like we are updating the bounding box when the `debugEnabled` flag is set to true. Shouldn't the behavior of the bounding box be decoupled from whether or not the debug mode enabled? The `debugEnabled` flag should only control whether or not the bounding box is visible, not when it is updated.
> 
> `updateBoundingBox` is called in the transform system too, so it shouldn't be necessary here at all



the bounding box is updated after the condition here because the without the helper entity being set, the update is not performed 
next, 
when we just add an object to the scene by drag and drop and haven't moved it, the transforms need to be appropriately calculated, so, I mark as dirty so the transform would be updated

bounding box would be recalculated with the appropriate transforms 
when u trigger the debug visible, it is also updated when the object itself is moved too

HexaField (approved):

PR #414, opened 2024-10-11
[IR-3427] Unit Test - Spatial XR coverage: Functions & Static Data

heysokam:
## References
https://tsu.atlassian.net/browse/IR-3427
Split from: https://tsu.atlassian.net/browse/IR-5704

PR #860, opened 2024-12-18
[IR-5702] Unit Test - Spatial XR coverage: System Executes

heysokam:
## References
https://tsu.atlassian.net/browse/IR-5702


heysokam:
@HexaField Can't figure out how to rebase, to clean the redundant commits from #414, without creating a ton of conflicts.
Should be doable, but it feels unnecessary, since the commits are the exact same that are on dev.
I spent 1h or so trying to rebase, but couldn't solve it. Let me know if I need to figure it out.

The todo tests are all with `it.todo` and marked with a doc-comment.
The tests for both this PR and #845 all pass, and both PRs should be good to merge. :writing_hand:

PR #861, opened 2024-12-18
IR-6300 Enhance engine settings handling with data type parsing and updates

mshafiqmk:
Improve the handling of engine settings by incorporating data type parsing, ensuring that each setting's data type is accurately determined and stored.

This change enhances the overall integrity and usability of engine settings.


barankyle (changes requested):
This is mostly good, but having getDataType infer the type from the value doesn't work for the boolean values that were converted. Since they were stored as a 0 or 1, getDataType thinks they're integers, and is setting dataType as `integer` accordingly. This is most likely to be a problem when feature-flag-settings is converted, since flagValue there is a boolean.

Note that I'm only seeing this when dealing with an engine-setting table that's already migrated booleans to 1s or 0s. When I do a full DB reinit, the booleans are being inserted as expected, as either 'true' or 'false'.

I suppose there's technically a possibility of actual string values of 'true' or 'false' being labeled as booleans by mistake, but since this is solely in the engine-settings table, and none of the other actual string settings have values anywhere near 'true' or 'false', it's most likely fine to infer that 'true' or 'false' are actually booleans. But if we could do this without the inherent inferences we're doing, that would be better.

mshafiqmk:
> This is mostly good, but having getDataType infer the type from the value doesn't work for the boolean values that were converted. Since they were stored as a 0 or 1, getDataType thinks they're integers, and is setting dataType as `integer` accordingly. This is most likely to be a problem when feature-flag-settings is converted, since flagValue there is a boolean.
> 
> Note that I'm only seeing this when dealing with an engine-setting table that's already migrated booleans to 1s or 0s. When I do a full DB reinit, the booleans are being inserted as expected, as either 'true' or 'false'.
> 
> I suppose there's technically a possibility of actual string values of 'true' or 'false' being labeled as booleans by mistake, but since this is solely in the engine-settings table, and none of the other actual string settings have values anywhere near 'true' or 'false', it's most likely fine to infer that 'true' or 'false' are actually booleans. But if we could do this without the inherent inferences we're doing, that would be better.

@barankyle I think We can fix the issue of booleans stored as '1' or '0' in the migration.

barankyle (approved):

PR #874, opened 2024-12-20
IR-6311: use avatar 2D render from RPM as thumbnail

jose-galvan:
## Summary

Use 2D render from Ready Player Me as avatar thumbnail

image

image

## References
closes #_insert number here_


packages/client-core/src/user/components/UserMenu/menus/AvatarCreatorMenu2.tsx
```diff
@@ -178,24 +186,14 @@ const AvatarCreatorMenu = (selectedSdk: string) => () => {
     }
     loading.set(LoadingState.Uploading)
 
-    const canvas = document.createElement('canvas')
```
HexaField, line 190: we still need to use this for non-RPM avatars, please keep the original functionality for user-uploaded avatars (which is under a feature flag)

HexaField (dismissed):

the-reality-engineer (approved):

PR #878, opened 2024-12-24
IR-6402: Fix Auth Flow for Safari

the-reality-engineer:
Refactor and Simplify Storage Access Logic

- Removed storage access permission checks from AuthService.ts and root-cookie-accessor-template.html.
- Simplified logic to rely solely on `document.hasStorageAccess()` for determining storage access.
- Removed unnecessary checks and error handling related to storage access permissions.
- Updated button IDs and functionality in root-cookie-accessor-template.html to reflect new logic:
  - Changed `cancelButton` to `denyButton`.
  - Streamlined the process for requesting storage access and handling user denial.


HexaField:
LGTM, will let @barankyle sign off before merging

the-reality-engineer:
Tried to test on MT QAT, but that deployment seems to be having issues with account sign-up (subdomain generation). @larryatinfinite @barankyle

barankyle (approved):
This looks to still all be working properly to me, even with a completely cross-domain usage (which we don't get on MT since those are still all using the same top-level domain - I modified the hostfile locally to check multiple separate domains). 

I think some of the logic that this removes was needed for Chrome when I first coded it, but I also think Chrome changed storage access sometime in the last few months that streamlined things and made those checks unnecessary. Different domains seem to be implicitly trusted, as they're all (locally) getting the needed cookies without needing to prompt for permission. Firefox still requires it, but it's working as intended.

PR #880, opened 2024-12-26
Fix IK and move to separate component and system

AidanCaruso:
## Summary
Fixes IK for the local client, and over the network. Breaks out IK logic into its own system, removes dependency on the IK user being a networked avatar, and adds tests.

## References
closes #_insert number here_


HexaField (changes requested):
Can we captialize "IK" in the system & component for consistency please

packages/engine/src/avatar/systems/AvatarIkSystem.tsx
```diff
@@ -0,0 +1,396 @@
...
+      local.elements[10] = 1
+      local.elements[11] = 0
+      local.elements[15] = 1
+
+      // quick dirty bone rotations for the ik solve to start relative to
+      if (bone.includes('Arm') || bone.includes('Hand')) {
+        if (bone.includes('left')) local.multiply(new Matrix4().makeRotationFromEuler(shoulderAngle.leftShoulderAngle))
+        else local.multiply(new Matrix4().makeRotationFromEuler(shoulderAngle.rightShoulderAngle))
+      }
+      if (bone.includes('Arm')) local.multiply(new Matrix4().makeRotationFromEuler(new Euler(Math.PI * -0.5, 0, 0)))
```
HexaField: let's freeze the static matrix4s as module scope definitions

HexaField (changes requested):

HexaField:
Just nitpicking above, once those are addressed this is good to merge!

HexaField (approved):

PR #889, opened 2025-01-03
Fix: support for Avaturn

jose-galvan:
## Summary

Set default value for previewEnabled flag. this small fixed is required to support Avaturn when menu is open from the UploadAvatar component

## References
closes #_insert number here_


CITIZENDOT:
Thank you!

CITIZENDOT (approved):

PR #895, opened 2025-01-06
IR-5923: Add user-friendly error message for file upload size limit

kattsushi:
…xceeded. Inform users with That file size is too large. Please upload a file under 1GB.

## References
closes #_insert number here_


packages/editor/src/functions/assetFunctions.ts
```diff
@@ -243,10 +245,21 @@ export const inputFileWithAddToScene = ({
     el.onchange = async () => {
       try {
         if (el.files?.length) {
-          const newFiles = sanitizeFiles(el.files)
-          const uniqueFiles = await filterExistingFiles(projectName, directoryPath, newFiles)
-
-          await handleUploadFiles(projectName, directoryPath, uniqueFiles)
+          const validFiles = Array.from(el.files).filter((file) => file.size <= config.client.maxFileSizeToUpload)
```
aditya-mitra: if I upload 2 files (one of them being under 1 gb and one above 1 gb), the error message is absent

we can also show the file name when warning the user, similar to how it is done in [`sanitizeFiles` function](https://github.com/ir-engine/ir-engine/blob/26f34766a9c1629ab45aa90baff2e927872b5a3b/packages/editor/src/functions/assetFunctions.ts/#L107-L119). we can also consider refactoring the entire logic to `santizeFiles`

kattsushi: Thanks @aditya-mitra good catch up, already addressed! 👍🏾

aditya-mitra (approved):

aditya-mitra (approved):

PR #906, opened 2025-01-08
Ir 1850 on public combine mesh

JT00y:
## References
closes #_insert number here_


packages/editor/src/panels/files/PublishModal.tsx
```diff
@@ -0,0 +1,450 @@
...
+  { label: 'Private', value: 'private' },
+  { label: 'Public', value: 'public' },
+  { label: 'Showroom', value: 'showroom' }
+]
+export const PublishModalWithProvider = (props) => (
+  <CurrentFilesQueryProvider>
+    <PublishModal {...props} />
+  </CurrentFilesQueryProvider>
+)
+export default function PublishModal(props: {
```
HexaField: This seems to be a copy/fork of the add/edit location modal. Are we able to consolidate these to ensure we aren't duplicating functionality? cc @aditya-mitra @CITIZENDOT

HexaField (changes requested):

packages/editor/src/panels/files/PublishModal.tsx
JT00y: We want to move the previous location Modal to editor folder with this PublishModal, is this the good way to do?

HexaField: We should break both modals down into smaller parts and reuse the common parts between them. This comes down to the designs, do we have designs for this new publish modal?

packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -120,14 +120,14 @@ const ensureProjectPermissionAndPublicOrAssetsDirectory = async (
...
 
-  if (!publicRegExp.test(resolvedPath) && !assetsRegExp.test(resolvedPath))
-    throw new Error(
-      'Not allowed to access this directory or file: ' +
-        path.join(inputPath) +
-        ' as it does not match the specified project: ' +
-        projectName +
-        ' or it is not in the public or assets folder'
-    )
+  //   if (!publicRegExp.test(resolvedPath) && !assetsRegExp.test(resolvedPath))
```
HexaField: we definitely cannot be disabling our security measures. what is breaking without this change?

packages/editor/src/panels/files/PublishModal.tsx
HexaField: this still looks like a copy of the other modal we have. can we deduplicate these where possible?

HexaField (dismissed):

packages/editor/src/panels/files/PublishModal.tsx
JT00y: Yes I will remove it.

dinomut1 (approved):

PR #909, opened 2025-01-08
fix(location): add the id when user update the location [IR-6269]

mikeplascdev:
## Summary
- add the id property when the user updates the location item
## References
https://tsu.atlassian.net/browse/IR-6269


hanzlamateen (changes requested):
image

I am still getting the error.

hanzlamateen (approved):

PR #922, opened 2025-01-10
fix element list errors and text wrapping

aditya-mitra:
## Summary

- fix multiple re-rendering in useEffect due to non memoized hook value
- fix react warnings in element list
- fix line wrapping problems

#### previous

image

#### now

image


CITIZENDOT (approved):
Works nicely!

PR #926, opened 2025-01-11
Implement selection-based helper visibility

SYBIOTE:
Enhance the visibility logic of helpers by incorporating the presence of a SelectTagComponent, allowing for more dynamic rendering based on selected entities.

### DIFF
--- a/packages/spatial/src/renderer/components/lights/DirectionalLightComponent.ts
+++ b/packages/spatial/src/renderer/components/lights/DirectionalLightComponent.ts
@@ -140,7 +140,6 @@ export const DirectionalLightComponent = defineComponent({
     }, [])
 
     useEffect(() => {
-      console.log('DEBUG run this shit')
       light.color.set(directionalLightComponent.color.value)
     }, [directionalLightComponent.color])
 
--- a/packages/spatial/src/renderer/components/lights/DirectionalLightComponent.ts
+++ b/packages/spatial/src/renderer/components/lights/DirectionalLightComponent.ts
@@ -41,6 +41,7 @@ import { useHookstate, useImmediateEffect, useMutableState } from '@ir-engine/hy
 import { EntityTreeComponent, UndefinedEntity } from '@ir-engine/ecs'
 import { S } from '@ir-engine/ecs/src/schemas/JSONSchemas'
 import { SelectTagComponent } from '@ir-engine/engine/src/scene/components/SelectTagComponent'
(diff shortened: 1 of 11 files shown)


SYBIOTE:
there is a small issue, im referencing the selectTagComponent m engine in the spatial package, if this is an invalid solution, ill create a separate tag for spatial and manage that in the engine package

HexaField:
> there is a small issue, im referencing the selectTagComponent m engine in the spatial package, if this is an invalid solution, ill create a separate tag for spatial and manage that in the engine package

There's a few options. Rather than having a concept of "select" in spatial or engine, it could be a "helper active tag component", and we could have a system in the engine package that handles the logic for adding helpers rather than doing that on the component reactors.

SYBIOTE:
> > there is a small issue, im referencing the selectTagComponent m engine in the spatial package, if this is an invalid solution, ill create a separate tag for spatial and manage that in the engine package
> 
> There's a few options. Rather than having a concept of "select" in spatial or engine, it could be a "helper active tag component", and we could have a system in the engine package that handles the logic for adding helpers rather than doing that on the component reactors.

thats precisely what i built out initially, then i saw that the select tag exists so i felt it would be redundant, cool ill revert some commits and we will be back to that implementation

SYBIOTE:
implemented it based on activeHelperTag

packages/spatial/src/renderer/components/lights/DirectionalLightComponent.ts
```diff
@@ -128,12 +133,14 @@ export const DirectionalLightComponent = defineComponent({
...
       directionalLightComponent.light.set(light)
       setComponent(entity, ObjectComponent, light)
+
       return () => {
         removeComponent(entity, ObjectComponent)
       }
     }, [])
 
     useEffect(() => {
+      console.log('DEBUG run this shit')
```
SamMazerIR: might want to remove this log :)

SYBIOTE: yup

SamMazerIR (approved):

HexaField:
This has introduced invalid imports. Imports from the engine package in the spatial package.

HexaField:
Fixed here https://github.com/ir-engine/ir-engine/commit/7cd626d9e699dcf0363ef21e6eff4e79330a513b

PR #927, opened 2025-01-13
IR-R6017 add loading animation to viewer startup 

kattsushi:
## Summary

https://github.com/user-attachments/assets/c0aaf755-f815-47fe-87ad-b2d00d21e2cc

## References
closes #_insert number here_


HexaField (changes requested):
We definitely need a more scalable and efficient form than 79 PNGs. Maybe Lottie JSON or animated SVG.

aditya-mitra:
I talked with Paul regarding this. 
This will be one-off case for a temporary period of time to be replaced with a robust svg animation solution. (Details of origin - this animation was made by marketing in blender. This later was exported and various methods like bitmapping were used to reduce the file.)

We have a few other solutions like GIF (not advised because the quality will be reduced) and video-codec formats that can be used.
Should we go ahead with an image frame by frame solution or a single video format?

kattsushi:
> We definitely need a more scalable and efficient form than 79 PNGs. Maybe Lottie JSON or animated SVG.

Yes, @Paul and I decided to go with the image sequence solution instead of using a video. Initially, the plan was to use SVGs or WebP images, as they are much more scalable in terms of load time and rendering compared to PNGs. However, the formats I received in Dropbox were only working in PNG, which is why I ended up using them.

That said, I ran into a series of performance issues, which I’ll explain in detail later, but the key point here is that I didn’t choose to go with an external dependency like `react-lottie `to load a JSON file, because in terms of performance, it would have been quite similar. According to [Bundlephobia](https://bundlephobia.com/package/react-lottie@1.2.10), the data load with the library and the JSON file would be almost the same with no benefit at all

This is outside the scope of the current PR, but I found several optimization opportunities regarding the viewer's load time. Specifically, in the LocationPage component, there is an issue with the rendering logic. The current approach doesn't behave as expected because what should be a Suspense with a fallback to the login component ends up rendering both components simultaneously. This results in both components using all available resources at once, causing unnecessary computation and slowing down the page load.
Current LocationPage implementation:
`etherealengine/packages/projects/projects/theinfinitereality/eepro-multitenancy/components/pages/Location/LocationPage.tsx`
```
const LocationPage = () => {
  const ready = useMutableState(LoadingUISystemState).ready.value;

  const currentUrl = window.location.origin;
  const clientUrl = config.client.clientUrl;

  useEffect(() => {
    if (!isDev && currentUrl === clientUrl) {
      RouterState.navigate('/dashboard');
    }
  }, [currentUrl, clientUrl]);

  return (
    <>
      {!ready && <LocationLoadingUI />} // <===
      <LocationRoutes />
    </>
  );
};
export default LocationPage;
```

the issue is that while the ready state is false, both the LocationLoadingUI and LocationRoutes are rendered simultaneously. Ideally, LocationRoutes should not render until ready is true, to optimize resource usage and improve the page load performance.

Additionally, if we look inside the LocationRoutes component, there is another Suspense for loading. This introduces more computation into the flow, making the loading process even slower. Multiple layers of Suspense lead to unnecessary rendering, which can negatively affect performance.
Current LocationRoutes implementation:

```
const LocationRoutes = () => {
  const ref = useRef<HTMLElement>(document.body);

  useSpatialEngine();
  useEngineCanvas(ref);
  useBrowserCheck();

  const projectsLoaded = useEngineInjection();

  if (!projectsLoaded)
    return <LoadingView fullScreen className="block h-12 w-12" title={t('common:loader.loadingProjects')} />;// <===

  return (
    <Suspense
      fallback={<LoadingView fullScreen className="block h-12 w-12" title={t('common:loader.loadingLocation')} />}// <===
    >
      <Routes>
        <Route path=":locationName" element={<LocationPage online />} />
      </Routes>
      <Debug />
    </Suspense>
  );
};
```
This leads to multiple Suspense components being stacked on top of each other, which causes the first loading screen to display above the others only by z-index up, but doesn't help with performance and leads to unexpected behavior.

If we dive deeper into another child component, such as LocationPage from the core, we see that it has similar logic, which compounds the issue. This results in multiple loading screens and slower performance.
Example from LocationPage core component:

```
return (
  <>
    <ThemeContextProvider>
      <StyledEngineProvider injectFirst>
        {!ready.value && (
          <LoadingView fullScreen className="block h-12 w-12" title={t('common:loader.loadingEngine')} /> // <===
        )}
        <LocationIcons />
      </StyledEngineProvider>
    </ThemeContextProvider>
  </>
);
```

That said, to make the loading process more efficient and scalable, we should avoid rendering LocationRoutes and other child components while ready is false. This can be done by moving the conditional rendering of LocationRoutes to ensure it only renders once the necessary data or components are ready.

This change would prevent unnecessary computation and improve the page load time by avoiding redundant Suspense components and overlapping loading states.

I believe the performance issues are having a greater impact than the animation implementation itself, and addressing these underlying concerns could significantly improve the overall experience.

aditya-mitra:
Good part pointing out those optimisations @kattsushi . These can certainly be implemented in the current engine.

For the animation part, can you try adding the animation video in an alternative PR and check how it performs? If it works well, we can surely replace the images.

the-reality-engineer:
very nice animation!

kattsushi:
> Good part pointing out those optimisations @kattsushi . These can certainly be implemented in the current engine.
> 
> For the animation part, can you try adding the animation video in an alternative PR and check how it performs? If it works well, we can surely replace the images.

sure thing! here is [#937](https://github.com/ir-engine/ir-engine/pull/937)

kattsushi:
> We definitely need a more scalable and efficient form than 79 PNGs. Maybe Lottie JSON or animated SVG.

@HexaField After some research and investigation, I came across this utility: svgasm (https://github.com/tomkwok/svgasm) It allowed me to transform all the frames provided by the UI team yesterday into a single animated SVG file. It turned out nicely done now 

—thank you for the feedback!

DanielBelmes:
The animated SVG looks amazing and will be much easier to replace in the future if needed. Well done. Testing locally

packages/ui/src/primitives/tailwind/LoadingView/index.tsx
```diff
@@ -22,11 +22,10 @@ Original Code is the Infinite Reality Engine team.
 All portions of the code written by the Infinite Reality Engine team are Copyright © 2021-2023 
 Infinite Reality Engine. All Rights Reserved.
 */
-
+import Text from '@ir-engine/ui/src/primitives/tailwind/Text'
 import React from 'react'
 import { twMerge } from 'tailwind-merge'
-
-import Text from '@ir-engine/ui/src/primitives/tailwind/Text'
+import LoadingAnimation from './animation.svg?react'
```
DanielBelmes, line 28: Nice that should ensure it's in the source and is loaded on page load not after

DanielBelmes:
image
Formatting was a bit odd when I turned animated on for my display but overall it's looking pretty good. It loads very fast and I think we do just in general have some route rendering issues.

DanielBelmes (approved):

DanielBelmes:
Linked PR https://github.com/theinfinitereality/irpro-multitenancy/pull/692

HexaField (approved):

PR #931, opened 2025-01-13
added conditional check to see if the cameraOrbit.focusedEntities is …

MbfloydIR:
…not null

## Summary
https://tsu.atlassian.net/browse/IR-6549
## References
closes #_insert number here_

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,6 +160,7 @@
     "jsdom-global": "3.0.2",
     "minio": "7.1.3",
     "patch-package": "^6.5.1",
+    "react-icons": "^5.4.0",
     "ts-node": "10.9.2",
     "typescript": "5.4.5"
   }
--- a/package.json
+++ b/package.json
@@ -159,8 +159,7 @@
     "jsdom": "25.0.1",
     "jsdom-global": "3.0.2",
     "minio": "7.1.3",
-    "patch-package": "^6.5.1",
-    "react-icons": "^5.4.0",
+    "patch-package": "^6.5.1",    
     "ts-node": "10.9.2",
     "typescript": "5.4.5"
   }
--- a/package.json
+++ b/package.json
@@ -159,7 +159,7 @@
     "jsdom": "25.0.1",
     "jsdom-global": "3.0.2",
     "minio": "7.1.3",
-    "patch-package": "^6.5.1",    
+    "patch-package": "^6.5.1",
     "ts-node": "10.9.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -160,6 +160,7 @@
     "jsdom-global": "3.0.2",
     "minio": "7.1.3",
     "patch-package": "^6.5.1",
+    "react-icons": "^5.4.0",
```
HexaField: please remove this dependency change

packages/spatial/src/camera/systems/CameraOrbitSystem.tsx
```diff
@@ -146,18 +146,20 @@ const execute = () => {
...
-            if (obj) box.expandByObject(obj)
-          }
-          if (box.isEmpty()) {
-            const entity = cameraOrbit.focusedEntities[0].value
-            const position = getComponent(entity, TransformComponent).position
-            editorCameraCenter.copy(position)
-          } else {
-            box.getCenter(editorCameraCenter)
-            distance = box.getBoundingSphere(sphere).radius
+          if (cameraOrbit.focusedEntities.value) {
```
HexaField: `cameraOrbit.focusedEntities.value` will always be truthy, since it's an array. did you mean to write `cameraOrbit.focusedEntities.value.length > 0` ?

HexaField (changes requested):

packages/spatial/src/camera/systems/CameraOrbitSystem.tsx
HexaField: Oh actually, this is already in an else checking for that, so even the length check will not do anything

MbfloydIR: @HexaField 
the cameraOrbit.focusedEntities.value is null when you zoom in and there are no focused entities.

The condition 
if (cameraOrbit.focusedEntities.length === 0) {
returns false when the cameraOrbit.focusedEntities.value is null

This is why I added the additional conditional check

HexaField: it should not be set null. we need to be looking at where it is being set as null. it is typed as an array of entities, so somewhere we are receiving a null or UndefinedEntity and setting that. that's the 'real' fix here, since the typing is lost anywhere else we are accessing the data

HexaField (approved):

PR #941, opened 2025-01-15
Partial update to support GCP services

barankyle:
## Summary

Added a new script for pruning Artifact Registry images. Updated every place prune_ecr_images to alternatively call that if running in/against GCP.

Updated findBuilderTags to work with Artifact Registry.

Updated run-builder.sh to use GCP. Some things are just stubbed out, like setting up the GCP CLI instead of the aws CLI.

Changed AWS_REGION variable to CLOUD_REGION to avoid having a lot of basically copy-pasted lines with a GCP-specific variable.

Resolves IGM-62

## References
closes #_insert number here_


barankyle:
This was included as part of #985

PR #948, opened 2025-01-16
Fix and layer camera gizmo functionality

SYBIOTE:
Enhance the camera gizmo by fixing its functionality and placing it on the appropriate gizmo layer for better organization and rendering.


packages/spatial/src/input/components/InputPointerComponent.ts
```diff
@@ -101,7 +101,7 @@ export const InputPointerComponent = defineComponent({
 
   usePointersForCamera(cameraEntity: Entity) {
     const pointers = useQuery([InputPointerComponent])
-    return pointers.filter((entity) => getComponent(entity, InputPointerComponent).cameraEntity === cameraEntity)
+    return pointers.filter((entity) => getComponent(entity, InputPointerComponent)?.cameraEntity === cameraEntity)
```
HexaField, line 104: how is this returning entities without the input pointer component? the nullish coalescing here should be redundant

HexaField (dismissed):

packages/spatial/src/input/components/InputPointerComponent.ts
SYBIOTE: There is a race condition case here, if I put this in a reactor, and the reactor triggers a lot of times, I noticed that in some cases, the entity is created for the input poiter emulator but this is called before the component is added, this caused the reactor to fall and the gizmo to disappear

Adding this covers the edge case

HexaField: hmm ok

dinomut1 (approved):

PR #949, opened 2025-01-16
update node 22

HexaField:
## References
closes #_insert number here_

### DIFF
--- a/.github/workflows/branch-build.yml
+++ b/.github/workflows/branch-build.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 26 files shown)


DanielBelmes:
I have ran latest node version locally for about a year so I feel like I've tested it enough. But usually it's the node servers we gotta worry about. If it's not up on QAT already I will likely deploy to it for testing. EDIT: Nice you already deployed it

barankyle (approved):
Looks good to me from some light testing on mt-qat. Since it built properly, backend servers don't crash, I can log in/out, update a scene with some simple geometry, and go to a location and connect to world/media servers, I think it's still all working.

PR #950, opened 2025-01-16
feat(IR-4495): Add lock icon to hierachy panel to lock transformations

kattsushi:
## Summary

https://github.com/user-attachments/assets/2c9e129a-06d7-4306-8bb1-3fc498972ba1

## References
Jira ticket related: [IR-4495](https://tsu.atlassian.net/issues/IR-4495?filter=-1)

## QA Steps

[IR-4495]: https://tsu.atlassian.net/browse/IR-4495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HexaField (changes requested):
we definitely do not want a component for this. components are for scene data and interactions in the 3d world. since this "locking" functionality is limited to the studio, we just need a hyperflux state that maps each EntityUUID to a boolean, and prevent the user from mutating properties for any entities that are set to true in this new state.

kattsushi:
> we definitely do not want a component for this. components are for scene data and interactions in the 3d world. since this "locking" functionality is limited to the studio, we just need a hyperflux state that maps each EntityUUID to a boolean, and prevent the user from mutating properties for any entities that are set to true in this new state.

yup make sense now to me, working on it

packages/editor/src/services/LockedState.ts
```diff
@@ -22,19 +22,31 @@ Original Code is the Infinite Reality Engine team.
...
-  name: 'LockedComponent',
-  jsonID: 'EE_locked',
-  schema: S.Bool(false)
+import { Entity } from '@ir-engine/ecs'
+import { defineState, getMutableState } from '@ir-engine/hyperflux'
+
+export const LockedState = defineState({
+  name: 'LockedState',
+  initial: () => ({
+    lockedEntities: new Map<Entity, boolean>() // Map to store locked state of entities
```
HexaField: this should be a record/object, otherwise it will not be reactive. Maps and Sets are not reactive in Hookstate.

packages/editor/src/services/LockedState.ts
```diff
@@ -22,19 +22,31 @@ Original Code is the Infinite Reality Engine team.
...
-import { S } from '@ir-engine/ecs/src/schemas/JSONSchemas'
-
-export const LockedComponent = defineComponent({
-  name: 'LockedComponent',
-  jsonID: 'EE_locked',
-  schema: S.Bool(false)
+import { Entity } from '@ir-engine/ecs'
+import { defineState, getMutableState } from '@ir-engine/hyperflux'
+
+export const LockedState = defineState({
```
HexaField: Let's call it `EntityHierarchyLockState` for clarity, and the `name` field should be namespaced as `"ir.editor.EntityHierarchyLockState"`

HexaField (changes requested):

packages/editor/src/services/LockedState.ts
kattsushi: sure good catch, on it

packages/editor/src/services/EntityHierarchyLockState.ts
```diff
@@ -25,28 +25,29 @@ Infinite Reality Engine. All Rights Reserved.
...
+    lockedEntities: {} as Record<Entity, boolean> // Map to store locked state of entities
   }),
   // Updates the locked state of a specific entity
   updateLocked: (entityId: Entity, isLocked: boolean) => {
-    const state = getMutableState(LockedState)
-    const updatedMap = new Map(state.lockedEntities.value) // Create a new Map to trigger reactivity
-    updatedMap.set(entityId, isLocked)
-    state.lockedEntities.set(updatedMap) // Replace the Map entirely
+    const state = getMutableState(EntityHierarchyLockState)
+    state.lockedEntities.set({
```
HexaField: the proper API for hookstate here is `state.lockedEntities[entityId].set(isLocked)` which will not force rerenders for all hooked entities, just the ones listening for the specified one

packages/editor/src/services/EntityHierarchyLockState.ts
```diff
@@ -25,28 +25,29 @@ Infinite Reality Engine. All Rights Reserved.
...
+      ...state.lockedEntities.get(),
+      [entityId]: isLocked
+    }) // Replace the Map entirely
   },
 
   // Retrieves the lock status of a specific entity
   isEntityLocked: (entityId: Entity): boolean => {
-    const state = getMutableState(LockedState)
-    return state.lockedEntities.get()?.get(entityId) ?? false // Default to false if not set
+    const state = getMutableState(EntityHierarchyLockState)
```
HexaField: if this doesn't need to be reactive, we can use `getState` here

HexaField (dismissed):

packages/editor/src/services/EntityHierarchyLockState.ts
kattsushi: yup ty getting familiar with that api

DanielBelmes (approved):
Looks great. I was considering the behavior when a child is locked but honestly the cost of querying that is gonna be annoying for a simple function. A lock that stops the gizmo from working when you select object.

PR #960, opened 2025-01-18
Envmap Refactor

AidanCaruso:
## Summary
Refactors envmaps to set data directly on material state entity children, instead of doing this through each child mesh's material referenced through threejs' APIs. Aimed at simplifying things and avoiding a ton of race conditions.
image

## References
closes #_insert number here_


HexaField (approved):
nice job!

PR #965, opened 2025-01-21
Add min and mag filters for mipmaps in skybox rendering and force minFilter for equirectangler skybox

SYBIOTE:
Introduce min and mag filters for mipmaps to ensure proper functionality of the skybox in Three.js.

A Skybox won't be able to utilize mipmaps and leaving them enabled breaks rendering.


HexaField (approved):

HexaField (changes requested):
can we also update the default scene to use the ktx2 compressed texture

SYBIOTE:
> can we also update the default scene to use the ktx2 compressed texture

Daniel updated it , points to ktx2, we should be able to merge this now

HexaField (approved):

PR #966, opened 2025-01-22
Improve Highlight Effect performance

DanielBelmes:
Will likely still drop on selection until we can find a way to improve further.

## References
closes #_insert number here_


packages/editor/src/systems/HighlightSystem.ts
```diff
@@ -23,13 +23,25 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
+import { MeshComponent } from '@ir-engine/spatial/src/renderer/components/MeshComponent'
+import { ObjectComponent } from '@ir-engine/spatial/src/renderer/components/ObjectComponent'
+import { VisibleComponent } from '@ir-engine/spatial/src/renderer/components/VisibleComponent'
 import { useEffect } from 'react'
 import { SelectionState } from '../services/SelectionServices'
 
+function highlightCallback(func) {
+  return (child, index) => {
+    if (!hasComponent(child, MeshComponent)) return
+    if (!hasComponent(child, ObjectComponent)) return
```
HexaField: object component check is redundant, since all mesh components will have an object component

packages/editor/src/systems/HighlightSystem.ts
```diff
@@ -23,13 +23,25 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
 import { defineSystem } from '@ir-engine/ecs/src/SystemFunctions'
 import { AnimationSystemGroup } from '@ir-engine/ecs/src/SystemGroups'
 import { HighlightComponent } from '@ir-engine/spatial/src/renderer/components/HighlightComponent'
+import { MeshComponent } from '@ir-engine/spatial/src/renderer/components/MeshComponent'
+import { ObjectComponent } from '@ir-engine/spatial/src/renderer/components/ObjectComponent'
+import { VisibleComponent } from '@ir-engine/spatial/src/renderer/components/VisibleComponent'
 import { useEffect } from 'react'
 import { SelectionState } from '../services/SelectionServices'
 
+function highlightCallback(func) {
```
HexaField: let's add types to the args

packages/spatial/src/renderer/components/HighlightComponent.ts
```diff
@@ -26,34 +26,23 @@ Infinite Reality Engine. All Rights Reserved.
...
-      if (!hasComponent(child, MeshComponent)) return
-      if (!hasComponent(child, ObjectComponent)) return
-      if (!hasComponent(child, VisibleComponent)) return
-      highlightObjects.add(getComponent(child, MeshComponent))
-    })
-  }
   const rendererComponent = getComponent(Engine.instance.viewerEntity, RendererComponent)
   const outlineEffect = rendererComponent?.effectInstances?.OutlineEffect as OutlineEffect
-  outlineEffect?.selection.set(highlightObjects)
+  outlineEffect?.selection.set(highlightQuery().map((entity) => getComponent(entity, ObjectComponent)))
```
HexaField: we can move the arrow function to module scope

HexaField (changes requested):

DanielBelmes:
dang never pressed git push. PR review updates are in @HexaField

SYBIOTE (approved):

HexaField (approved):
@DanielBelmes please re-review the additional changes I made. There is an edge case possible where a highlight component is added to a child that is a visible mesh, then loses the mesh or visible component, and then upon unselection, that mesh would not have the highlight component removed. calling removeComponent on an entity that does not have it is a no-op, so we are safe to call it on all children.

PR #969, opened 2025-01-22
Use ternary operator to make getOptional faster

DanielBelmes:
## References
closes #_insert number here_

### DIFF
--- a/packages/ecs/src/ComponentFunctions.ts
+++ b/packages/ecs/src/ComponentFunctions.ts
@@ -461,9 +461,7 @@ export const getOptionalMutableComponent = <C extends Component>(
   entity: Entity,
   component: C
 ): State<ComponentType<C>> | undefined => {
-  return !bitECS.hasComponent(HyperFlux.store, entity, component)
-    ? undefined
-    : (component.stateMap[entity]! as State<ComponentType<C>> | undefined)
+  return component?.stateMap?.[entity]
 }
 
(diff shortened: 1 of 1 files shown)


packages/ecs/src/ComponentFunctions.ts
```diff
@@ -461,9 +461,7 @@ export const getOptionalMutableComponent = <C extends Component>(
   entity: Entity,
   component: C
 ): State<ComponentType<C>> | undefined => {
-  return !bitECS.hasComponent(HyperFlux.store, entity, component)
-    ? undefined
-    : (component.stateMap[entity]! as State<ComponentType<C>> | undefined)
+  return component?.stateMap?.[entity]
```
DanielBelmes, line 464: So we previously had something here different from getOptionalComponent where we are forcing it to be a value. This is causing tests to fail. Might be causing something at runtime to fail as well. Thoughts? It doesn't match with the return type we had to begin with.

the-reality-engineer (commented):
So `hasComponent` check will still be necessary for the `getOptionalMutableComponent`, because the state will be preemptively created w/ `none` value in useComponent and useOptionalComponent (this is the behavior we've had for a while, and will be restored in the authoring branch). In other words, the state existing does not necessarily indicate that the component is added to an entity, or that it is safe to access. 

The suggested change *should* be safe _only_ for `getOptionalComponent`. I can make the change in the authoring branch, where other changes to ComponentFunctions are in progress.

DanielBelmes:
@speigg yes please! After you see this comment just close the PR.

PR #977, opened 2025-01-23
Fixed a bug in migrating aws-settings

barankyle:
## Summary

The S3 setting `endpoint` is normally not set at all in a production environment. But the migration to engine-settings for this value was setting it to an empty string. This was resulting in S3 trying to point to the endpoint <empty string>, which would fail on all commands.

Default is now to set this to undefined, both on migrating existing aws-settings tables and new seeds.

Also made S3 client endpoint require that s3.endpoint be a non-zero-length string in order to use it, otherwise it defaults to undefined, to make sure any existing errant migrations won't cause this problem.

Resolves IR-6663 IR-6665 IR-6669 IR-6673

## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/media/storageprovider/s3.storage.ts
+++ b/packages/server-core/src/media/storageprovider/s3.storage.ts
@@ -139,9 +139,7 @@ export class S3Provider implements StorageProviderInterface {
         profile: config.aws.s3.roleArn ? 'role' : 'default',
         filepath: credentialsPath
       }),
-      endpoint: config.server.storageProviderExternalEndpoint
-        ? config.server.storageProviderExternalEndpoint
-        : config.aws.s3.endpoint,
+      endpoint: config.server.storageProviderExternalEndpoint || config.aws.s3.endpoint || undefined,
       region: config.aws.s3.region,
       forcePathStyle: true,
       maxAttempts: 5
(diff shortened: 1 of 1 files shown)


packages/server-core/src/media/storageprovider/s3.storage.ts
```diff
@@ -139,9 +139,12 @@ export class S3Provider implements StorageProviderInterface {
         profile: config.aws.s3.roleArn ? 'role' : 'default',
         filepath: credentialsPath
       }),
-      endpoint: config.server.storageProviderExternalEndpoint
-        ? config.server.storageProviderExternalEndpoint
-        : config.aws.s3.endpoint,
+      endpoint:
+        config.server.storageProviderExternalEndpoint?.length > 0
```
HexaField: can we use `typeof` instead of `.length` ?

HexaField: oh, it defaults to an empty string right. we can just use an equality check, the `?.length > 0` is redundant

barankyle: Check that it's `!== ''`, or something else?

barankyle: Nevermind, I think I see what you're getting at, we can do externalEndpoint || s3.endpoint || undefined and it'll treat empty strings or null/undefined as falsey and then default to undefined.

barankyle:
I ended up deciding to leave engine-settings' value column as a string. I just tweaked the logic around setting the S3 client's endpoint so that an empty string there is evaluated as a falsey value and undefined is used instead.

For minIO I think we have to provide an endpoint, and that has perfectly fine defaults in .env.local.default, so I think it's acceptable to not default to undefined there.

HexaField (approved):

PR #980, opened 2025-01-23
iR-6544, 6545 - fix dropdown menu from getting clipped

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-6544
https://tsu.atlassian.net/browse/IR-6545

## References
closes #_insert number here_


CITIZENDOT (changes requested):
https://github.com/user-attachments/assets/e63129ae-bc8c-4235-9a17-b37d3fd94057

- Unable to select options
- Dropdown do not seem to appear when clicked second time.

On a side note, To fix the the linked tickets, we can set a reasonable max height on those specific drop-downs I guess? So all the options would be scrollable within that height.

Example: https://github.com/ir-engine/ir-engine/blob/7cd626d9e699dcf0363ef21e6eff4e79330a513b/packages/client-core/src/admin/components/project/AddEditProjectModal.tsx#L477-L486

achen5671:
> 2025-01-24.11-54-50.mp4 
> * Unable to select options
> * Dropdown do not seem to appear when clicked second time.
> 
> On a side note, To fix the the linked tickets, we can set a reasonable max height on those specific drop-downs I guess? So all the options would be scrollable within that height.
> 
> Example:
> 
> https://github.com/ir-engine/ir-engine/blob/7cd626d9e699dcf0363ef21e6eff4e79330a513b/packages/client-core/src/admin/components/project/AddEditProjectModal.tsx#L477-L486

we could do that as a quick fix, but I think it still be worth to address the main issue of the drop downs getting clipped due to parent containers.

On hindsight, it will be good to see what other platforms do

achen5671:
@CITIZENDOT I thought about it and I think Im going with your approach. it's much more simple for now to have a fix height. There should be design discussions if we want to have dynamic height.

Im closing this pr in favor of https://github.com/ir-engine/ir-engine/pull/988

CITIZENDOT:
> On hindsight, it will be good to see what other platforms do

Material UI sets max height on the parent component: https://mui.com/material-ui/react-select/#placeholder

image

CITIZENDOT:
We might have to bring this back, Dropdowns in `/admin/settings` are clipped by other Accordions. It isn't fixable with `maxHeight` property. I'll figure something out.

CITIZENDOT:
Here: https://github.com/ir-engine/ir-engine/pull/1055

PR #982, opened 2025-01-24
rerun selective engine tests twice

aditya-mitra:
## Summary

Fix the flaky tests inside `GLTFLoader.test.tsx` and `MaterialLibrarySystem.test.tsx` which are failing inside GitHub actions.

### DIFF
--- a/packages/engine/src/gltf/GLTFLoader.test.tsx
+++ b/packages/engine/src/gltf/GLTFLoader.test.tsx
@@ -92,7 +92,7 @@ const setupEntity = () => {
   return entity
 }
 
-describe('GLTF Loader', async () => {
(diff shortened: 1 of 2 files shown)


packages/engine/src/gltf/GLTFLoader.test.tsx
```diff
@@ -92,7 +92,7 @@ const setupEntity = () => {
   return entity
 }
```
heysokam, line 94: Remember to add a `@todo` note as a reminder for whoever refactors these tests in the future.
Something like:
`/** @todo Retry is undesirable. Remove the option after the code that fails has been refactored */`

HexaField (approved):

PR #985, opened 2025-01-24
Updates to support GCP

shayansadeghieh:
## References
closes #_insert number here_


barankyle (approved):

infinitelos (approved):

dockerfiles/builder/Dockerfile-builder
```diff
@@ -17,16 +17,30 @@ CMD ["scripts/strip-engine.sh"]
 
 # copy then compile the code
 
-ENV APP_ENV=production
+ENV APP_ENV=development
```
HexaField, line 20: curious about this change, what does this affect?

barankyle: I'm not sure it affects anything, I think APP_ENV is hard-set to production with most of the tasks that are run.

PR #987, opened 2025-01-25
Ensure loaded URLs are parsed correctly

SYBIOTE:
Refactor URL parsing logic to improve clarity and maintainability. Update relevant functions to utilize the new parsing methods, ensuring that URLs are transformed consistently across the application.

why we need this:
when we publishing we save changes to scene, which causes the URLs in the gltf file to be cleaned, when we duplicate or delete thereafter, we end up using the cleaned urls instead of the parsed ones, 
putting a parser in the loading process for the file loader and the use Loaders, mitigates the issue


DanielBelmes:
Smoke testing of specific issue is done. Do not see bug anymore

DanielBelmes (approved):

PR #990, opened 2025-01-28
Increase `z-index` of Modal background

CITIZENDOT:
## Summary

RC-Dock's `z-index` is 250, and the Modal background's `z-index` was 50. Hence, it wasn't visible currently. Bumping modal's z-index to 1000 will make it standout in almost all pages.

## References
closes https://tsu.atlassian.net/browse/IR-6764

### DIFF
--- a/packages/ui/src/primitives/tailwind/ClickawayListener/index.tsx
+++ b/packages/ui/src/primitives/tailwind/ClickawayListener/index.tsx
@@ -33,7 +33,7 @@ const ClickawayListener = (props: { children: JSX.Element }) => {
   const childOver = useHookstate(false)
   return (
     <div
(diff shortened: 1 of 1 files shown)


aditya-mitra (approved):
this works fine.

i think we now need to have z index priority so that when others are using z-index they just use that priority key. this could be a custom tailwind config value such as z-higest, z-higher, z-high, z-moderate, etc.

CITIZENDOT:
> this works fine.
> 
> i think we now need to have z index priority so that when others are using z-index they just use that priority key. this could be a custom tailwind config value such as z-higest, z-higher, z-high, z-moderate, etc.

Will add this during the colors refactor.

PR #991, opened 2025-01-28
[IR-6563] Unit Tests for ECS Layers

heysokam:
## References
https://tsu.atlassian.net/browse/IR-6563


packages/ecs/src/Layers.test.tsx
```diff
@@ -0,0 +1,797 @@
...
+/**
+ * @note
+ * Other related code that also has ECSLayers specific tests:
+ * - EntityFunctions.test.tsx : createEntity removeEntity
+ * - UUIDComponent.test.tsx   : Almost every function
+ * */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { TransformComponent } from '@ir-engine/spatial/src/transform/components/TransformComponent'
```
HexaField: we should not import files from packages that this package does not consume. rather than the TransformComponent we can define our own component in this file

HexaField (changes requested):

packages/ecs/src/UUIDComponent.ts
```diff
@@ -65,31 +65,31 @@ export const UUIDComponent = defineComponent({
...
     // set new uuid
-    _getUUIDState(uuid, layer).set(entity)
+    UUIDComponentFunctions._getUUIDState(uuid, layer).set(entity)
   },
 
   onRemove: (entity, component) => {
     const uuid = component.value
     const layer = LayerComponent.get(entity)
-    _getUUIDState(uuid, layer).set(UndefinedEntity)
+    UUIDComponentFunctions._getUUIDState(uuid, layer).set(UndefinedEntity)
```
HexaField, line 78: accessing via UUIDComponentFunctions will incur a small performance loss. Inside this file, we don't need to access it via that namespace

HexaField (changes requested):

packages/ecs/src/Layers.test.tsx
heysokam: Solved in #ce91ca1

HexaField (approved):

PR #995, opened 2025-01-29
IR-6682: Update project-based permissions in console

hanzlamateen:
## References
https://tsu.atlassian.net/browse/IR-6682


mikeplascdev (approved):
LGTM

packages/server-core/src/media/file-browser/file-browser.test.ts
```diff
@@ -216,85 +216,85 @@ describe('file-browser.test', () => {
...
-            .service(fileBrowserPath)
-            .create('projects/' + testProject2Name + '/../../' + testProject1Name + '/public/test', {
-              user: user2
-            }),
-        {
-          message: 'Missing required project permission for ' + project1.name
-        }
-      )
-    })
+    // it('will not create a directory in project 1 as user 2, who has only editor permission', async () => {
```
HexaField, line 219: are these tests no longer valid?

packages/server-core/src/hooks/set-in-context.ts
```diff
@@ -31,9 +31,16 @@ import { HookContext } from '../../declarations'
  * following setField hook.
  * https://hooks-common.feathersjs.com/hooks.html#setfield
  */
-export default (propertyName: string, propertyValue: string, inData?: false) => {
+
+export enum ContextScope {
```
HexaField, line 35: this abstraction is a little worrying, what exactly is this doing?

HexaField (dismissed):

packages/server-core/src/hooks/set-in-context.ts
hanzlamateen: So, `setInContext` hook is required to to manipulate data in context.params.query, context.data or root context object. This enum allow to easily set that data from a hook object without writing custom code.

i.e. `setInContext('type', 'public'))`  will set context.params.query.type = 'public'

`setInContext('type', 'public', ContextScope.Data))`  will set context.data.type = 'public'

packages/server-core/src/media/file-browser/file-browser.test.ts
hanzlamateen: Nope, because now editor also has these various permissions which they previously don't. Do get in touch with Chris Ball if you have concerns giving editor access to these.

barankyle: Shouldn't we just remove all of these tests that were formerly checking reviewer access?

hanzlamateen: I commented them out if at a later point in time we need to introduce a permission with limited permission, then maybe these tests can be leveraged.

Though I can delete if desired.

barankyle (approved):

PR #998, opened 2025-01-29
IR-6619: fix re-renders and performance issues in the react side on loading view component for loading engine viewer

kattsushi:
## Summary
# Description
Optimized loading performance and reduced unnecessary renders in the engine loading process

## Key Changes
- Removed redundant LoadingView component instances throughout the application for the Location (Viewer loading)
- Implemented memoization techniques for location-related hooks to prevent re-renders
- Added conditional rendering for LoadingView (animated) based on route path
- Optimized engine loading sequence by consolidating loading states

## Technical Details
- Used React.memo() for components handling location services and content
- Added route-based loading optimization for /location paths
- I just use useRef for tracking and debugging render cycles
- Consolidated loading states to prevent cascading re-renders

## Performance Improvements
- Reduced initial loading time by removing duplicate loading screens
- Decreased number of re-renders during engine initialization ( LoadingView was rendering around 20 times before xe )
- Optimized memory usage by preventing unnecessary component mounts

## Testing
- Verified loading sequence in development environment
- Confirmed proper engine initialization
- Tested route transitions and loading states

## References
Pull requests reference [#712](https://github.com/theinfinitereality/irpro-multitenancy/pull/712)


packages/client-core/src/components/LoadWebappInjection.tsx
```diff
@@ -32,35 +32,38 @@ import LoadingView from '@ir-engine/ui/src/primitives/tailwind/LoadingView'
...
-
   const projectComponents = useHookstate(null as null | any[])
   const projects = useFind(projectsPath)
 
   useEffect(() => {
     if (!projects.data.length) return
     loadWebappInjection(projects.data as string[])
       .then((result) => {
-        projectComponents.set(result)
+        projectComponents.set(result || []) // Ensure we set an empty array if result is null/undefined
```
HexaField: FYI, projects should NEVER return empty, since projects define routes and the default project should always be present. so this is redundant

packages/client-core/src/components/LoadWebappInjection.tsx
```diff
@@ -32,35 +32,38 @@ import LoadingView from '@ir-engine/ui/src/primitives/tailwind/LoadingView'
...
       })
       .catch((e) => {
         console.error(`Failed to import webapp load event for project ${projects.data} with reason ${e}`)
         projectComponents.set([])
       })
   }, [projects.data])
 
-  if (!projectComponents.value) {
+  // Skip rendering if login is required and components aren't loaded yet
+  if (props.loginRequired && !Array.isArray(projectComponents.value)) {
```
HexaField: what is the condition in which Array.isArray is different to a nullish check? It should always either be an array or null

packages/client-core/src/components/LoadWebappInjection.tsx
```diff
@@ -32,35 +32,38 @@ import LoadingView from '@ir-engine/ui/src/primitives/tailwind/LoadingView'
 import React, { Suspense, useEffect } from 'react'
 import { useTranslation } from 'react-i18next'
 
-export const LoadWebappInjection = (props: { children: React.ReactNode; fallback?: JSX.Element }) => {
+export const LoadWebappInjection = (props: {
+  children: React.ReactNode
+  fallback?: JSX.Element
+  loginRequired?: boolean
```
HexaField: utilizing non-authentication makes no sense here, since the projects API service requires authentication

packages/client-core/src/world/Location.tsx
```diff
@@ -23,79 +23,57 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
   online?: boolean
 }
 
-const LocationPage = ({ online }: Props) => {
-  const { t } = useTranslation()
-  const params = useParams()
-  const ready = useMutableState(LoadingUISystemState).ready
-
-  useNetwork({ online })
+const LocationServices = memo(() => {
```
HexaField: why memo? these hooks should only be used once across the entire webpage, and are just wrapping a useEffect

packages/client/src/pages/location/location.tsx
```diff
@@ -24,38 +24,45 @@ Infinite Reality Engine. All Rights Reserved.
...
 
+import { LoadingUISystemState } from '@ir-engine/client-core/src/systems/LoadingUISystem'
+import { useMutableState } from '@ir-engine/hyperflux'
 import '../mui.styles.scss' /** @todo Remove when MUI is removed */
 import '../styles.scss'
 
 const LocationRoutes = () => {
   const ref = useRef<HTMLElement>(document.body)
+  const loadingState = useMutableState(LoadingUISystemState)
+  const ready = useMemo(() => loadingState.ready.value, [loadingState.ready.value])
```
HexaField: useMemo does nothing here. hookstate eliminates the need for react's hooks. these are functionally equivalent, aside from the extra garbage react hooks create

packages/client/src/store.tsx
```diff
@@ -62,6 +62,7 @@ getMutableState(DomainConfigState).merge({
 
 export default function ({ children }): JSX.Element {
   const { t } = useTranslation()
+  const isLocation = window.location.pathname.includes('/location')
```
HexaField, line 65: let's pass this in rather than asume it from the pathname. we may want this functionality in other routes in the future

HexaField (dismissed):

packages/client/src/pages/location/location.tsx
kattsushi: Yeah, it’s probably an issue more related to how Hookstate works internally (but I think it’s because the ready state inside the system is mutated a couple of times). This is causing unnecessary re-renders:

With useMemo:
Screenshot 2025-01-29 at 6 01 38 p m

Without useMemo:
Screenshot 2025-01-29 at 6 00 09 p m


Is there a way to avoid this with some approach for Hookstate instead of using React hooks?

HexaField: re-renders shouldn't be a problem. if the data doesn't change the effects shouldn't change. What problem is re-renders causing?

packages/client-core/src/world/Location.tsx
kattsushi: something similar to my comment bellow, is the way that i found that avoid re-renders on LocationPage

without memo()
Screenshot 2025-01-29 at 6 25 23 p m

with memo()
Screenshot 2025-01-29 at 6 22 17 p m

packages/client-core/src/components/LoadWebappInjection.tsx
kattsushi: This issue is more related to whether the current page is a location page or not when AppPage is implemented. Currently, the logic is:
`<AppPage loginRequired={!isLocationPath}>`

However, I can change the prop name to isLocationPage to avoid rendering the fallback loading by default. This is because, in that specific flow, we already have a LoadingView with an animation. By doing this, we can fix the flickering loading state that appears in the middle when the location is still loading.

packages/client-core/src/components/LoadWebappInjection.tsx
kattsushi: gocha

packages/client-core/src/components/LoadWebappInjection.tsx
kattsushi: you're right

packages/client/src/pages/location/location.tsx
kattsushi: Maybe before it wasn’t as noticeable because the `LoadingView` only consisted of a spinning spinner. However, in this new use case, with an `SVG logo animation,` when a re-render happens in the middle of the animation, it becomes much more noticeable, and the experience in many cases is not ideal. That said, most of the performance issues were already fixed by removing all the deeply nested instances of the same `LoadingView` component for the Location loading case and while debugging, I noticed those re-renders in important components like Location. If you think re-renders aren’t as critical in this case, I can revert them back to how they were 👍🏾

HexaField: Hmm, the component itself should not trigger the SVG to restart if it isn't being affected by the state. Might be a problem with how it's being interpreted by vite or react.

DanielBelmes:
Things feel significantly better. There was definitely improvements to load. Though we definitely still have a lot of loading screens. Pretty much runs through them all. I switched the animation to WebP as the SVG was really struggling with resources.  From some research that is the common drawback to svg animation since it needs to be computed. So for less quality we're just gonna go with WebP. It looks good to me.

DanielBelmes (approved):

DanielBelmes:
Couldn't figure out how to delay the loading screen 2500

PR #1000, opened 2025-01-30
IR-7310 Fix model transform upload regex to allow things in 'public' folder

dinomut1:
Hotfix for sports template wizard demo


packages/common/src/model/ModelTransformFunctions.ts
```diff
@@ -404,7 +404,7 @@ const doUpload = async (projectName, fileName, buffer) => {
 }
 
 const toProjectAndFileName = (fUploadPath: string, srcBaseURL: string): [string, string] => {
-  const pathCheck = /projects\/([^/]+\/[^/]+)\/assets\/([\w\d\s\-|_./]*)$/
+  const pathCheck = /projects\/([^/]+\/[^/]+)\/(?:assets\/)?([\w\d\s\-|_./]*)$/
```
DanielBelmes: I believe if we're just trying to allow for uploading to public we should modify the regex to 
`projects\/([^/]+\/[^/]+)\/(?:assets|public)\/([\w\d\s\-|_./]*)$`

We don't super have to mark it as a non capturing group 

Below is a table of test cases with if they match with the provided regex.
Original:
projects/projects/TheProject/assets/ Y
projects/projects/TheProject/public/ X
projects/projects/TheProject/blah/    X

Current:
projects/projects/TheProject/assets/ Y
projects/projects/TheProject/public/ Y
projects/projects/TheProject/blah/    Y

Suggested
projects/projects/TheProject/assets/ Y
projects/projects/TheProject/public/ Y
projects/projects/TheProject/blah/    X

DanielBelmes (dismissed):

packages/common/src/model/ModelTransformFunctions.ts
DanielBelmes: Also noting that the or(|) modifier should just either or both strings but when testing I had to wrap in a capture group to act correctly.

HexaField (dismissed):
we must have comprehensive position and negative test case coverage for regexes. please prove this change by writing appropriate tests

SamMazerIR:
added more test coverage, let me know if i'm missing anything

DanielBelmes (approved):

PR #1001, opened 2025-01-30
[iR-6609] iframe ECS

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-6609
## References
closes #_insert number here_


packages/engine/src/scene/components/IFrameComponent.ts
```diff
@@ -0,0 +1,86 @@
...
+      return
+    }, [iframeComponent.src])
+
+    useEffect(() => {
+      setCallback(entity, iframeCallbackName, () => toggleOpen(entity))
+
+      return () => {
+        removeCallback(entity, iframeCallbackName)
+      }
+    }, [entity])
```
SamMazerIR: I'm not sure if this dependency is doing anything, could it be empty?

packages/engine/src/scene/components/IFrameComponent.ts
```diff
@@ -0,0 +1,86 @@
...
+
+import { Entity } from '@ir-engine/ecs/src/Entity'
+import { S } from '@ir-engine/ecs/src/schemas/JSONSchemas'
+import { addError, clearErrors } from '../functions/ErrorFunctions'
+
+const interactMessage = 'Click'
+const iframeCallbackName = 'iframeCallback'
+
+const toggleOpen = (iframeEntity: Entity) => {
+  const iframeComponent = getComponent(iframeEntity, IFrameComponent)
```
SamMazerIR, line 41: this. may never be an issue with how it is currently used, but I'd use getOptionalComponent and check for null. - probably paranoia/personal preference

achen5671: I tried to use getOptionalComponent and it wouldnt work. Correct me if Im wrong but wouldnt getOptionalComponent only get on initial render?

SamMazerIR: getOptionalComponent returns nullable, so you have to check if it is null or not. getComponent shows a warning that will eventually be an error if this happens, and you'll just unexpectedly have an undefined variable iirc. What happened when you tried optional?

SamMazerIR (approved):

PR #1003, opened 2025-01-31
fix(ui-button): remove padding when is icon only [IR-6680]

mikeplascdev:
## Summary
- Remove padding when only display the icon

## References
closes #_insert number here_

## QA Steps
- Go to [iR Studio](https://ir-engine-mt-dev.theinfinitereality.io/welcome?priceId=price_1Q4c7u2LAvefLedZZBFlhdte) 
- Proceed with join process
- Complete join process
- Go to /dashboard page after joining

## Evidence
Screenshot 2025-01-31 at 2 48 57 p m

## Jira
[IR-6680](https://tsu.atlassian.net/browse/IR-6680)

[IR-6680]: https://tsu.atlassian.net/browse/IR-6680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aditya-mitra (changes requested):
The button is not rounded when there is padding because the [base figma designs](https://www.figma.com/design/ln2VDACenFEkjVeHkowxyi/iR-Engine-Design-Library-File?node-id=2130-9827&m=dev) are rectangular in an icon only state. 

image

We are adhering the primitives to figma design library.
The alternative solution can be to use a native `<button>` with rounded corners since it is one-off case.

mikeplascdev:
@aditya-mitra 

The problem is this `px-4` if we only add an icon without text, the padding is added and makes the icon smaller, regardless of the size.

https://github.com/ir-engine/ir-engine/blob/e7893e451d2ba4e3402d2f51543a2746f9c45a52/packages/ui/src/primitives/tailwind/Button/index.tsx#L62

## Evidence
https://github.com/user-attachments/assets/f82db489-3923-4c9b-8bc8-5f84c8078bfe

aditya-mitra (commented):
Had a closer look on the [figma designs](https://www.figma.com/design/a3qQCRxNWxP7K1titmso0e/Console-Design?node-id=5684-33139&m=dev)

Screenshot 2025-02-07 at 9 15 07 AM


It seems that the plus (add) button in the center is a tertiary button primitive with added border radius. Can you add this configuration in that "add scenes" button and see if it is working? 

Ideally, this button should be as per the figma design has configured and we should be able to achieve it by adding the necessary props.

mikeplascdev:
fixed here @aditya-mitra

PR #997, opened 2025-01-29
Colors Refactor

CITIZENDOT:
## Summary

Update color scheme to follow the new designs & variables.

Figma: https://www.figma.com/design/ln2VDACenFEkjVeHkowxyi/iR-Engine-Design-Library-File?node-id=8163-18744&t=a9KSvvL7PoFE1Dg7-4

## References
closes https://tsu.atlassian.net/browse/IR-6777

## QA Steps

|                                     Light Mode Studio                                     |                                     Dark Mode Studio                                      |
| :---------------------------------------------------------------------------------------: | :---------------------------------------------------------------------------------------: |
| image | image |

|                                   Light Mode Dashboard                                    |                                    Dark Mode Dashboard                                    |
| :---------------------------------------------------------------------------------------: | :---------------------------------------------------------------------------------------: |
| image | image |

PR #1007, opened 2025-02-04
TALK ABOUT: the sliders we show in our components shouldn't be resized to the exact width of their parent

Rezmason:
This commit sizes sliders in panels to be 3/5 the reported current width of their parent, rounded down to the nearest pixel. Even this is a bad solution to the problem of sizing sliders; their width contributes to their parents' reported width, which can cause runaway reactive loops that expand slider widths indefinitely.

We need a better solution— probably something CSS based.

### DIFF
--- a/packages/ui/src/components/editor/Slider/index.tsx
+++ b/packages/ui/src/components/editor/Slider/index.tsx
@@ -81,7 +81,7 @@ const Slider = ({
 
   useEffect(() => {
     const observer = new ResizeObserver(() => {
-      if (parentRef.current) setWidth(parentRef.current?.offsetWidth)
+      if (parentRef.current) setWidth(Math.floor((parentRef.current?.offsetWidth * 3) / 5))
     })
     observer.observe(parentRef.current as Element)
     return () => observer.disconnect()


aditya-mitra:
Sorry, am unable to get the 3/5th part mentioned above. How about we add a maximum width if we want to prevent the slider for expanding?

CITIZENDOT:
I don't understand why we're manually setting width to the parent's width instead of setting `w-full` class. I've updated this in the Colors Refactor: https://github.com/ir-engine/ir-engine/pull/997 PR.

Rezmason:
> I don't understand why we're manually setting width to the parent's width instead of setting `w-full` class. I've updated this in the Colors Refactor: #997 PR.

I'll take a look at that tomorrow. 👍 Thanks, Citizen Dot!™ 🦸

@aditya-mitra the `floor(3/5 * width)` setting is haphazard, inspired by the proportion of the existing UI components in the node editors— but this PR is only meant to prompt this discussion, not to merge the change in as-is. That's why I put "TALK ABOUT".

If #997 fixes this specific problem in the `irpro-wizard` UI, then This PR gets closed-with-comment. Otherwise we should evolve this PR into something actually mergable

Rezmason:
image

PR #1008, opened 2025-02-04
Minor style change to improve the legibility of the Create Scene dialog

Rezmason:
—including when it is extended by other projects through UI injection.

supports [#IR-5557](https://tsu.atlassian.net/browse/IR-5557)


aditya-mitra (approved):
Looks good!
In next PRs, if you add a before and after image of the what the changes do, it can help me get the context quicker.

PR #1009, opened 2025-02-04
IR-6884: Changes to split out analytics events from general logging

hanzlamateen:
## Summary

This PR introduced a new section in client logger to handle anayltics events. These events are then sent to server where it logs them to BigQuery.

Following is the code snippet used to log anayltics event on client side:
```
// Import logger
import multiLogger from '@ir-engine/common/src/logger'

// Initialize logger
const logger = multiLogger.child({ component: 'client-core:{FILE_NAME}', modifier: clientContextParams })

// Send analytics log event to server
logger.analytics({ event_name: 'view_fullscreen', event_value: true })
```

This PR required following ENVs to be set:

```
BQ_PROJECT_ID
BQ_DATASET_ID
BQ_TABLE_ID
```

Also merge: https://github.com/theinfinitereality/irpro-multitenancy/pull/716

## References
https://tsu.atlassian.net/browse/IR-6884


packages/server-core/src/cluster/logs-api/logs-api.class.ts
```diff
@@ -50,16 +51,23 @@ export class LogsApiService implements ServiceInterface<void, any, LogsApiParams
...
-    const { msg, level } = logItem
+  _processLogItem = async (logItem, userId?: string) => {
+    const { msg, level, action } = logItem
+
+    delete logItem.action
+
+    if (action === 'analytics') {
+      await logToBigQuery({ ...logItem })
+      return
+    }
```
hanzlamateen: @infinitelos @barankyle should we add following check in below if condition so that it logs to BQ only when its enabled and there is a fallback to old analytics logging approach if BQ is not setup?

```suggestion
    if (action === 'analytics' && process.env.BQ_PROJECT_ID && process.env.BQ_DATASET_ID && process.env.BQ_TABLE_ID) {
      await logToBigQuery({ ...logItem })
      return
    }
```

barankyle: It can't hurt.

infinitelos (commented):
I think we need to check if config variable to see if BQ should be enabled like `BQ_ENABLED=True` and if the var doesn't exist or is false then the engine can still run without bq. We might need deployments where it is not enabled.

infinitelos (approved):

PR #1010, opened 2025-02-04
Fix lint errors when trying to commit scss changes

DanielBelmes:
## Summary
An eslint rule was put into stylelint causing an error when it detected SCSS file changes and ran. This moves the linting rule change to the right config.

Likely wasn't caught since the PR to change the rule didn't change any SCSS triggering the error.

## References
closes #_insert number here_


MbfloydIR (approved):

DanielBelmes:
merged straight in to unblock devs experiencing this issue.

PR #1014, opened 2025-02-05
Update viewer designs to fit mobile UI

CITIZENDOT:
## References
closes https://tsu.atlassian.net/browse/IR-6892

## QA Steps
| Menu Name   | Screenshot                                                                                |
| ----------- | ----------------------------------------------------------------------------------------- |
| **Profile** | image |
| **Share**   | image |
| **Emote**   | image |
| **Friends** | image |


DanielBelmes:
good bit of conflicts @aditya-mitra @CITIZENDOT

aditya-mitra:
fixed @DanielBelmes !

DanielBelmes:
Was smoke testing a little this morning and it looks like the warning to rotate to landscape also appears on desktop portrait. We should only show that message just for mobile.

CITIZENDOT:
> Was smoke testing a little this morning and it looks like the warning to rotate to landscape also appears on desktop portrait. We should only show that message just for mobile.

Fixed

achen5671 (approved):

achen5671:
Screenshot 2025-02-19 at 3 20 11 PM

should mobile view also take into account vertical view?

aditya-mitra:
> Screenshot 2025-02-19 at 3 20 11 PM
> should mobile view also take into account vertical view?

nope, only landspace modes

DanielBelmes:
Some notes. I did change the Profile Circle on the top left to be closer to the designs. It was defined as 80x80 for mobile so I went ahead and configured that. I also changed the chat text to primary as it was being defaulted to color: white from the body. Possibly might want to configure that body styling for all text to default to text primary as well. But that could lead to some unintended style effects. Also removed the media breakpoint for the joystick as it was  being lost on tablet sizes.

Further note I think we should reconsider the flow of "Want to use chat" pop up. I think design wise we could show chat and if selected that popup can appear prompting the user to verify their age. 

Lastly, I think our modals on mobile should really have a close button. It's a little odd to close them and a button to close can communicate better to the user.

DanielBelmes (approved):

DanielBelmes:
Did some more changes 😅  The top and bottom button rows were actually not centered by like 50px because we weren't pinning them to the left. This will let us do a better portrait implementation when we get to it. I simplified the loading indication for the video button. I didn't like that it was the odd one out using a different component so I implemented a loading state into those buttons that can be trigged. I also found the joystick was hugging the left side of the screen on tablet. I decided to just make it position based off of percentage which seems to work really nicely. Which also makes it more ready for also working in portrait mode.

CITIZENDOT:
Thank you @DanielBelmes

PR #1016, opened 2025-02-05
disable TransformGizmoSystem when EditorHelperState::gizmoEnabled is false

Rezmason:
We have a "gizmoEnabled" property on EditorHelperState, which `irpro-wizard` toggles when it is running. TransformGizmoSystem's reactor now returns early when gizmoEnabled is false.

Closes [#IR-6881](https://tsu.atlassian.net/browse/IR-6881).


dinomut1 (approved):
I don't see any issue with this

HexaField (approved):

PR #1022, opened 2025-02-06
add back entityExists check to ObjectLayer

DanielBelmes:
## Summary
There was an existing issue with ObjectLayer where three.js was still calling disableLayer on entities we have removed. Wanted to make this PR to note this behavior. Ideally I'd like to revisit after GLTF changes. But incase we need to fix the bug before then this exists.

## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/renderer/components/ObjectLayerComponent.ts
+++ b/packages/spatial/src/renderer/components/ObjectLayerComponent.ts
@@ -25,7 +25,7 @@ Infinite Reality Engine. All Rights Reserved.
 
 import { Object3D } from 'three'
 
-import { Entity, Types } from '@ir-engine/ecs'
+import { Entity, entityExists, Types } from '@ir-engine/ecs'
 import { defineComponent, hasComponent, removeComponent, setComponent } from '@ir-engine/ecs/src/ComponentFunctions'
 
 const maxBitWidth = 32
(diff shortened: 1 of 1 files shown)


packages/spatial/src/renderer/components/ObjectLayerComponent.ts
```diff
@@ -85,6 +85,7 @@ export const ObjectLayerMaskComponent = defineComponent({
   },
 
   enableLayer(entity: Entity, layer: number) {
+    if (!entityExists(entity)) return
```
HexaField, line 88: what is the case that we are using this API on an entity that doesn't exist? I'd rather not bloat these functions with checks if something is broken upstream

HexaField (changes requested):

packages/spatial/src/renderer/components/ObjectLayerComponent.ts
DanielBelmes: I explained in the summary. When we delete an entity. We are removing it from bitecs before three. Then three tries to access the object layer and hits our proxy object trying to access these functions which throw errors since the entity doesn't exist. There's definitely another fix here. But if it comes to it i'd like this as a temp fix if we need it.

DanielBelmes: Note this PR is also needed to fix it on ECS Authoring Layer. Also adding entities in can trigger the error on enableLayer.

PR #1024, opened 2025-02-07
IR-6006: Allow users without project permission but with editor:write scope to read public permission

hanzlamateen:
## Summary
Updated settings to ignore editor:write scope as they can work solely on permissions. Also, updated unit tests.

## References
https://tsu.atlassian.net/browse/IR-6006


barankyle (approved):
It seems like the bug this was addressing was sidestepped by something else, as I don't even see the project-setting.FIND call from the client any more. Maybe I just don't have Shopify set up right on my mt-qat project. But the code overall looks good.

PR #1029, opened 2025-02-10
Three173

AidanCaruso:
## Summary
This is a half baked experiment I'm playing with on weekends in an effort to explore the process of taking our version of threejs 2 years into the future. Includes webxr support. Among things that don't work are realism-effects (which needs updating/patching) quarks particles, mesh-bvh (currently in a half-updated state), CSM, and occasional depth issues with webxr, and probably more that I've missed.

## References
closes #_insert number here_


HexaField:
https://github.com/ir-engine/ir-engine/pull/1854

PR #1033, opened 2025-02-10
Refactor component reactor to query reactor

HexaField:
## References
closes #_insert number here_


AidanCaruso (changes requested):
Looks good, need to fix editor control functions failing on a single test case though

AidanCaruso (approved):

PR #1039, opened 2025-02-11
added function to scene functions to be able to see if a scene exists

MbfloydIR:
added logic to not show quit to dashboard popup is the modified scene no longer exists

## Summary
https://tsu.atlassian.net/browse/IR-6973
## References
closes #_insert number here_


packages/editor/src/functions/sceneFunctions.tsx
```diff
@@ -68,8 +79,8 @@ export const saveSceneGLTF = async (
     const existingScene = await API.instance.service(staticResourcePath).find({
```
HexaField: the other `existingScene` variable can be removed from saveSceneGLTF, as it is now redundant

HexaField (dismissed):

packages/editor/src/functions/sceneFunctions.tsx
MbfloydIR: implemented

DanielBelmes (approved):

PR #1043, opened 2025-02-11
Ir 6992 fix editor controls

the-reality-engineer:
## References
closes #_insert number here_


the-reality-engineer:
This PR attempts to detangle the scopes of responsibility between various systems. 

- The InputState should have no knowledge of specific camera controls. 
  - `InputState.capturingCameraOrbitEnabled` was removed
- The OrbitCameraSystem should have no knowledge of FlyCameraComponent, nor should it need to know anything about the editor state. 
- Generally, handling input between different systems should not require systems to be aware of data in other scopes of responsibility. Scope of responsibility needs to be respected. Otherwise, we introduce far more complexity then necessary, and maintenance becomes diffcult as different systems become entangled with one another. 

A couple issues remain: 

- several tests not passing
- in the examples routes, the initial camera position is not being respected. This seems to be because of the SpectatorReactor overriding the viewerEntity w/ FlyControlComponent 
  - this only happens in the example routes, probably related to the `Engine.instance.userId` not being reactive
- filterEntitiesByViewer in ClientInputFunctions needs to somehow be applied to each input heuristic, so that only entities that are part of the scene in a given viewer are included in the resulting `intersectionData` passed to `findRaycastedInput`

AidanCaruso (dismissed):
Code looks much better now, really nice work. However I'm seeing mouse middle click to pan and drag doesn't work anymore. This should be addressed before merge, as it's probably a pretty small fix.

AidanCaruso:
https://github.com/user-attachments/assets/b37b1fe7-5c7b-4c52-875e-6d9b6af7d88c

Fixed the above panning issue, however I'm seeing a lot of erroneous camera logic, above shows the avatar camera controls reenabled after exiting preview and selecting entities. This prevents the camera from moving freely, as it gets stuck to the transform of the selected entity.

HexaField:
I think it's high timewe fix the camera starting at the origin in the studio. Going to look at that now.

HexaField (approved):

the-reality-engineer:
@DanielBelmes what was the issue w/ this PR?

PR #1048, opened 2025-02-12
Fix removal of loaded gltf entities

AidanCaruso:
## Summary
Currently, loaded gltf entities are never cleaned up when the gltf component entity is removed because the isAncestor check will always return false if the entity its being checked with no longer exists. This fix is a workaround that instead checks for any loaded entities that are still children, then calls removeEntityNodeRecursively for a postorder traversal removal of entities. The idea being to ensure that of the gltf component hierarchy, only loaded entities that are still children will be removed.

## References
May fix https://tsu.atlassian.net/browse/IR-7111 - only reproduceable on deploys at the moment, will need to merge in this fix to test.

### DIFF
--- a/packages/editor/src/functions/addMediaNode.ts
+++ b/packages/editor/src/functions/addMediaNode.ts
@@ -145,8 +145,9 @@ export async function addMediaNode(
        */
       AssetState.loadAsync(url, false, UUIDComponent.generateUUID(), UndefinedEntity, Layers.Authoring as LayerID).then(
         (entity) => {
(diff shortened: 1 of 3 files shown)


AidanCaruso:
Closing for now as this will be solved in the node id refactor

HexaField:
Discussed the change - the new solution causes problems and this will be properly addressed in this PR, which solves the underlying problem around how the SourceComponent state maps are handled https://github.com/ir-engine/ir-engine/pull/1051

PR #1052, opened 2025-02-13
IR-6677-Implement-New-UI-Visualizers-icons-for-Directional-Lights-Spawn-Points-

SYBIOTE:
## References
closes #_insert number here_


packages/spatial/src/input/functions/ClientInputHeuristics.ts
```diff
@@ -113,6 +113,7 @@ export function findProximity(
 
   //TODO spatialInputObjects or inputObjects?  - inputObjects requires visible and group components
   for (const inputEntity of spatialInputObjectsQuery()) {
+    console.log('DEBUG: inputEntity', inputEntity)
```
heysokam: Just noticed this leftover console.log
Seems unintentional

SYBIOTE: The ecs integration has regressed this entirely, converting to draft 

Yup, missed that one , thanks for noticing that

SYBIOTE:
fixed the regressions, should be ready to review now

HexaField (commented):
this is likely going to conflict with @speigg 's current work quite a bit

SamMazerIR (approved):

PR #1054, opened 2025-02-13
IR-6893: fixed(uploads) issues with the sanitization of the file name for uploads and prevent unexpected failures

kattsushi:
IR-6893: fixed(uploads) issues with the sanitization of the file name for uploads and prevent unexpected failures

## Summary

https://github.com/user-attachments/assets/f26d22a3-582b-4549-a1e6-6bcfd29c7029

adding functionality to create folders with spaces and parenthesis for example `Folder (1)`

https://github.com/user-attachments/assets/6e36e846-4f94-4776-9ca3-23b1fead37e0

## References
Jira Ticked related [IR-6893](https://tsu.atlassian.net/browse/IR-6893) [IR-6321](https://tsu.atlassian.net/browse/IR-6321)

## QA Steps

[IR-6893]: https://tsu.atlassian.net/browse/IR-6893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IR-6321]: https://tsu.atlassian.net/browse/IR-6321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/common/src/utils/cleanFileName.ts
+++ b/packages/common/src/utils/cleanFileName.ts
@@ -23,14 +23,105 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
 Infinite Reality Engine. All Rights Reserved.
 */
 
-/**
- * This method takes a filename (with or without included path) and returns a cleaned version of it.
- * ensures toLower file extension, truncates a file name if too long
- * @param fullFileName
- * @param useStorageProviderLengthRestrictions
- */
 import { END_WITH_ALPHANUMERIC_REGEX, SANITIZE_FILENAME_REGEX, START_WITH_ALPHANUMERIC_REGEX } from '../regex'
 
+/**
+ * Encodes a filename by replacing special characters with a safe format:
+ * 1. Converts parentheses to underscores/hyphens: '(' -> '_', ')' -> '-'
+ * 2. Preserves existing hyphens by doubling them: '-' -> '--'
+ * 3. Converts spaces to single hyphens
+ * 4. Adds 'x' prefix if filename starts with special chars
+ * 5. Adds 'y' suffix if filename ends with special chars
+ *
+ * @param displayName - The original filename to encode
+ * @returns The encoded filename safe for storage/transmission
+ *
+ * @example
+ * // Returns "hello--world"
+ * getEncodedFileName("hello-world")
+ *
+ * // Returns "x_test-y"
+ * getEncodedFileName("(test)")
+ *
+ * // Returns "my-file-name"
+ * getEncodedFileName("my file name")
+ */
+export const getEncodedFileName = (displayName: string) => {
+  let encoded = displayName
+    .replace(/\(/g, '_') // opening parenthesis to underscore
+    .replace(/\)/g, '-') // closing parenthesis to hyphen
+    .replace(/-/g, '--') // preserve hyphens by doubling them
+    .replace(/\s/g, '-') // spaces to single hyphens
+
+  // Add 'x' at the beginning if it starts with special characters
+  if (/^[-_]/.test(encoded)) {
+    encoded = 'x' + encoded
+  }
+  // Add 'y' at the end if it ends with special characters
+  if (/[-_]$/.test(encoded)) {
+    encoded = encoded + 'y'
+  }
+  return encoded
+}
+
+/**
+ * Decodes a filename that was previously encoded by getEncodedFileName by:
+ * 1. Removing 'x' prefix and 'y' suffix if they were added during encoding
+ * 2. Converting underscores back to opening parentheses '(' in specific positions
+ * 3. Converting hyphens back to closing parentheses ')' when appropriate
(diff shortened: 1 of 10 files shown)


packages/common/src/utils/cleanFileName.ts
```diff
@@ -42,9 +42,17 @@ export const cleanFileNameString = (fullFileName: string, useStorageProviderLeng
     const lastDotIndex = hasExtension ? _lastDotIndex : fileName.length
 
     // Split the name into the part before and after the dot
-    let nameWithoutExtension = fileName.substring(0, lastDotIndex).replace(' ', '_')
+    let nameWithoutExtension = fileName.substring(0, lastDotIndex)
     const extension = fileName.substring(lastDotIndex + 1).toLowerCase()
 
+    // Sanitize the name: replace special chars and ensure alphanumeric start/end
+    nameWithoutExtension = nameWithoutExtension
```
jose-galvan: This change renames valid file names like **_3d-model.prefab.gltf_** to **_3d-model-prefab.gltf_**, which may cause issues with the default assets.
image

jose-galvan (changes requested):

packages/common/src/utils/cleanFileName.ts
kattsushi: oh good catch up let me check it out

kattsushi: fixed ty

packages/common/src/utils/cleanFileName.ts
```diff
@@ -42,9 +42,17 @@ export const cleanFileNameString = (fullFileName: string, useStorageProviderLeng
...
+    let nameWithoutExtension = fileName.substring(0, lastDotIndex)
     const extension = fileName.substring(lastDotIndex + 1).toLowerCase()
 
+    // Sanitize the name while preserving dots except the last one
+    nameWithoutExtension = nameWithoutExtension
+      .replace(/[^a-zA-Z0-9\s()_.-]/g, '-') // Allow dots along with other valid chars
+      .replace(/\s+/g, '-') // Replace spaces with dash
+      .replace(/^[^a-zA-Z0-9]+/, '') // Remove non-alphanumeric from start
+      .replace(/[^a-zA-Z0-9]+$/, '') // Remove non-alphanumeric from end
+      .replace(/-+/g, '-') // Replace multiple dashes with single dash
```
aditya-mitra: we have a [common file](https://github.com/ir-engine/ir-engine/blob/aaa707306aac320a2c508c91096b12a0bd7133aa/packages/common/src/regex/index.ts) for keeping all the regexes. we can store these regexes there

kattsushi: addressed! ty @aditya-mitra

packages/common/src/utils/cleanFileName.ts
```diff
@@ -29,6 +29,30 @@ Infinite Reality Engine. All Rights Reserved.
...
+
+  // Combine sanitized name with original extension
+  const finalName = hasExtension ? `${sanitizedName}.${extension}` : sanitizedName
+
+  return new File([file], finalName, {
+    type: file.type,
+    lastModified: file.lastModified
+  })
+}
+
```
aditya-mitra: @kattsushi this sanitation is good!

would like to point out that `sanitizeFileName` and `cleanFile` are synonyms.
instead of creating a whole different function, can we add your sanitations functions inside `cleanFileName` function?

both are doing the same job and both are have same function definitions.

kattsushi: Yeah, I agree. I actually had the same thought. However, I also wanted to split the functionality of sanitization in the frontend because I’m getting an error when allowing spaces. The cleanFileName function is being used in a lot of places on the server side, and I want to maintain the composition to allow that and avoid breaking test and functionality. But after seeing the Jira board, there’s an issue that specifically requests allowing spaces in the name [""New Folder (1)" is not seen as a valid filename and cannot be renamed or edited"](https://tsu.atlassian.net/issues/IR-6321?filter=-1) In that case, the same composed function you proposed would be the best solution.

kattsushi: Hey ty @aditya-mitra has already addressed and also resolved the ticket [IR-6321](https://tsu.atlassian.net/browse/IR-6321), which involved allowing spaces and parentheses in a specific feature. Following this, I implemented encode and decode functions to mask file and folder names while accommodating spaces. These functions ensure that the name is displayed correctly to the user with spaces and parentheses like for example `New Folder (1)`, while still decoding and saving the data in compliance with the existing validation rules.

aditya-mitra (approved):

jose-galvan (approved):

DanielBelmes (approved):

PR #1055, opened 2025-02-13
Improve Dropdown UX & Fix Admin Panel issues

CITIZENDOT:
## Summary

1. Dropdown is now powered by Popup hence, it's always visible no matter the elements nearby.
2. A dropdown in `Admin Panel > Settings > Projects` is not functional due to this, and a bug in Accordion. It's fixed now.

## References
closes https://tsu.atlassian.net/browse/IR-7220


anselm (approved):
works for me

packages/ui/src/primitives/tailwind/Select/index.tsx
```diff
@@ -123,6 +123,7 @@ const Select = ({
 
         const newDirection = spaceBelow >= spaceAbove ? 'down' : 'up'
         const _maxHeight = newDirection === 'down' ? 0.8 * spaceBelow : 0.8 * spaceAbove
+        console.log(options, newDirection, _maxHeight)
```
achen5671: might want to remove random logs

packages/ui/src/primitives/tailwind/Select/index.tsx
```diff
@@ -232,182 +223,214 @@ const Select = ({
...
+                }
+              }}
+            />
+          ))
+        ) : (
+          <div className="flex h-12 items-center justify-center bg-ui-background text-text-secondary">
+            No options available
+          </div>
+        )}
+        {/* {} */}
```
achen5671: remove

achen5671 (changes requested):
1. When you try to click on the dropdown in a modal, it closes the modal and doesnt set the value
Screenshot 2025-02-13 at 4 44 56 PM


2. When you scroll, the dropdown stays on the screen (I think this is low priority, maybe it's what we want)
Screenshot 2025-02-13 at 4 43 31 PM

CITIZENDOT:
> 1. When you try to click on the dropdown in a modal, it closes the modal and doesnt set the value

Fixed this.

2nd issue indeed seems low priority, I'll see if it can be fixed.

hanzlamateen (approved):

CITIZENDOT:
Only possible way I see to maintain the dropdown position through scroll is, by observing the position on every frame using `requestAnimationFrame`, which is an overkill for a minor usecase. We'll use this solution when this is causing any inconvenience.

achen5671 (approved):

achen5671:
looks good!

PR #1057, opened 2025-02-13
Fixing non serialized authoring layer data

DanielBelmes:
## Summary
This if condition was causing non serialized data in the authoring layer to be overridden with null.

## References
closes #_insert number here_


HexaField (approved):

HexaField (changes requested):
sky station no longer loads now :(

HexaField (approved):

PR #1059, opened 2025-02-14
Fix Lookdev Prefabs & Add Error and Suspense Boundary to Studio Dock Tabs

HexaField:
## Summary

- add suspense and error boundaries to studio panels
- fix overwrite lookdev bug
- add regression test for overwrite lookdev

## References
closes #_insert number here_


dinomut1 (approved):
This works for me

PR #1060, opened 2025-02-14
copy tailscale key and crt in macos

aditya-mitra:
## Summary

the tailscale key and crt were missing when copying them using the tailscalecerts.sh script

now it copies them from the correct location where present

### DIFF
--- a/scripts/tailscalecerts.sh
+++ b/scripts/tailscalecerts.sh
@@ -28,10 +28,12 @@ case "$(uname -s)" in
         exit 1
     fi
     domain=$(/Applications/Tailscale.app/Contents/MacOS/Tailscale cert 2>&1 | grep -o '".*"' | sed 's/"//g')
+    cp ~/Library/Containers/io.tailscale.ipn.macos/Data/$domain.key ./
+    cp ~/Library/Containers/io.tailscale.ipn.macos/Data/$domain.crt ./
     mkdir -p ./certs/tailscale && chmod a+r ./certs/tailscale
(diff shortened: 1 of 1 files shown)


scripts/tailscalecerts.sh
```diff
@@ -28,10 +28,12 @@ case "$(uname -s)" in
...
     fi
     domain=$(/Applications/Tailscale.app/Contents/MacOS/Tailscale cert 2>&1 | grep -o '".*"' | sed 's/"//g')
+    cp ~/Library/Containers/io.tailscale.ipn.macos/Data/$domain.key ./
+    cp ~/Library/Containers/io.tailscale.ipn.macos/Data/$domain.crt ./
     mkdir -p ./certs/tailscale && chmod a+r ./certs/tailscale
     /Applications/Tailscale.app/Contents/MacOS/Tailscale cert $domain 2>&1>/dev/null
-    sudo cp ./$domain.crt ./certs/tailscale/cert.pem
-    sudo cp ./$domain.key ./certs/tailscale/key.pem
+    sudo mv ./$domain.crt ./certs/tailscale/cert.pem
+    sudo mv ./$domain.key ./certs/tailscale/key.pem
```
aditya-mitra, line 36: think adding `sudo` can be skipped as well

DanielBelmes:
I'm not positive on this solution. I believe depending on a particular mac os setting it changes what needs to be done and I have to figure out what that setting is. Sometimes it gets saved in the folder of the script ran. Sometimes it gets put into the sandbox location.

PR #1064, opened 2025-02-14
Fixed bugs in changing avatars

barankyle:
## Summary

Restored default-project's onUpdate eventHook. It now adds new avatars it contains, and removes any existing avatars that were in it that no longer have any associated static-resources.

Updated avatar.hooks.ts to use raw _patch on userAvatarPath for avatar deletion, as prior implementation did not work for internal calls that do not have a user associated with them.

Resolves IR-7214

## References
closes #_insert number here_


packages/projects/default-project/projectEventHooks.ts
```diff
@@ -127,9 +128,34 @@ const config = {
         )
     )
   },
-  // onUpdate: (app: Application) => {
-  //   return installAvatarsFromProject(app, avatarsFolder)
-  // },
+  onUpdate: async (app: Application) => {
+    await Promise.all(
```
HexaField: can we move this to a helper function in server-core that takes in a list of files and a project name, such that we can re-use it for other projects that provide avatars?

HexaField (changes requested):

packages/server-core/src/util/updateAvatarRecords.ts
```diff
@@ -0,0 +1,37 @@
...
+          app,
+          manifestName,
+          path.resolve(avatarsFolder, file).replace(projectRelativeFolder + '/', '')
+        )
+      )
+  )
+  const existingAvatars = await app.service(avatarPath).find({
+    query: {
+      isPublic: true,
+      project: 'ir-engine/default-project'
```
HexaField: need to use manifest name here

HexaField (approved):

HexaField (changes requested):
whoops meant to specify request changes instead of approve... lol

HexaField (approved):

PR #1068, opened 2025-02-15
IR-6711 postprocessing fixes

SamMazerIR:
## Summary
adding render pass registration/de-registration and reuse
fixed/confirmed the following effects:
- SSAO
- Bloom
- Depth of Field
- TRAA
- LUT1D
- LUT3D
- Motion Blur

Disabled the following effects (removed from options list only):
- shockwave effect - not working, not needed
- SSR - no proper visual change, cutting losses
- SSGI -  no proper visual change, cutting losses (same shader as SSR)
- Texture Effect - engine freezes when assigning texture

## References
[https://tsu.atlassian.net/browse/IR-6711](https://tsu.atlassian.net/browse/IR-6711)


HexaField (changes requested):
TRAA seems to crash and stop all effects from working

DanielBelmes:
I've ended up disabling all realism-effects effects. I know they will simply not work or have compatibility issues.

HexaField (approved):

PR #1076, opened 2025-02-18
Move toggle functionality to icon button

SYBIOTE:
Relocate the toggle action from a button to an icon for improved user experience.


packages/ui/src/components/editor/ComponentDropdown/index.tsx
```diff
@@ -85,11 +85,12 @@ export default function ComponentDropdown({
           <Tooltip content={isMinimized.value ? 'maximize' : 'minimize'}>
             <button
               className="my-1 text-text-secondary"
-              // title={isMinimized.value ? 'maximize' : 'minimize'}
+              //title={isMinimized.value ? 'maximize' : 'minimize'}
               onClick={toggleMinimized}
             >
               <HiOutlineChevronRight
                 className={twMerge('h-4 w-4 transition-transform duration-300', !isMinimized.value ? 'rotate-90' : '')}
+                onClick={toggleMinimized}
```
CITIZENDOT, line 93: does this make a difference? button (parent) already has this onClick handler right?

SYBIOTE: It does, the icon does not bubble the event to the button, and if the button does not have the onclick defined the icon does not handle the event either

CITIZENDOT: Okay, since the div already has the `toggleMinimized` callback on it, another callback (or) button is not necessary.
Below would work, since parent div can handle clicks (toggles).

```tsx
<Tooltip content={isMinimized.value ? 'maximize' : 'minimize'}>
  <HiOutlineChevronRight
    className={twMerge(
      'h-4 w-4 text-text-secondary transition-transform duration-300',
      !isMinimized.value ? 'rotate-90' : ''
    )}
  />
</Tooltip>
```

SYBIOTE: still does not work as intended, please test this, the click on the arrow itself does not trigger the action

MbfloydIR (approved):

PR #1079, opened 2025-02-18
Michael/ir 6580 audio

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-6580
## References
closes #_insert number here_


DanielBelmes (changes requested):
- Implement cursor-pointer on play button
- Remove show controls in viewport(it's not implemented for audio)
-

DanielBelmes (approved):

MbfloydIR:
> * Implement cursor-pointer on play button
> * Remove show controls in viewport(it's not implemented for audio)

implemented

HexaField (commented):
Bit confused about why a new component is necessary - why can't we just use the media component for this? A tag component (since it doesn't seem like mediaUUID does anything) is redundant here. Can we please look at simplifying this here?

We can certainly create a separate audio UI section in the media component, but adding extra components, especially ones serialized to a scene, will only complicate things.

@DanielBelmes lets talk about this tomorrow

MbfloydIR:
> Bit confused about why a new component is necessary - why can't we just use the media component for this? A tag component (since it doesn't seem like mediaUUID does anything) is redundant here. Can we please look at simplifying this here?
> 
> We can certainly create a separate audio UI section in the media component, but adding extra components, especially ones serialized to a scene, will only complicate things.
> 
> @DanielBelmes lets talk about this tomorrow

@HexaField 

This was to simplified the video and audio Node Editor UI, as designed in the Figma.
https://www.figma.com/file/DXwFHGa5g0peqqSjyjpxQz?node-id=12480%3A6806&mode=dev

As far as I was aware, all Node editors relied on a corresponding component.

They wanted the media component's and Positional Audio Node Editor to be absorbed into the video and audio components.
But since there was no Non positional Audio component, I needed to create one, hence the General Audio Component.

It's media node reference, allows for the use of an external media component as the source of the audio. This is really only useful in the case of multiple positional audios that share the same source media component. Like speakers in a school would.

But it allowed me to create the node editor UI and functionality that was requested.

PR #1081, opened 2025-02-19
Implement Undo Redo on ECS

HexaField:
## References
closes #_insert number here_


dinomut1 (changes requested):
Seeing a no-op on trying to undo a translate with the gizmo

To reproduce:
1. Enter default scene in the studio
2. Select the platform
3. Move the platform somewhere
4. Press Ctrl-Z
5. Observe the platform does not return to original position

dinomut1 (dismissed):
Still unable to undo moving the platform in a new scene with the transform gizmo

HexaField:
@dinomut1 ok, found two last bugs which seemed to be due to undoing the first snapshot, and it now ensures there are snapshots to undo/redo before dispatching respective actions

AidanCaruso (changes requested):
What I'm seeing:

- Drag and dropping an asset, then translating it or otherwise creating a snapshot for it, will result in the asset entity being duplicated
- UUIDComponent onRemove errors when removing an asset that has been translated or otherwise had an undo snapshot created for it
- Removing entities cannot be undone. Is this out of scope for this PR?

AidanCaruso:
@HexaField Cool, undoing removed entities works now! I'm still seeing the asset duping behavior which is possibly related to the uuid component errors on removal of one of those drag n dropped assets

https://github.com/user-attachments/assets/851a9c39-b7e8-42b7-b412-a707f6cfec31

HexaField:
> @HexaField Cool, undoing removed entities works now! I'm still seeing the asset duping behavior which is possibly related to the uuid component errors on removal of one of those drag n dropped assets
> 
>  IR.Engine.-.Google.Chrome.2025-02-25.23-21-41.mp4

Think I got this now too - was recreating the entity unnecessarily.

AidanCaruso (approved):

PR #1086, opened 2025-02-20
Scene Deltas

dinomut1:
WIP, creating system for storing modifications to nested assets in a gltf file


packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -83,6 +84,13 @@ import { SelectionState } from '../services/SelectionServices'
 const tempMatrix4 = new Matrix4()
 const tempVector = new Vector3()
 
+const isInActiveScene = (entity: Entity) => {
```
HexaField: can we put this on EditorState as a helper funciton? Would be nice to reuse in a few places i reckon

packages/editor/src/systems/SceneDeltaState.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const createSourcedEntity = (source = 'testSource', nodeID = 'testNodeID') => {
+  const entity = createEntity()
+  setComponent(entity, SourceComponent, source as SourceID)
+  setComponent(entity, NodeIDComponent, nodeID as NodeID)
+  setComponent(entity, UUIDComponent, NodeIDComponent.getUUIDBySourceAndNodeID(source as SourceID, nodeID as NodeID))
+  setComponent(entity, TransformComponent)
+  setComponent(entity, EntityTreeComponent)
+
+  const obj3d = new Group()
```
HexaField, line 54: just curious, do we still need to add objects to entities? that was just an exporter quirk that no longer applies right?

packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -783,6 +786,20 @@ const loadMaterial = async (options: GLTFParserOptions, materialIndex: number) =
 
   await Promise.all(promises)
 
+  //apply deltas
+  const deltaState = getState(SceneDeltaState)
+  const sourceDelta = deltaState[options.documentID.replaceAll(/\?hash=[^-]+/g, '')]
```
HexaField: let's move this regex to GLTFComponent and add a helper function so we can reuse it, rather than redefining it everywhere time we need it

packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -783,6 +786,20 @@ const loadMaterial = async (options: GLTFParserOptions, materialIndex: number) =
...
+  if (sourceDelta) {
+    const nodeID = getComponent(materialEntity, NodeIDComponent)
+    const nodeDelta = sourceDelta[nodeID]
+    if (nodeDelta) {
+      const materialDelta = nodeDelta['materialParameters']
```
HexaField: let's define either a const or a global type override for `'materialParameters'` rather than hardcoding a magic string

packages/engine/src/gltf/SceneDeltaExporterExtension.ts
```diff
@@ -0,0 +1,65 @@
...
+import { getComponent, iterateEntityNode } from '@ir-engine/ecs'
+import { getState } from '@ir-engine/hyperflux'
+import { cleanStorageProviderURLs } from '../assets/functions/parseSceneJSON'
+import { SourceComponent } from '../scene/components/SourceComponent'
+import { SceneDeltaRegistry, SceneDeltaState } from '../scene/systems/SceneDeltaState'
+import { GLTFSceneExportExtension } from './exportGLTFScene'
+import { GLTFComponent } from './GLTFComponent'
+import { NodeIDComponent } from './NodeIDComponent'
+
+export const SCENE_DELTA_EXTENSION_NAME = 'IR_scene_delta'
```
HexaField, line 35: do we want this to be called a 'scene delta' or is a 'delta' or something else more appropriate?

packages/engine/src/scene/systems/SceneDeltaState.tsx
```diff
@@ -0,0 +1,83 @@
...
+    if (!source.value[nodeID]) source[nodeID].set({} as MaterialDeltaEntry)
+    const componentMap = source[nodeID].get(NO_PROXY_STEALTH) as MaterialDeltaEntry
+    componentMap['materialParameters'] = props
+  }
+})
+
+export const SceneDeltaSystem = defineSystem({
+  uuid: 'ir.engine.SceneDeltaSystem',
+  insert: { with: SceneObjectSystem },
+  reactor: () => {
```
HexaField, line 70: if this system is just used to define a reactor, that reactor can be moved to `SceneDeltaState` and we can get rid of the system

HexaField (dismissed):

packages/engine/src/scene/systems/SceneDeltaState.tsx
dinomut1: When it was in a state reactor the reactor wasn't being created until the state was getting used, due to us lazily defining states only when data is added to them. I had to move it to a system to get it to be active immediately on entering a scene

HexaField: gotcha ok, i would have thought the GLTF loader would initialize it before it was needed, but this makes sense

packages/editor/src/systems/SceneDeltaState.test.tsx
dinomut1: I think it's still needed for the renderer to work? I don't know

packages/engine/src/gltf/SceneDeltaExporterExtension.ts
dinomut1: I think scene delta for now because they are necessarily tied to the GLTF scene spec. This implementation is for top level deltas that all get saved directly into the root of the scene, rather than partials that get attached to nodes individually

MichaelEstes (approved):

PR #1089, opened 2025-02-20
IR-7246 Fix child GLTF transforms

AidanCaruso:
## Summary
Fixes child gltf entities in the authoring layer never getting their world matrices calculated properly, which causes issues with the gizmo erroneously using the local position instead of world position.

## References
closes https://tsu.atlassian.net/browse/IR-7246


HexaField (approved):

HexaField:
This is fine for now, though we should also seem if removing the line un-dirtying the authoring layer entities in the TransformSystem fixes this too, as even though that may be slower, could be a more elegant solution

AidanCaruso:
> This is fine for now, though we should also seem if removing the line un-dirtying the authoring layer entities in the TransformSystem fixes this too, as even though that may be slower, could be a more elegant solution

@HexaField 

This was the first thing I tried, it doesn't fix the problem - they still are never touched by the dirtySortedEntities loop, probably because the transform system only queries the simulation layer for sortedTransformEntities. 

Pretty sure fixing that would require sorting both the simulation and authoring dirty entities, as well as potentially doubling the matrix calculations, so for now I've opted to stick with doing it only in editor control functions. 

This should be fine (although clunky) as long as we stick to them for manipulating the transforms of authoring layer entities. If this proves problematic and we want more dynamic matrix updates (maybe for precalculated physics stuff baked into authoring or something) it shouldn't be hard to backpeddle in the future with a better solution.

PR #1092, opened 2025-02-20
Fix MoreOptionsMenu on studio

CITIZENDOT:
## Summary

There's a bug in MoreOptionsMenu which made it non-functional in studio (but functional in dashboard). This PR fixes it.

## References
closes #_insert number here_


aditya-mitra (approved):
looks good

PR #1095, opened 2025-02-20
update max users per instance to 10

DanielBelmes:
…etting

## Summary
funny enough the max was really just a UI change. It was always 10 within the resolver

## References
closes #_insert number here_

### DIFF
--- a/.env.local.default
+++ b/.env.local.default
@@ -138,7 +138,7 @@ INSTANCESERVER_SHUTDOWN_DELAY_MS=30000
 
 # P2P Instance variables ----------
 P2P_INSTANCE_ENABLED=true
-P2P_INSTANCE_MAX_CONNECTIONS=6
+P2P_INSTANCE_MAX_CONNECTIONS=10
 
 # Email variables ----------------
 SMTP_HOST=email-smtp.us-west-1.amazonaws.com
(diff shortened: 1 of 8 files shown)


aditya-mitra (approved):
Looks good!

will carve out a `constants.ts` for the client-core in a cleanup/tech-debt PR

barankyle:
We had set it to 5 for a reason, which was that having more than 5 avatars in an instance concurrently was causing significant performance problems. Has that been improved by recent changes?

barankyle:
Also, the P2P limit is separate, I believe. That's controlled by an ENV_VAR P2P_INSTANCE_MAX_CONNECTIONS, which we've been setting to 6. Is P2P being capped at 5 or 6 users?

DanielBelmes:
> Also, the P2P limit is separate, I believe. That's controlled by an ENV_VAR P2P_INSTANCE_MAX_CONNECTIONS, which we've been setting to 6. Is P2P being capped at 5 or 6 users?

We haven't done direct testing. But the change to 5 was supposed to just be for beta. The GLTF loader changes and other improvements are theoretically able to support past 10. @HexaField has reported results much higher than 10 locally.


Good to know about P2P

HexaField (approved):

PR #1096, opened 2025-02-20
Remove lock and unlock(Not supported well)

DanielBelmes:
## References
closes #_insert number here_


jose-galvan:
tested and error is gone 👍🏾

jose-galvan (approved):

PR #1099, opened 2025-02-21
IR-7232 use old url with correct hash to allow avatar to load after saving

HexaField:
## References
closes #_insert number here_


AidanCaruso (approved):
dang, well it definitely works now!

PR #1111, opened 2025-02-23
fix(identity-provider): ensure only one identity provider [IR-5339]

mikeplascdev:
## Summary
- Delete existing users

## References
closes #_insert number here_

## QA Steps
- go to signup

### DIFF
--- a/packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
+++ b/packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
@@ -309,6 +309,30 @@ const isSearchQuery = (context: HookContext) => {
   return queryLength === 3 && query.email && !query.email.$like && !query.email.$notlike
 }
 
+const ensureUniqueIdentity = async (context: HookContext) => {
+  const { userId } = context.data
+  if (!userId) {
+    return
+  }
+  try {
+    const { data, total } = await context.app.service(identityProviderPath).find({
+      query: {
+        userId
+      }
+    })
+
+    if (total > 0) {
+      await Promise.all(
+        data.map(async (identity) => {
+          await context.app.service(identityProviderPath).remove(identity.id)
+        })
+      )
+    }
+  } catch (err) {
+    throw new NotFound()
+  }
+}
+
 export default {
   around: {
     all: [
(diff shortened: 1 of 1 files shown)


HexaField (changes requested):
This doesn't look quite right, does this remove the ability for users to link multiple identity providers completely? If so, we definitely need to rethink this.

HexaField:
We probably just want the UI to not display the email if the user hasn't activated the magiclink.

hanzlamateen:
@mikeplascdev so what you can do is on signup page, when someone clicks on any of the SSO icon, check if they have an email provider. If so, then remove it. Because if they have already verified it then they shouldn't be able to reach signup page.

@HexaField your thoughts?

barankyle:
> @mikeplascdev so what you can do is on signup page, when someone clicks on any of the SSO icon, check if they have an email provider. If so, then remove it. Because if they have already verified it then they shouldn't be able to reach signup page.
> 
> @HexaField your thoughts?

But we still have the add-additional-providers flow from the user menu at a location, right? Which everyone can access regardless of being signed-in users or guests.

hanzlamateen:
> > @mikeplascdev so what you can do is on signup page, when someone clicks on any of the SSO icon, check if they have an email provider. If so, then remove it. Because if they have already verified it then they shouldn't be able to reach signup page.
> > @HexaField your thoughts?
> 
> But we still have the add-additional-providers flow from the user menu at a location, right? Which everyone can access regardless of being signed-in users or guests.

@barankyle Okay, so what if in my previous comment, we also include check to see if they have guest identity-provider? Because if someone has email identity-provider as well as guest identity provider, this means they havnt verified their email address yet.

Alternatively we will have to upgrade login-token service to work independant of identity-provider service and store email in it instead of identity-provider-id and on sucessful verification, add entry into identity-provider table. Therefore identity-provider can only have email entry when its verified.

barankyle:
I think doing the guest identity-provider check would be okay. Your latter suggestion seems a little more secure/less prone to weird bugs in case a guest identity-provider somehow isn't removed, but I'm not sure if there's sufficient time to do it.

mikeplascdev:
Another fix has been opened in the multitenancy project
https://github.com/theinfinitereality/irpro-multitenancy/pull/753

PR #1112, opened 2025-02-24
 Add Reactor Profiling Debug Menu

HexaField:
## References
closes #_insert number here_


packages/ui/src/primitives/tailwind/Tabs/index.tsx
```diff
@@ -110,7 +110,17 @@ const Tabs = ({
...
+        ? tabsData.map((tab, index) => {
+            return (
+              <>
+                <ErrorBoundary key={index} fallback={<></>}>
+                  <Suspense fallback={<></>}>{tab.bottomComponent}</Suspense>
+                </ErrorBoundary>
+              </>
+            )
+          })
+        : tabsData[currentTab.value]?.bottomComponent}
```
aditya-mitra: we should wrap the error boundary and suspense inside the `ReactorDebug` or make a higher order component such `ReactorDebugWrapper` with the error boundaries and suspense in it

HexaField: Any time we map an array to a subtree, we need to be wrapping it in suspense and error boundaries. This prevents any errors or suspense from crashing the entire UI. We've started doing this in the studio, this is the correct pattern here.

aditya-mitra: we might need to add error boundaries when there is a single element as well in that case.
we can map over the [`DebugTabs` values](https://github.com/ir-engine/ir-engine/pull/1112/files#diff-e7c3ea17cd35c689bea85b17184731bd08283c5328b6e4c709e2904452c9c932R75-R80) and wrap them with error boundary and suspense.

aditya-mitra (approved):
awesome

PR #1119, opened 2025-02-24
IR-5342: prevent tooltips from displaying while the screen is opaque

jose-galvan:
## Summary

Currently while the viewer is still loading a scene you can hover over the bottom of the screen when the screen is opaque
and tooltips for the user menu will be displaying, with this change pointer events are disabled 
while the screen is still opaque. 

## References
closes #_insert number here_

## QA Steps

- Enter a location
- While it's still loading hover over where the buttons are displayed at the bottom 

**Expected result**: tooltips should not be visible while the screen is opaque


aditya-mitra (approved):
looks good

PR #1125, opened 2025-02-25
System runtime memory optimization

HexaField:
## Summary

- Disables many unused systems
- Begins work on some optimizations around system execution garbage (new arrays, new sets, object/array construction/deconstruction, new functions)

## References
closes #_insert number here_


HexaField:
@DanielBelmes much more to do but this is a good step

DanielBelmes (approved):

PR #1129, opened 2025-02-25
[IR-7430] add back ecom cart icon on location

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-7430

Screenshot 2025-02-25 at 2 54 45 PM

## References
closes #_insert number here_


KonradWinkowski (approved):

packages/client-core/src/util/ViewerMenuState.tsx
```diff
@@ -39,7 +39,9 @@ export const ViewerMenuState = defineState({
       emote: false,
       friends: false,
       social: false,
-      embedframe: true
-    }
+      embedframe: true,
+      ecommerce: false
```
HexaField: we don't want to reference ecommerce here. this functionality should be generically extensible

HexaField (dismissed):

packages/client-core/src/util/ViewerMenuState.tsx
HexaField: @aditya-mitra @CITIZENDOT we really want this menu state and the hotbar state to be generic, we shouldn't have any menus explicitly referenced here, they need to be strings we register like had with the old viewer

achen5671: removed

packages/client-core/src/components/ViewerInteractions/index.tsx
```diff
@@ -74,10 +75,15 @@ export const ViewerInteractions = () => {
     )
   }
 
+  const viewerMenuState = getMutableState(ViewerMenuState)
```
aditya-mitra:
```suggestion
  const viewerMenuState = useMutableState(ViewerMenuState)
```

packages/client-core/src/components/ViewerInteractions/index.tsx
```diff
@@ -74,10 +75,15 @@ export const ViewerInteractions = () => {
...
+
   return (
     <div style={{ opacity: 1 - loadingScreenOpacity.value }} className="relative h-dvh w-full p-6">
       <div className="pointer-events-auto absolute left-0 top-0 h-fit w-full pt-[inherit]">
-        <MediaIconsBox />
+        <div className="flex items-center justify-between px-6">
+          <MediaIconsBox />
+          {viewerMenuState.userMenus.ecommerce && viewerMenuState.menuComponents.get(NO_PROXY).ecommerce}
```
aditya-mitra: we should add all the menucomponents in a generic way
all of them should go to the top right portion of the screen/viewer

packages/client-core/src/util/ViewerMenuState.tsx
aditya-mitra: we have this functionality.
just that it needs to rendered somewhere. will port it in as designs are now confirmed.

achen5671: updated

PR #1133, opened 2025-02-25
fix compressed pulish filename regex

JT00y:
## References
closes #_insert number here_


packages/common/src/model/ModelTransformFunctions.ts
```diff
@@ -416,9 +417,20 @@ const doUpload = async (projectName, fileName, buffer) => {
 
 const toProjectAndFileName = (fUploadPath: string, srcBaseURL: string): [string, string] => {
   // TODO: remove srcBaseURL if it's unnecessary
-  const [_, projectName, fileName] =
-    MATCH_ASSET_PROJECT_FILENAME_REGEX.exec(fUploadPath) ??
-    MATCH_ASSET_PROJECT_FILENAME_REGEX.exec(pathJoin(srcBaseURL, fUploadPath))!
+  let matchResult: RegExpExecArray | null
+  if (srcBaseURL.includes('publish')) {
```
dinomut1: let's get a more stringent conditional to determine if this is a publish url. Either that or make a separate function for the publish workflow

dinomut1 (changes requested):

packages/common/src/model/ModelTransformFunctions.ts
```diff
@@ -83,14 +84,14 @@ import {
...
 import ModelTransformLoader from './ModelTransformLoader'
-
 /**
  * https://ir.world/projects/ir-engine/default-project/assets/collisioncube-LOD0.glb
  * Match 1: projects/ir-engine/default-project/assets/collisioncube-LOD0.glb
  * Group 1: ir-engine/default-project
  * Group 2: collisioncube-LOD0.glb
  */
 export const MATCH_ASSET_PROJECT_FILENAME_REGEX = /projects\/([^/]+\/[^/]+)\/(?:assets|public)\/([\w\d\s\-|_./]*)$/
+const MATCH_ASSET_PUBLISH_FILENAME_REGEX = /projects\/([^/]+\/[^/]+)\/public\/publish\/([\w\d\s\-|_./]*)$/
```
dinomut1: just add an optional lookahead for the publish folder:

```const MATCH_ASSET_PROJECT_FILENAME_REGEX = /projects\/([^/]+\/[^/]+)\/(?:assets|public(?:\/publish)?)\/([\w\d\s\-|_./]*)$/```

dinomut1 (changes requested):

packages/common/src/model/ModelTransformFunctions.ts
JT00y: okay thanks, want to checkin my next pr for subfolder will make this REGEX to new RegExp(`/projects\/([^/]+\/[^/]+)\/(?:assets|public(?:\/publish\\/${sceneName}\\/([\\w\\d\\s\\-|_./]*)$`)

dinomut1 (approved):

PR #1148, opened 2025-02-26
GLTF Loader was assuming the existence of Collider component defines visibility

DanielBelmes:
## Summary
I think we should just rely on our authoring layer to decide visibility

## References
closes #_insert number here_


HexaField (changes requested):
this is required for our colliders to not load as visible cc @dinomut1

packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -1456,9 +1455,6 @@ const loadNode = async (options: GLTFParserOptions, nodeIndex: number) => {
     for (const extension in nodeDef.extensions) {
       const Component = ComponentJSONIDMap.get(extension) as any // todo
       if (!Component) continue
-      if (Component === ColliderComponent) {
```
HexaField, line 1454: actually, maybe removing this second one is okay, @dinomut1 will have the correct understanding regarding content. but defintely need to revert the other one coming from extras

HexaField (dismissed):

packages/engine/src/gltf/GLTFLoaderFunctions.ts
DanielBelmes: With the newest changes it is working as intended. Though I don't agree with this process. We should trust the authoring layer to have the right information for the components for the most part.

HexaField: this hasn't anything to do with the authoring layer, this is due to how our assets are set up. GLTF will load meshes as visible by default, but we don't want our colliders to be visible by defualt, so we need to manually remove the visible component

AidanCaruso: This change seems fine to me. 

We shouldn't be assuming visibility based on the existence of a collider component. That should be decided by the existence of the ECS visibility extension. 

Only case this is OK is if we're using legacy ECS embedding via node extras (ie the sky station colliders) in which case this assumption is fine - which this PR doesn't touch.

AidanCaruso (approved):

MichaelEstes (approved):

PR #1150, opened 2025-02-26
Fix entities not disappearing when switching scenes in studio

HexaField:
## References
closes #_insert number here_


AidanCaruso (approved):
yay!

PR #1157, opened 2025-02-27
fixed Collider-toggle prefab not working due to physicsWorld ref being undefined in callbacks

MbfloydIR:
updated the useEffect that sets up the collider callbacks to be updated if the physicsWorld dependency updates.

This updated the reference to the physicsWorld used inside the callback to the latest reference

## Summary
https://tsu.atlassian.net/browse/IR-7234

## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/physics/components/ColliderComponent.tsx
+++ b/packages/spatial/src/physics/components/ColliderComponent.tsx
@@ -148,6 +148,8 @@ const ColliderReactor = function () {
   }, [physicsWorld, triggerComponent, hasCollider])
 
   useEffect(() => {
+    if (!physicsWorld) return
+
     setCallback(entity, 'Disable Collision', () => {
       if (!physicsWorld) return
       Physics.setCollisionLayer(physicsWorld, entity, CollisionGroups.None)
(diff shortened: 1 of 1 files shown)


packages/spatial/src/physics/components/ColliderComponent.tsx
```diff
@@ -160,7 +160,7 @@ const ColliderReactor = function () {
       removeCallback(entity, 'Disable Collision')
       removeCallback(entity, 'Enable Collision')
     }
-  }, [])
+  }, [physicsWorld])
```
HexaField, line 165: let's also add a check for `if (!physicsWorld) return` at the start of this useEffect

HexaField (dismissed):

SamMazerIR (approved):

packages/spatial/src/physics/components/ColliderComponent.tsx
SamMazerIR: ah yeah that feels right

MbfloydIR: implemented

SamMazerIR (approved):

PR #1159, opened 2025-02-27
fix(user-details): get project scenes [IR-7035]

mikeplascdev:
## Summary
- get scenes by location name
## References
closes #_insert number here_

## QA Steps
- Open Studio
- Navigate to My Locations on the Dashboard
- Click Edit button for any published location
- Click the 'Scene' drop down
- Check if available scenes belongs to the current account

## Jira 
[IR-7035](https://tsu.atlassian.net/browse/IR-7035)

[IR-7035]: https://tsu.atlassian.net/browse/IR-7035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
+++ b/packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
@@ -71,6 +71,7 @@ import ErrorDialog from '@ir-engine/ui/src/components/tailwind/ErrorDialog'
 import { CheckCircleLg, Copy02Sm, EllipsisVertical } from '@ir-engine/ui/src/icons'
 import LoadingView from '@ir-engine/ui/src/primitives/tailwind/LoadingView'
 import Toggle from '@ir-engine/ui/src/primitives/tailwind/Toggle'
+import { get } from 'lodash'
 import { LoaderUtils, Quaternion, Vector3 } from 'three'
 import { NotificationService } from '../../../common/services/NotificationService'
 
@@ -111,7 +112,7 @@ const LOCATION_MAX = 10
 
 export default function AddEditLocationModal(props: {
   action: string
-  location?: LocationType
+  location?: LocationType & { projectName: string }
   sceneID?: string | null
   sceneModified?: boolean
   inStudio?: boolean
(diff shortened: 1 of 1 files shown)


packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
```diff
@@ -111,7 +112,7 @@ const LOCATION_MAX = 10
 
 export default function AddEditLocationModal(props: {
   action: string
-  location?: LocationType
+  location?: LocationType & { projectName: string }
```
HexaField: seems to be a bit strange overloading our interfaces like this, why does this property have a projectName field if it does not have it the location schema?

HexaField (changes requested):

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
mikeplascdev: I already remove it 2b0e0afa90b89d853c7318233144fda821432900

Property is added here
https://github.com/theinfinitereality/irpro-multitenancy/blob/fa502302f121968113dbcd2b449834ad3a7f0b0e/services/hooks/social/location.hooks.ts#L114

hanzlamateen (changes requested):
@mikeplascdev I dont think we need to pass project prop in the query object here. The same dialog code is used for admin panel, where all locations are shown and if we pass project here so it will cause issues in admin panel.

image

The solution for this ticket to fix `<Select />` control of UI library. There are 2 issues with it:
1. If you hover the scene field in below screenshot, the cursor will show disabled icon, which is correct. Though clicking on it in disabled state is causing dropdown to open this is wrong.
2. The color of Select control's text in disabled state too light.

image

These 2 changes should fix the issue with this ticket.

mikeplascdev:
another PR to fix this issue
https://github.com/ir-engine/ir-engine/pull/1190

PR #1164, opened 2025-02-27
IR-7168: update default scene thumbnail

jose-galvan:
## Summary

Remove thumbnail from scenes that use the default thumbnail, this is needed 
since the studio and dashboard will set a fallback image to the scene based on the theme applied 
which can only be achieved in the absence of thumbnail. 

This applies to scenes that have just been created and don't have any change since the save process 
will update the scene thumbnail. 

**Fallback image in dark mode**

image

**Fallback image in light mode**

image

## References
closes #_insert number here_

## QA Steps

1.- npm run migrate
2.- npm run dev 
3.- create a new scene 
3.- Change from dark mode to light mode and viceversa 

Expected result: fallback image should update based on theme applied

### DIFF
--- a/packages/editor/src/panels/scenes/SceneItem.tsx
+++ b/packages/editor/src/panels/scenes/SceneItem.tsx
@@ -23,11 +23,12 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
 Infinite Reality Engine. All Rights Reserved.
 */
 import { PopoverState } from '@ir-engine/client-core/src/common/services/PopoverState'
+import { ThemeState } from '@ir-engine/client-core/src/common/services/ThemeService'
 import { deleteScene } from '@ir-engine/client-core/src/world/SceneAPI'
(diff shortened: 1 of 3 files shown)


packages/editor/src/panels/scenes/SceneItem.tsx
```diff
@@ -72,16 +73,18 @@ export default function SceneItem({
     PopoverState.hidePopupover()
   }
 
+  const defaultThumbnail = theme?.value === 'dark' ? '/iR-logo-Modal-light.png' : 'iR-logo-Modal-dark.png'
```
aditya-mitra:
```suggestion
  const defaultThumbnail = theme?.value === 'dark' ? '/iR-logo-Modal-light.png' : '/iR-logo-Modal-dark.png'
```

aditya-mitra (approved):
looks good

just 1 comment to resolve

PR #1169, opened 2025-02-27
updated the reparenting method so the entity being reparented will maintain their world position, rotation, and scale

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-7281

## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/functions/EditorControlFunctions.ts
+++ b/packages/editor/src/functions/EditorControlFunctions.ts
@@ -513,7 +513,7 @@ const reparentObject = (
 
     EditorControlFunctions.positionObject([entity], [worldPosition], TransformSpace.world)
     EditorControlFunctions.rotateObject([entity], [worldRotation], TransformSpace.world)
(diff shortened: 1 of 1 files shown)


packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -597,6 +626,7 @@ export const EditorControlFunctions = {
   positionObject,
   rotateObject,
   rotateAround,
+  worldScaleObject,
```
HexaField: i don't think we want to expose this function as an API, is there a potential use case for it outside of the reparent operation?

HexaField (changes requested):

packages/editor/src/functions/EditorControlFunctions.ts
MbfloydIR: not yet, i just thought it could be useful 
I'll remove it

MbfloydIR: implemented

HexaField (approved):

PR #1171, opened 2025-02-27
[IR-7380] Importing GLTF can show erroneous component

MichaelEstes:
## References
closes #_insert number here_


SamMazerIR (approved):

packages/ui/src/components/editor/properties/mesh/index.tsx
```diff
@@ -37,10 +38,37 @@ import { Material } from 'three'
 import Accordion from '../../../../primitives/tailwind/Accordion'
 import GeometryEditor from './geometryEditor'
 
+const MaterialIsInAuthoringLayer = (material: Material): boolean => {
+  return !!UUIDComponent.getEntityByUUID(material.uuid as EntityUUID, Layers.Authoring)
```
AidanCaruso: I would like to point out though that we should be getting away from using the threejs material uuid for these types of lookups. I'm trying to phase this out as its no longer gaurunteed to be in sync with the entity uuid after the ecs authoring/layers branch was merged in. Just change this to use the material entity's UUID instead, which you can source via the mesh's MaterialInstanceComponent. This will be cleaner and avoid the Array.isArray check. After that this is good to go.

AidanCaruso (changes requested):
Overall looks good, except for the minor uuid change thats needed. As a side note, I'm going to make a follow up PR that fixes material properties not populating, which itself is a minor fix related to erroneous useage of dependencies only set for the properties panel.

AidanCaruso:
https://github.com/ir-engine/ir-engine/pull/1188 to fix material properties not populating. Your branch works beautifully with this change.

packages/ui/src/components/editor/properties/mesh/index.tsx
```diff
@@ -37,10 +38,37 @@ import { Material } from 'three'
 import Accordion from '../../../../primitives/tailwind/Accordion'
 import GeometryEditor from './geometryEditor'
 
+const MaterialIsInAuthoringLayer = (material: Material): boolean => {
```
HexaField: prefer camelCase for function names

packages/ui/src/components/editor/properties/mesh/index.tsx
```diff
@@ -37,10 +38,37 @@ import { Material } from 'three'
...
   const entity = props.entity
   const { t } = useTranslation()
   const meshComponent = getComponent(entity, MeshComponent)
+
+  const materialEditor = () => {
+    if (Array.isArray(meshComponent?.material) && meshComponent.material.length > 0) {
+      return meshComponent.material.map((material, index) => {
+        if (!MaterialIsInAuthoringLayer(material)) return null
+        return (
+          <Accordion title={t('editor:properties.mesh.materialEditor')} key={index}>
```
HexaField: we should key by data instead of list index, such that if something is unshifted, the key still points to the correct child - should help with implicit memoization

HexaField (changes requested):

HexaField (approved):

AidanCaruso (approved):

PR #1174, opened 2025-02-28
Adding QOL updates to hide children of GLB files for better UX

SamMazerIR:
## Summary
branch name might be wrong, will sort out the tickets
adding QOL updates to hide children of GLB files for better UX with model files (and by extension prefabs)

- fixing prefab save icon in hierarchy panel (the added param exportRoot must be false or you get infinitely recursive model references)

hiding children of GLB files in hierarchy:
- renamed and repurposed the no longer used feature flag "ShowModelChildren" to "HideGlbChildren" - this is currently not used, as it will require a separate PR to MT, so this feature is enabled with a @todo to hook it up to the flag by uncommenting 2 lines of code for all hiding of glb related work
- disabled dropping entities onto glb files (would make them a child that is also hidden) and gave it a red highlight when hovering over to drop
- disabled save icon for GLB files (as we do not support saving changes to GLB currently - this all still works for GLTF)
- NOTE: you can still persist changes to the GLB entity components by saving scene

- added prefab name validation in save prefab modal, as well as i18n for the modal text
- also fixed issue with applyColliders on GLTFComponent

### DIFF
--- a/packages/common/src/constants/FeatureFlags.ts
+++ b/packages/common/src/constants/FeatureFlags.ts
@@ -56,7 +56,7 @@ export const FeatureFlags = {
       SceneComplexityNotification: 'ir.editor.ui.sceneComplexityNotification',
       TransformPivot: 'ir.editor.ui.transformPivot',
       Hierarchy: {
-        ShowModelChildren: 'ir.editor.ui.hierarchy.showModelChildren'
+        HideGlbChildren: 'ir.editor.ui.hierarchy.showModelChildren'
       },
       PointClick: 'ir.editor.ui.pointClick'
     }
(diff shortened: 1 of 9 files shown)


packages/common/src/constants/FeatureFlags.ts
```diff
@@ -56,7 +56,7 @@ export const FeatureFlags = {
       SceneComplexityNotification: 'ir.editor.ui.sceneComplexityNotification',
       TransformPivot: 'ir.editor.ui.transformPivot',
       Hierarchy: {
-        ShowModelChildren: 'ir.editor.ui.hierarchy.showModelChildren'
+        HideGlbChildren: 'ir.editor.ui.hierarchy.showModelChildren'
```
HexaField: we spoke on discord and agreed revert the renaming such that we don't have to put a PR in to multitenancy

HexaField (changes requested):

HexaField (approved):

PR #1175, opened 2025-02-28
remove get create folder compression

JT00y:
This will fix the issue of GET regex permission problem which block compression


dinomut1 (changes requested):
When I try compress on publish the folder structure gets screwed up. There is an extra "publish" folder within publish where the compressed assets get moved, along with some other files
Screenshot 2025-02-28 at 11 04 07 AM

dinomut1:
ok seems good

dinomut1 (approved):

PR #1180, opened 2025-02-28
IR-7242 Instance Crashes When Using a Mobile Browser

the-reality-engineer:
## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/resources/ResourceState.tsx
+++ b/packages/spatial/src/resources/ResourceState.tsx
@@ -28,6 +28,10 @@ import {
   BufferAttribute,
   Cache,
   CompressedTexture,
+  CubeReflectionMapping,
+  CubeRefractionMapping,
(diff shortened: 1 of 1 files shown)


AidanCaruso:
We need to make sure texture discarding is done conditionally - throwing away the skybox texture breaks all subsequent threejs references to that texture. Namely, envmaps.

packages/spatial/src/resources/ResourceState.tsx
```diff
@@ -645,7 +654,7 @@ export const ResourceState = defineState({
     resources: {} as Record<string, Resource>,
     totalVertexCount: 0,
     totalBufferCount: 0,
-    debug: false
+    debug: true
```
HexaField: revert debug to true

HexaField (approved):

PR #1181, opened 2025-02-28
Fix GLTF entities deleting themselves

DanielBelmes:
## Summary
Since entities update their GLTF document occasionally this unload logic was deleting the authoring layer entities whenever they were triggering the use effect. Moved cleanup logic to scene switching logic.

## References
closes #_insert number here_


achen5671 (approved):

DanielBelmes:
Issue was misunderstood and was a bug on addMediaNode's setup

PR #1185, opened 2025-02-28
fix default value to match schema; list

infinitelos:
## Summary
the default value event_properties needs to be a list to match the schema.

## References
closes #_insert number here_


Rezmason (approved):
Excellent, excellent work 🌟

PR #1187, opened 2025-03-01
updated the shadow system to recreate the csm on cascades update

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-7359

## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/scene/systems/ShadowSystem.tsx
+++ b/packages/engine/src/scene/systems/ShadowSystem.tsx
@@ -151,7 +151,7 @@ const EntityCSMReactor = (props: { entity: Entity; rendererEntity: Entity; rende
       if (!hasComponent(rendererEntity, RendererComponent)) return
       rendererComponent.csm.set(null)
     }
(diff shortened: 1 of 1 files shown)


packages/engine/src/scene/systems/ShadowSystem.tsx
```diff
@@ -193,6 +193,7 @@ const EntityCSMReactor = (props: { entity: Entity; rendererEntity: Entity; rende
 
   useEffect(() => {
     if (!csm) return
+    csm.updateFrustums()
```
HexaField: this call and the whole useEffect is redundant now, let's  remove the csm.UpdateFrustums call, comment out the useEffect and add a todo to add support for dynamic cascades

HexaField (dismissed):

packages/engine/src/scene/systems/ShadowSystem.tsx
MbfloydIR: implemented

DanielBelmes (approved):
Working as intended. Smoke test looked great in preview mode. And published great

PR #1188, opened 2025-03-01
Fix material properties never populating

AidanCaruso:
## References
closes https://tsu.atlassian.net/browse/IR-7634

### DIFF
--- a/packages/editor/src/panels/properties/materialeditor.tsx
+++ b/packages/editor/src/panels/properties/materialeditor.tsx
@@ -179,7 +179,6 @@ export function MaterialEditor(props: { materialUUID: EntityUUID }) {
   useEffect(() => {
     prototypeName.set(material.type)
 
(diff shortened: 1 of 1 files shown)


packages/editor/src/panels/properties/materialeditor.tsx
```diff
@@ -179,7 +179,6 @@ export function MaterialEditor(props: { materialUUID: EntityUUID }) {
   useEffect(() => {
     prototypeName.set(material.type)
 
-    if (currentSelectedMaterial.value === null) return
     materialParameters.set({})
```
HexaField: `materialParameters.set({})` here is also redundant since its set again on the next line

AidanCaruso: Oh missed that. Probably a holdover from the way we used to do this

HexaField (approved):

PR #1194, opened 2025-03-01
Prevent window scroll when focusing numerical input

SYBIOTE:
Implement event listeners to disable scrolling during focus on numerical input fields. This change enhances user experience by preventing unintended scroll actions.


HexaField (dismissed):
we need to make sure any lingering event listeners are cleaned up if this component is unmounted if handleBlur is not called. we want all event listeners to be handled in a reversable useEffect responding to a state, rather than callbacks

SYBIOTE:
okay, ill add it to a unmount effect as well

packages/ui/src/components/editor/input/Numeric/index.tsx
```diff
@@ -49,6 +49,9 @@ function toPrecisionString(value: number, precision?: number) {
   return value.toLocaleString('fullwide', { useGrouping: false })
 }
 
+const disableScroll = (event: Event) => {
+  event.stopPropagation()
+}
```
aditya-mitra: move this function to inside of `useEffect` since it is local in nature

aditya-mitra (approved):
Looks good

PR #1195, opened 2025-03-02
Fix model transfrom when handling textures from 1st party assets

dinomut1:
if an asset is being compressed which references images in a read-only project, the resulting compressed images should be saved in the project that compression is taking place within

### DIFF
--- a/packages/common/src/model/ModelTransformFunctions.ts
+++ b/packages/common/src/model/ModelTransformFunctions.ts
@@ -609,7 +609,7 @@ const createTextureOperations = (
   return operations
 }
 
-const transformTexture = async (resultCache: Map<string, Texture>, operation: TextureOperation) => {
+const transformTexture = async (resultCache: Map<string, Texture>, operation: TextureOperation, index: number) => {
   const { shouldResize, shouldConvertToKTX, texture, params } = operation
 
   const hash = hashTextureOperation(operation)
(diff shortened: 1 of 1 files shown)


packages/common/src/model/ModelTransformFunctions.ts
```diff
@@ -662,6 +662,15 @@ const transformTexture = async (resultCache: Map<string, Texture>, operation: Te
     texture.setMimeType('image/ktx2')
     texture.setURI(texture.getURI().replace(/\.[^.]+$/, '.ktx2'))
   }
+
+  if (shouldResize || shouldConvertToKTX) {
+    //wipe relative path from URI
+    const uri = texture.getURI()
+    let newURI = uri.split('/').at(-1)!
+    const [_, fileName, extension] = /(.*)\.([^.]+)$/.exec(newURI)!
+    newURI = `${fileName}-${index}.${extension}`
```
Rezmason, line 671: What context or benefit does this index provide? And does this do anything zany if we compress an already compressed GLTF?

dinomut1: let's say there's a project with folder structure
base
-material1
--baseColor.png
-material2
--baseColor.png
-model.gltf

Where model.gltf references baseColor images from both folders. Since we're wiping the relative path URI, adding an index prevents overloading of filenames when saving the resulting ktx2 files into the same folder.
So the result will be 

compressed-model.gltf
baseColor-0.ktx2
baseColor-1.ktx2

instead of 
compressed-model.gltf
baseColor.ktx2 (2 files on overloaded)

JT00y (approved):

PR #1199, opened 2025-03-02
Fix color for dock panel button

SYBIOTE:
Update the border color for the dock panel max button to ensure it matches the primary text color.


aditya-mitra (approved):
looks good

PR #1206, opened 2025-03-03
fixed Marquee selection 

MbfloydIR:
updated Marquee selection so it also works on non gtlf entities and only on visible entities

## Summary
https://tsu.atlassian.net/browse/IR-7388
## References
closes #_insert number here_


DanielBelmes (approved):
Alright I fixed the extra bug I was seeing that was on live. Hopefully doesn't get reported inbetween :)

PR #1208, opened 2025-03-03
Hide avatars while tposed

AidanCaruso:
## Summary
Stops avatars from spawning in a tpose

## References
closes https://tsu.atlassian.net/browse/IR-5998

### DIFF
--- a/packages/client-core/src/networking/AvatarSpawnSystem.tsx
+++ b/packages/client-core/src/networking/AvatarSpawnSystem.tsx
@@ -155,7 +155,7 @@ export const AvatarSpawnReactor = (props: { sceneEntity: Entity }) => {
     return () => {
       removeComponent(selfAvatarEntity, VisibleComponent)
     }
-  }, [rig?.bonesToEntities, userAvatar.avatar.modelResource?.url])
+  }, [rig?.bonesToEntities.hips])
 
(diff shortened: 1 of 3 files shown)


packages/client-core/src/networking/AvatarSpawnSystem.tsx
```diff
@@ -150,6 +165,7 @@ export const AvatarSpawnReactor = (props: { sceneEntity: Entity }) => {
         entityUUID: (userID + '_avatar') as any as EntityUUID
       })
     )
+    removeComponent(selfAvatarEntity, VisibleComponent)
```
HexaField: rather than removing visible here, and the above useEffect can be replaced with 
```ts
 const rig = useOptionalComponent(selfAvatarEntity, AvatarRigComponent)
  useEffect(() => {
    setVisibleComponent(selfAvatarEntity, !!rig?.bonesToEntities?.hips?.value)
  }, [rig?.bonesToEntities.hips])
  ```
  Since we can expect this effect to be implicitly reversable
  
  We should probably also move this to the avatar system in a rig related reactor somewhere rather than in client-core

HexaField (changes requested):

HexaField (approved):

PR #1211, opened 2025-03-03
Add key to map of env reactors

DanielBelmes:
## Summary
React errors were being thrown because this tree wasn't utilizing keys. Added some simple keys to these components.

## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/scene/systems/EnvironmentSystem.tsx
+++ b/packages/engine/src/scene/systems/EnvironmentSystem.tsx
@@ -74,18 +74,52 @@ const EnvMapReactor = () => {
       {materialComponentEntities.map((materialComponentEntity, index) => {
         switch (envMapComponent) {
           case 'Skybox':
-            return <EnvMapSkyboxReactor key={envMapComponent + '-' + materialComponentEntity + '-' + index} entity={materialComponentEntity} rootEntity={entity} />
+            return (
+              <EnvMapSkyboxReactor
+                key={envMapComponent + '-' + materialComponentEntity + '-' + index}
+                entity={materialComponentEntity}
+                rootEntity={entity}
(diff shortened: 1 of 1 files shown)


packages/engine/src/scene/systems/EnvironmentSystem.tsx
```diff
@@ -71,21 +71,21 @@ const EnvMapReactor = () => {
   const materialComponentEntities = useChildrenWithComponents(entity, [MaterialStateComponent])
   return (
     <>
-      {materialComponentEntities.map((materialComponentEntity) => {
+      {materialComponentEntities.map((materialComponentEntity, index) => {
         switch (envMapComponent) {
           case 'Skybox':
-            return <EnvMapSkyboxReactor entity={materialComponentEntity} rootEntity={entity} />
+            return <EnvMapSkyboxReactor key={materialComponentEntity + '-' + index} entity={materialComponentEntity} rootEntity={entity} />
```
HexaField: should probably key with the type as well as index, to ensure they are treated as different children but still memoize correctly

HexaField (changes requested):

packages/engine/src/scene/systems/EnvironmentSystem.tsx
DanielBelmes: As in "(Skybox | Bake | Probes | ...)-entity-index"?

HexaField: Yeah

DanielBelmes: added and fixed the error

HexaField (approved):

PR #1212, opened 2025-03-04
Unable to add trigger target  Add  Apps

MbfloydIR:
corrected the way the properties are committed to account for modifying property

## Summary
https://tsu.atlassian.net/browse/IR-7656

## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/functions/EditorControlFunctions.ts
+++ b/packages/editor/src/functions/EditorControlFunctions.ts
@@ -96,7 +96,11 @@ const addOrRemoveComponent = <C extends Component<any, any>>(
   for (const entity of entities) {
     if (hasComponent(entity, SceneComponent)) continue
     if (add) {
-      setComponent(entity, component, args)
+      if (args) {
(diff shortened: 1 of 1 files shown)


packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -96,7 +96,11 @@ const addOrRemoveComponent = <C extends Component<any, any>>(
   for (const entity of entities) {
     if (hasComponent(entity, SceneComponent)) continue
     if (add) {
-      setComponent(entity, component, args)
+      if (args) {
+        EditorControlFunctions.modifyProperty([entity], component, args)
```
DanielBelmes, line 100: So we'll never call setComponent with args? Only way 102 can run is if args is falsey. So if args is passed as false it calls setComponent or undefined,null,etc....

HexaField: setComponent doesn't support period separated strings. modifyProperty handles that, so this is the correct solution here

HexaField (approved):

PR #1217, opened 2025-03-04
IR-7675: SSO signup error caused by undefined userId

jose-galvan:
## Summary

The SSO signup process throws an error (as shown in the image below), which is usually resolved by signing up again using the same SSO provider.

This error occurs because the code attempts to access a property on an object within a code block that is only executed when the object doesn't exist. The solution is to use entity.userId instead of existingEntity.userId.

image

**Note:** This should solve IR-7675 & IR-7115

## References
closes #_insert number here_

## QA Steps

- Sign up using SSO with an account that has not been used or linked to any existing user
**Expected result:** SSO authentication doesn't throw an error and user can complete signup process. 

## Evidence
https://github.com/user-attachments/assets/c6c0b587-46ff-438e-af3b-58d5e0d4ffd0

## Jira
[IR-7675](https://tsu.atlassian.net/browse/IR-7675)

[IR-7675]: https://tsu.atlassian.net/browse/IR-7675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/server-core/src/user/strategies/apple.ts
+++ b/packages/server-core/src/user/strategies/apple.ts
@@ -161,7 +161,7 @@ export class AppleStrategy extends CustomOAuthStrategy {
       await this.app.service(identityProviderPath).remove(null, {
         query: {
           type: 'guest',
-          userId: existingEntity.userId
+          userId: entity.userId
         }
       })
       await this.userLoginEntry(newIP, params)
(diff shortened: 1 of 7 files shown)


mikeplascdev (approved):
LGTM

barankyle (approved):

hurairahmateen (approved):

hurairahmateen:
LGTM

PR #1220, opened 2025-03-04
Added UUID checks on service remove calls

barankyle:
## Summary

If ID is null, it could wipe the entire table instead of the single entry that is intended.

Resolves IR-7690

## References
closes #_insert number here_


hanzlamateen (approved):

hanzlamateen:
@barankyle I have introduced an `isValidId` method to perform the check.

PR #1225, opened 2025-03-04
add touch events to select

aditya-mitra:
## Summary

previously, touch events on select used to have event propagated which closed the modal
now, touch events work properly with browsing the menu and selecting an item

`touchmove` is added to not close the dropdown items when the select menu items are browsed using touch

## References
closes #_insert number here_

## QA Steps

use browser's touch system to check this


CITIZENDOT (approved):
works nicely

PR #1226, opened 2025-03-04
remove force focus on select input

achen5671:
## References
closes #_insert number here_


aditya-mitra:
why do we want this?

achen5671:
> why do we want this?

this was forcing focus on select inputs which causes input to not be interactable on a modals


Screenshot 2025-03-04 at 11 08 59 AM

aditya-mitra:
converting to draft after discussions (need to find the correct approach)

aditya-mitra (approved):

PR #1236, opened 2025-03-04
Ir 6677 implement new UI visualizers icons and volumes ( functionality only )

DanielBelmes:
## Related Changes
https://github.com/theinfinitereality/irpro-ecommerce-tools/pull/263
https://github.com/theinfinitereality/irpro-wizard/pull/179

## References
closes #_insert number here_

## QA Steps
@SYBIOTE author


packages/editor/src/systems/ActiveHelperSystem.ts
```diff
@@ -0,0 +1,204 @@
...
+      if (getComponent(entity, ActiveHelperComponent).helperDefaultGizmo !== UndefinedEntity) continue
+      const componentStudioIcon = componentStudioIconState.get(NO_PROXY)
+      const entityComponents = getAllComponents(entity)
+      const targetComponent: any = entityComponents.find((component) =>
+        Object.keys(componentStudioIcon).find((key) => key === component.name)
+      )
+      const iconHelper = createHelperEntity(
+        entity,
+        () => {
+          const iconGizmo = createIconGizmo(componentStudioIcon[targetComponent?.name])
```
MichaelEstes: Not sure how often this query changes, but we're leaking a texture here each time this runs

SYBIOTE: whenever we add a new entity in the scene which has an ActiveHelperComponent, we dont generate a new helper of any entity which already has a helper 
i would assume the texture and the Sprite would be cleaned up automatically when we destroy the Entity

MichaelEstes: Not if you create the texture directly from the threejs loader, that's outside of the engine's context and threejs doesn't clean anything up automatically. You could store all the sprites in an array in the useeffect and call dispose on them in the cleanup function or make a child react component for each entity and use the useTexture hook

SYBIOTE: roger that

SYBIOTE: re implemented with query reactors, should solve the issue

MbfloydIR (approved):

kattsushi (approved):
LGTM, pls take care of conflicts

hzraslan (approved):

DanielBelmes:
@SYBIOTE Can we add a minimum and maximum size for the icons. Otherwise looking good. Might run it by Steve.

SYBIOTE:
Sure

DanielBelmes:
I approve(can't actually since I remade this)

packages/editor/src/systems/ActiveHelperSystem.tsx
```diff
@@ -0,0 +1,224 @@
...
+        setComponent(entity, BoundingBoxComponent)
+      }
+      return iconGizmo
+    },
+    activeHelperComponent.enabled.value,
+    ObjectLayerMasks.NodeIcon,
+    'icon-helper'
+  )
+
+  setComponent(entity, ActiveHelperComponent, { helperIconGizmo: studioIcon })
```
HexaField, line 147: this is going to set the component every time the reactor runs, potentially causing a recursive scheduling and cause lag

packages/editor/src/systems/ActiveHelperSystem.tsx
```diff
@@ -0,0 +1,224 @@
...
+}
+
+const reactor = () => {
+  useEffect(() => {
+    InputHeuristicState.addHeuristic(1, studioIconGizmoInputHeuristic as HeuristicFunctions)
+  }, [])
+
+  return (
+    <QueryReactor
+      Components={[ActiveHelperComponent, SimulationLayerComponent]}
```
HexaField, line 213: simulation layer component is redundant here, it is already included in the query reactor by specifying the layer (defaults to simulation layer)

packages/engine/src/audio/components/PositionalAudioComponent.ts
```diff
@@ -40,9 +40,7 @@ import {
   MediaComponent,
   MediaElementComponent
 } from '@ir-engine/engine/src/scene/components/MediaComponent'
-import { useMutableState } from '@ir-engine/hyperflux'
 import { NameComponent } from '@ir-engine/spatial/src/common/NameComponent'
-import { RendererState } from '@ir-engine/spatial/src/renderer/RendererState'
 
 import { S } from '@ir-engine/ecs/src/schemas/JSONSchemas'
 import { ActiveHelperComponent } from '../../../../spatial/src/common/ActiveHelperComponent'
```
HexaField, line 46: we need to use absolute package imports, not cross-package relative imports

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -274,6 +275,7 @@ export const GLTFComponentReactor = () => {
   useEffect(() => {
     if (!sceneLoaded || !scene) return
     setComponent(entity, SceneComponent, { active: true })
+    setComponent(entity, ActiveHelperComponent, { volumeEnabled: true })
```
HexaField, line 278: this is problematic, as this entity can be on the authoring layer and we do not want this being serialized

packages/engine/src/scene/components/EnvMapBakeComponent.ts
```diff
@@ -63,9 +64,17 @@ export const EnvMapBakeComponent = defineComponent({
...
   reactor: function () {
     const entity = useEntityContext()
-    const debugEnabled = useHookstate(getMutableState(RendererState).nodeHelperVisibility)
+    const activeHelperComponent = useOptionalComponent(entity, ActiveHelperComponent)
+    const debugEnabled =
+      activeHelperComponent !== undefined &&
+      activeHelperComponent.enabled.value &&
+      (activeHelperComponent.selected.value || activeHelperComponent.hovered.value)
+
+    const helperEntity = useHelperEntity(entity, () => new Mesh(sphereGeometry, helperMeshMaterial), debugEnabled)
```
HexaField, line 73: we cannot reuse geometries and meshes like this, they need to be new instances each time a new mesh is created, or we get weirdness with resource state tracking and GPU uploads

packages/engine/src/scene/components/PortalComponent.ts
```diff
@@ -92,8 +93,12 @@ export const PortalComponent = defineComponent({
 
   reactor: function () {
     const entity = useEntityContext()
-    const debugEnabled = useHookstate(getMutableState(RendererState).nodeHelperVisibility)
     const portalComponent = useComponent(entity, PortalComponent)
+    const activeHelperComponent = useOptionalComponent(entity, ActiveHelperComponent)
+    const debugEnabled =
```
HexaField, line 98: All this complexity

```
const debugEnabled =
      activeHelperComponent !== undefined &&
      activeHelperComponent.enabled.value &&
      (activeHelperComponent.selected.value || activeHelperComponent.hovered.value)
      ```
      
      Should not have to be repeated every time we need to know if the helper should be enabled or not. This shows that we have either multiple sources of truth or wrong abstractions

File: packages/engine/src/scene/components/debug/SplineHelperComponent.ts [hunk 8]
```diff
@@ -0,0 +1,163 @@
...
+const lineMaterial = () => new LineBasicMaterial({ color: 'white', opacity: 0.35 })
+const createLineGeom = () => {
+  const lineGeometry = new BufferGeometry()
+  lineGeometry.setAttribute('position', new BufferAttribute(new Float32Array(ARC_SEGMENTS * 3), 3))
+  return lineGeometry
+}
+const greenMeshMaterial = () => new MeshBasicMaterial({ color: 'lightgreen', opacity: 0.2 })
+const redMeshMaterial = () => new MeshBasicMaterial({ color: 'red', opacity: 0.2 })
+
+export const SplineHelperComponent = defineComponent({
```
HexaField, line 52: this whole file is a merge conflict. the spline helper got moved into the spline component

packages/spatial/src/common/ActiveHelperComponent.ts
```diff
@@ -23,6 +23,26 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
+    selected: S.Bool({ default: false }),
+    helperSelectedGizmo: S.Entity(UndefinedEntity), // manages the elaborate gizmo
+    directional: S.Bool({ default: false }),
+    directionalEntities: S.Array(S.Entity(UndefinedEntity)),
+    lineEntities: S.Array(S.Entity(UndefinedEntity)),
+    volumeEnabled: S.Bool({ default: false }),
+    volumeControlled: S.Bool({ default: true }),
+    sizeFactor: S.Number({ default: 0.25 })
+  }),
+  reactor: () => {
```
HexaField, line 45: this reactor does nothing, please remove it

packages/spatial/src/common/ActiveHelperComponent.ts
```diff
@@ -23,6 +23,26 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
 Infinite Reality Engine. All Rights Reserved.
 */
 
+import { S, UndefinedEntity } from '@ir-engine/ecs'
 import { defineComponent } from '@ir-engine/ecs/src/ComponentFunctions'
 
-export const ActiveHelperComponent = defineComponent({ name: 'ActiveHelperComponent' })
+export const ActiveHelperComponent = defineComponent({
+  name: 'ActiveHelperComponent',
+  jsonID: 'EE_activeHelper',
```
HexaField, line 31: we do not want a jsonID here - this component should not be serializable

packages/spatial/src/renderer/WebGLRendererSystem.tsx
```diff
@@ -254,11 +254,22 @@ const cameraReactor = () => {
...
     if (engineRendererSettings.nodeHelperVisibility.value) camera.layers.enable(ObjectLayers.NodeHelper)
     else camera.layers.disable(ObjectLayers.NodeHelper)
   }, [engineRendererSettings.nodeHelperVisibility])
 
+  useEffect(() => {
+    if (engineRendererSettings.nodeIconVisibility.value) camera.layers.enable(ObjectLayers.NodeIcon)
+    else camera.layers.disable(ObjectLayers.NodeIcon)
+  }, [engineRendererSettings.nodeIconVisibility])
+
+  /*useEffect(() => {
```
HexaField, line 268: please remove unused added commented out code

packages/spatial/src/renderer/constants/ObjectLayers.ts
```diff
@@ -50,15 +50,16 @@ export const ObjectLayers = {
   PhysicsHelper: 6 as ObjectLayer,
   AvatarHelper: 7 as ObjectLayer,
   NodeHelper: 8 as ObjectLayer,
+  NodeIcon: 9 as ObjectLayer,
```
HexaField, line 53: we do not need extra object layers here, one for all helpers should be enough. we can use entity existence, visibility etc as toggles. this is just so that we can easily turn off layers for previews, screenshots & baked rendering

packages/engine/src/scene/components/EnvMapBakeComponent.ts
SYBIOTE: this is preexisting code, ill change it while deprecating useHelperEntity

packages/engine/src/scene/components/PortalComponent.ts
SYBIOTE: yup, its ugly, should be fixed in the refactor, but we definitely want this behavior

packages/spatial/src/renderer/constants/ObjectLayers.ts
SYBIOTE: i dont agree, we have both the advanced helpers and icons , we cant specifically hide the icons unless we have a separate layer , and we have UX direction to specifically hide the icons only

packages/spatial/src/renderer/WebGLRendererSystem.tsx
SYBIOTE: sure

packages/spatial/src/common/ActiveHelperComponent.ts
SYBIOTE: yes

packages/spatial/src/common/ActiveHelperComponent.ts
SYBIOTE: ok

packages/engine/src/gltf/GLTFComponent.tsx
SYBIOTE: ill have to think about this one, we definitely want the bounding box on models

packages/engine/src/audio/components/PositionalAudioComponent.ts
SYBIOTE: sure

packages/editor/src/systems/ActiveHelperSystem.tsx
SYBIOTE: great

packages/editor/src/systems/ActiveHelperSystem.tsx
SYBIOTE: ok , ill find a better solution

packages/engine/src/gltf/GLTFComponent.tsx
HexaField: then we need to use `getSimulationCounterpart` to ensure we get the simulation entity

packages/spatial/src/renderer/constants/ObjectLayers.ts
HexaField: object camera/render layers are not the right tool for the job here, visibility or existence is the right thing.

HexaField: if there is a single source of truth whether or not the icons are visible or not (which there should be) this won't be a problem

HexaField: we have limited object layers but unlimited boolean flags. the ECS is the right abstraction layer for this, not object layers.

packages/engine/src/gltf/GLTFComponent.tsx
SYBIOTE: yes, i implemented that

packages/engine/src/scene/components/debug/SplineHelperComponent.ts
SYBIOTE: removed

packages/spatial/src/renderer/constants/ObjectLayers.ts
SYBIOTE: hmm, the point is that while we want to hide the icons visually, we still want thier functionality in the scene, visibility or existance will take away the functionality as well, so this seemed like the most appropriate and simplest course of action, ill do an alternative implementation

we dont want it to be affected by the node helper layer either, as that gets utilized by volumes

SYBIOTE: ill just make it so, that when u hide the icons the you disable the functionality too, seems more appropiate anyways

HexaField: why do we want their functionality if they are not visible? what funcitonality do they have when they are not visible?

PR #1248, opened 2025-03-05
Added support for mobile touch in InteractableComponent

derek-m-ir:
## Summary
Currently, InteractableComponent only fires callbacks on button.up.  On mobile, the component currently doesn't handle touch.

## References
closes #_insert number here_

## QA Steps
1.) Create a Scene with a simple Primitive Geometry.  
2.) Add an Interactable Component - Link Component
3.) Set the full URL (scheme included) ie: https://www.google.com
4.) Open Scene on mobile device 
5.) Tap on Primitive Geometry
6.) website should open in new tab


packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -290,14 +291,15 @@ export const InteractableComponent = defineComponent({
...
+
         if (
           buttons.Interact?.pressed &&
           !buttons.Interact?.dragging &&
           getState(InputState).capturingEntity === UndefinedEntity
         ) {
           InputState.setCapturingEntity(entity)
 
-          if (buttons.Interact?.up) {
+          if (buttons.Interact?.up || (isMobile && buttons.Interact?.touched)) {
```
HexaField, line 302: the spatial input system is designed to abstract out the need to detect the device/modality we are currently using.

@speigg what do you suggest is the correct solution here?

HexaField (changes requested):

packages/engine/src/interaction/components/InteractableComponent.ts
HexaField: could we just use `buttons.Interact.down` or `buttons.Interact.pressed` here instead?

derek-m-ir: Touch must be used for mobile. I've already tested for 'down' and it's never fired.  This solution works. 
Though I think the real issue is before this code gets called where it's not interpreting mobile gestures correctly.  I do agree that it's a bad place for the isMobile check but without a refactor/redesign it's kinda the only place to put it right now.  Ideally, there would be different impls of the input system with common functionality abstracted away and the correct impl used for the platform.

SamMazerIR: the default button state has touched set to true, so there must be something missing here. Touched is meant to be the start of a button press. I think we need to look at this deeper, possibly check why the initial press isn't registering a click properly

derek-m-ir: I'm looking back to the source of the PointerEvent in ClientInputHooks.tsx (onPointerClick()).
I attached to my iPad with Safari and I'm stepping through the code to see what's is happening.

PR #1250, opened 2025-03-05
Scene deltas addendum

dinomut1:
Fix scene saving problem


AidanCaruso (approved):
So cool to see this working!

PR #1254, opened 2025-03-06
Proper Resource Caching And Refetching

the-reality-engineer:
## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/renderer/WebGLRendererSystem.tsx
+++ b/packages/spatial/src/renderer/WebGLRendererSystem.tsx
@@ -58,6 +58,7 @@ import { getNestedChildren } from '@ir-engine/ecs'
 import { S } from '@ir-engine/ecs/src/schemas/JSONSchemas'
 import { Effect, EffectComposer, EffectPass, OutlineEffect, Pass } from 'postprocessing'
 import { CameraComponent } from '../camera/components/CameraComponent'
+import { ResourceState, ResourceType } from '../resources/ResourceState'
 import { createWebXRManager, WebXRManager } from '../xr/WebXRManager'
 import { XRState } from '../xr/XRState'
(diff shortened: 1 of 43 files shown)


HexaField (changes requested):
Envmaps do not appear to be working.

packages/spatial/src/renderer/WebGLRendererSystem.tsx
```diff
@@ -584,3 +599,14 @@ export const WebGLRendererSystem = defineSystem({
     )
   }
 })
+
+// temp helper
+globalThis.reuploadAllTextures = () => {
```
HexaField: disable helper

packages/spatial/src/renderer/WebGLRendererSystem.tsx
```diff
@@ -213,9 +214,59 @@ export const RendererComponent = defineComponent({
 
     useEffect(() => {
       const canvas = rendererComponent.canvas.value as HTMLCanvasElement
+
+      const handleWebGLContextLost = (e) => {
```
HexaField, line 218: we should test context loss again

the-reality-engineer: context restore is still not working. We should leave that for a future PR. Needs to be investigated further.

the-reality-engineer:
Env maps should be working properly now, including when avatars are switched.

AidanCaruso (changes requested):
Nice. Background vanishing on gltf unload seems fixed and no more threejs texture update warnings, but still not seeing envmaps working properly in some quick testing. 
image

HexaField:
https://github.com/ir-engine/ir-engine/pull/1421

PR #1019, opened 2025-02-06
IR-6901-Implement Moderation & Attachments API with S3 Upload Support

mshafiqmk:
## References
[IR-6448](https://tsu.atlassian.net/browse/IR-6448)
[IR-6449](https://tsu.atlassian.net/browse/IR-6449)
[IR-6901](https://tsu.atlassian.net/browse/IR-6901)

## QA Steps

[IR-6448]: https://tsu.atlassian.net/browse/IR-6448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IR-6449]: https://tsu.atlassian.net/browse/IR-6449?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IR-6901]: https://tsu.atlassian.net/browse/IR-6901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1266, opened 2025-03-06
move banned route from custom router to xrengine.config.ts default 

kattsushi:
## Summary
created PR in favor of [1019](https://github.com/ir-engine/ir-engine/pull/1019#issuecomment-2702962018) implementation feedback
## References
closes #_insert number here_


HexaField (commented):
This is a good first step. I'm wondering where the requirement for this being a new page and not a popup came from? There is a popup already with the old banning system you can see in WarningXRUISystem and useLoadLocation

kattsushi:
> This is a good first step. I'm wondering where the requirement for this being a new page and not a popup came from? There is a popup already with the old banning system you can see in WarningXRUISystem and useLoadLocation

hi @HexaField yea at first, I implemented it as a simple modal. However, during testing with @mshafiqmk , we noticed that while the modal would appear for the banned user, other users could still see that the banned user was in the same location or channel. This made it somewhat absurd to allow the banned user to remain in the location. I also tried filtering the banned user (peerUser) from the nearby users list, but that only removed them from the chat that was annoying, not the location itself. To address this issue and ensure the banned user is fully restricted from entering the instance, @mshafiqmk and I decided to create a separate route. This way, the banned user is completely restricted from accessing the instance, avoiding the odd situation where other users could still see them. That’s why we went with this approach.

also when i create the route in the CustomRouter I thought of treating the 'banned user' route similarly to how we handle custom error pages like 404 Not Found or 503 Internal Server Error. However, for a banned user, the correct status code to use is 403 Forbidden, as it specifically indicates that access is denied due to restrictions or a ban.

howerver let me check how WarningXRUISystem works to check if actually is a better option

HexaField:
The backend services should be using the banned table to prevent the user from entering the instance. We don't need/want a separate route for this, especially if network logic isn't secured, a malicious user would just be able to intercept the redirect and join the location they're banned from. We need to ensure the instance-attendance and instance-provision tables respect this.

HexaField:
The backend services should be using the banned table to prevent the user from entering the instance. We don't need/want a separate route for this, especially if network logic isn't secured, a malicious user would just be able to intercept the redirect and join the location they're banned from. We need to ensure the instance-attendance and instance-provision services respect this.

HexaField:
A modal should be fine. The xrui warning system needs to be rewritten, it won't suffice. But we also want to clean up the old implementation of the location ban logic.

mshafiqmk:
> The backend services should be using the banned table to prevent the user from entering the instance. We don't need/want a separate route for this, especially if network logic isn't secured, a malicious user would just be able to intercept the redirect and join the location they're banned from. We need to ensure the instance-attendance and instance-provision services respect this.

#1301 added required checks in the backend

HexaField (approved):

PR #1287, opened 2025-03-07
[INT] [IR-7791] Camera should collide with walls, floors, ceilings and geometry in published spaces

MichaelEstes:
## Summary
Makes sure that all entities sourced to a GLTF reflect the same camera occlusion layer

Reverts using the indirect flag when creating a BVH. Sends and receives the geometry's groups to/from the BVH generation worker.

## References
closes #_insert number here_


HexaField (changes requested):
I don't think this is right. The camera layer specifies that something should occlude the camera, the scene layer specifies it should be rendered and is the 'default' layer.

This change will result in all objects occludes the camera, not those configured to.

I actually saw this working fine in the apartment, so I wonder if when the GLTFComponent sets the camera layer if it doesn't do it on all children correctly. Otherwise this might just be user error and we should set the "occludeCamera" option in the gltf component true by default.

packages/spatial/src/camera/components/FollowCameraComponent.ts
```diff
@@ -455,7 +455,7 @@ const getMaxCamDistance = (cameraEntity: Entity, target: Vector3) => {
   let maxDistance = Math.min(followCamera.thirdPersonMaxDistance, raycastProps.rayLength)
 
   // Check hit with mid ray
-  raycaster.layers.set(ObjectLayers.Camera)
+  raycaster.layers.set(ObjectLayers.Scene)
```
HexaField: i think this can be reverted too right?

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -210,10 +211,32 @@ export const GLTFComponentReactor = () => {
   const documentLoaded = useHookstate(false)
 
   useEffect(() => {
+    if (!documentLoaded.value) return
     const occlusion = gltfComponent.cameraOcclusion.value
-    if (!occlusion) ObjectLayerMaskComponent.disableLayer(entity, ObjectLayers.Camera)
-    else ObjectLayerMaskComponent.enableLayer(entity, ObjectLayers.Camera)
-  }, [gltfComponent.cameraOcclusion])
+
+    const validLayerEntity = (child: Entity) =>
```
HexaField: we need to use `SourceComponent.getEntitiesBySource()` otherwise we may cause side effect entities in the hierarchy (like gizmos, xrui, helpers etc) to have their layer changed unnecessary. generally the GLTFComponent should never iterate down the entity tree in it's entirety, just the entities it loaded

HexaField (changes requested):

HexaField (approved):

packages/spatial/src/transform/functions/MeshBVHSystem.tsx
```diff
@@ -135,8 +141,7 @@ function acceleratedRaycast(raycaster: Raycaster, intersects: Array<Intersection
       }
     }
   } else if (ValidMeshForBVH(mesh)) {
-    // only use fallback for meshes that don't have an entity assigned - should only be XRUI
-    if (!mesh.entity) origMeshRaycastFunc.call(mesh, raycaster, intersects)
+    origMeshRaycastFunc.call(mesh, raycaster, intersects)
```
HexaField: guessing this was accidentally reverted? we only ever want the fallback to run for XRUIs, since they are not picked up by the system MeshComponent query and generally are fast enough being planes that BVH is not required (since it's just planes and would require constant recalcs as it changes size upon ui changes)

HexaField (changes requested):

packages/spatial/src/transform/functions/MeshBVHSystem.tsx
MichaelEstes: Oops, ya good catch

HexaField (commented):
Nice work, do you reckon we could write some unit regression tests that fail if this is reverted? You can borrow the BVH testing improvements in this branch https://github.com/ir-engine/ir-engine/pull/1421

EDIT: whoops i forgot we cherry picked and merged those changes separately: https://github.com/ir-engine/ir-engine/issues/1331 they are used in https://github.com/ir-engine/ir-engine/blob/IR-7791-Camera-collisions/packages/editor/src/systems/TransformGizmoSystem.test.ts#L82

HexaField (approved):

PR #1301, opened 2025-03-10
Implement user ban checks in instance server and signaling logic

mshafiqmk:
## References
[IR-7903](https://tsu.atlassian.net/browse/IR-7903)
## QA Steps

[IR-7903]: https://tsu.atlassian.net/browse/IR-7903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/server-core/src/networking/instance-provision/instance-provision.hooks.ts
```diff
@@ -23,12 +23,33 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
 Infinite Reality Engine. All Rights Reserved.
 */
 
+import { Forbidden } from '@feathersjs/errors'
+import { moderationBanPath } from '@ir-engine/common/src/schema.type.module'
 import { disallow } from 'feathers-hooks-common'
 
+const isThisUserBanned = async (context: any) => {
```
HexaField, line 30: this will apply to all locations and channels, we only want the ban to apply for the requested location

packages/instanceserver/src/NetworkFunctions.ts
```diff
@@ -152,6 +153,17 @@ export const authorizeUserToJoinServer = async (app: Application, instance: Inst
   const userId = user.id
   // disallow users from joining media servers if they are not age verified
   if (instance.channelId && !user.ageVerified) return false
+  const thisUserBanned = (await app.service(moderationBanPath).find({
```
HexaField, line 156: same here, we must scope this to the specific channel or location

HexaField (changes requested):

packages/server-core/src/networking/instance-signaling/instance-signaling.ts
```diff
@@ -81,7 +82,16 @@ const peerJoin = async (app: Application, data: InstanceSignalingDataType, param
 
   const user = params.user
   if (!user) throw new BadRequest('Must be logged in to join instance')
-
+  const thisUserBanned = await app.service(moderationBanPath).find({
```
HexaField, line 85: these must also only apply for the requesting location or channel

HexaField (changes requested):
we also want to remove the old location ban table and service

packages/server-core/src/networking/instance-provision/instance-provision.hooks.ts
mshafiqmk: Based on what was decided, if a user is banned from one location, the ban will apply to all locations.

packages/instanceserver/src/NetworkFunctions.ts
mshafiqmk: Based on what was decided, if a user is banned from one location, the ban will apply to all locations.

HexaField:
@mshafiqmk ^^ changes still requested

packages/instanceserver/src/NetworkFunctions.ts
HexaField: all locations across all domains? that doesn't sound right

packages/client-core/src/social/services/LocationService.ts
```diff
@@ -99,9 +98,6 @@ export const LocationState = defineState({
 
   socialLocationRetrieved: (location: LocationType) => {
     let bannedUsers = [] as string[]
-    location.locationBans.forEach((ban) => {
```
HexaField, line 102: we can also remove the other `bannedUsers` array creation, deduplication and removal from currentLocation state, as it is already unused

packages/client-core/src/social/services/LocationService.ts
```diff
@@ -63,7 +63,6 @@ export const LocationSeed: LocationType = {
     updatedAt: ''
   },
   locationAuthorizedUsers: [],
-  locationBans: [],
```
HexaField, line 66: we should also remove `bannedUsers` from `currentLocation` in `LocationState`

packages/server-core/src/social/location-ban/migrations/20230727072648_location-ban.ts
```diff
@@ -1,94 +0,0 @@
...
-The Original Developer is the Initial Developer. The Initial Developer of the
-Original Code is the Infinite Reality Engine team.
-
-All portions of the code written by the Infinite Reality Engine team are Copyright © 2021-2023 
-Infinite Reality Engine. All Rights Reserved.
-*/
-
-import type { Knex } from 'knex'
-
-import { locationBanPath } from '@ir-engine/common/src/schemas/social/location-ban.schema'
```
HexaField, line 28: we do not want to remove this (or any) migration, we just need to in-line the `locationBanPath` as a const

HexaField (changes requested):

packages/server-core/src/social/location-ban/migrations/20230727072648_location-ban.ts
mshafiqmk: in this is case we should only keep migration right , we can remove services and other files ?

HexaField: yes

HexaField (approved):

PR #1305, opened 2025-03-10
Fixed issue with getting builder images

barankyle:
## Summary

Needed to construct repo URL based on GCP-specific repo structure

Resolves IR-7934

## References
closes #_insert number here_


packages/server-core/src/projects/project/project-helper.ts
```diff
@@ -863,11 +861,31 @@ export const findBuilderTags = async (): Promise<Array<ProjectBuilderTagsType>>
       logger.error(err)
       return []
     }
-  } else if (gcpECRRegex) {
+  } else if (isGCP) {
     const arClient = new ArtifactRegistryClient()
     let images = [] as IDockerImage[]
+    let gcpBuilderRepo = `${process.env.SOURCE_REPO_URL}/${process.env.SOURCE_REPO_NAME_STEM}-builder`
+    switch (config.server.releaseName) {
```
HexaField: can we figure out a proper solution to go into dev? totally fine with a hack solution for the release but we should figure out what the missing data/config/discrepancy is here

HexaField (changes requested):

packages/server-core/src/projects/project/project-helper.ts
barankyle: Yeah, it probably just requires another ENV_VAR to account for how GCP has another layer of folders.

barankyle: AWS and Docker repos look something like `https://some.url.com/repo_name`. Google for some reason has things stored at e.g. `https://some.url.com/repo_name_first/repo_name_second`. And the way that Datatonic implemented the first/second naming is not 100% consistent.

hanzlamateen:
@barankyle seems like this PR has 0 changes since int got merged to dev

barankyle:
Yes, I wasn't sure if anyone was going to merge the dev version before int got merged back into dev, closing this since it's not needed.

PR #1308, opened 2025-03-10
[INT] [IR-7631] gcp colliders sporadically spawn in world zero

MichaelEstes:
## References
closes #_insert number here_


packages/spatial/src/physics/components/RigidBodyComponent.tsx
```diff
@@ -139,11 +139,15 @@ const RigidBodyReactor = () => {
   const component = useComponent(entity, RigidBodyComponent)
   const physicsWorld = Physics.useWorld(entity)!
 
+  useEffect(() => {
```
HexaField, line 142: while i would like to get to the bottom of this (i suspect it could be oddities in the physics functions & physics systems) this seems fine for now

HexaField (approved):

PR #1310, opened 2025-03-11
Ensure branch-build workflow runs on int and stg

HexaField:
## References
closes #_insert number here_


DanielBelmes:
already included in another fix

PR #1312, opened 2025-03-11
Enhance UserSearchInput with improved input handling and debounce timing

mshafiqmk:
Improve the UserSearchInput component by adding input change handling and adjusting the debounce timing from 300ms to 500ms. This enhances user experience by reducing the frequency of updates during input.


CITIZENDOT (commented):
what does onInputChange do? why can't we handle it using onChange?

mshafiqmk:
> what does onInputChange do? why can't we handle it using onChange?

onChange did not post user input as user type text in input 

onChange post even when some thing is change in select

CITIZENDOT (approved):
LGTM!

PR #1315, opened 2025-03-11
Need to update the default Shadow Bias values for lights. (Example Point light does not show shadows with default bias values.)

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-7132


packages/ui/src/components/editor/properties/light/shadowProperties/index.tsx
```diff
@@ -45,9 +52,35 @@ type LightShadowPropertiesProps = {
...
+  const shadowMapResolution = getState(RendererState).shadowMapResolution
+  const cameraEntity = getState(ReferenceSpaceState).viewerEntity
+  const camera = getComponent(cameraEntity, CameraComponent)
+  const { scene } = getComponent(getState(ReferenceSpaceState).viewerEntity, RendererComponent)
   const lightComponent = useComponent(props.entity, props.component) as any
 
+  const calculateShadowBias = () => {
+    const boundingBox = new Box3()
+
+    scene.traverse((object) => {
```
HexaField: we need to use entity functions (iterateEntityNode) instead of threejs scene traversal

HexaField (changes requested):

packages/ui/src/components/editor/properties/light/shadowProperties/index.tsx
```diff
@@ -43,11 +51,37 @@ type LightShadowPropertiesProps = {
...
+    boundingBox.getSize(size)
+    const sceneScale = Math.max(size.x, size.y, size.z)
+
+    const cameraNear = camera.near
+    const cameraFar = camera.far
+    const depthRangeFactor = cameraFar - cameraNear
 
+    // Calculate bias based on scene scale, resolution, and depth range
+    const bias = -(sceneScale / (shadowMapResolution * depthRangeFactor))
+    lightComponent.shadowBias.set(bias)
```
HexaField: we need to use the commitProperty API to ensure history and the proper authoring workflow is used for mutating the scene

HexaField (changes requested):

packages/ui/src/components/editor/properties/light/shadowProperties/index.tsx
hzraslan: Hi @HexaField, Using commitProperty API

HexaField (approved):

PR #1316, opened 2025-03-11
do not show instance chat close button when age verified on mobile

aditya-mitra:
## Summary

on mobile devices, the close should not appear when the user has already age verified. it should only be shown to guest users.

## References
closes https://tsu.atlassian.net/browse/IR-7938

## QA Steps

on mobile, the close button is not visible when user is age verified
image

on mobile, the close button is visible when the user has NOT age verified
image


CITIZENDOT (approved):

aditya-mitra:
redesigned in https://github.com/ir-engine/ir-engine/pull/1335

PR #1317, opened 2025-03-11
Mobile: Reduce spacing b/w QR Code & Inputs in Share menu

CITIZENDOT:
## References
closes https://tsu.atlassian.net/browse/IR-7933

## QA Steps

| Before                                                                                    | After                                                                                     |
| ----------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------- |
| image | image |

### DIFF
--- a/packages/client-core/src/user/menus/ShareMenu.tsx
+++ b/packages/client-core/src/user/menus/ShareMenu.tsx
@@ -197,7 +197,7 @@ const ShareMenu = (): JSX.Element => {
 
   return (
     <div className="relative z-50 h-fit max-h-[90dvh] min-w-[720px] rounded-2xl bg-surface-4 pb-4 pt-[16.5px] smh:max-h-[60dvh] smh:w-[50vw] smh:max-w-2xl smh:pb-0">
-      <div className="mx-8 grid grid-cols-3 gap-x-3 gap-y-3">
+      <div className="mx-[32px] grid grid-cols-3 gap-x-3 gap-y-3">
         <div className="col-span-3 flex w-full items-center justify-center smh:hidden">
           <Text fontWeight="medium" className="text-text-primary">
             {t('user:usermenu.share.description-share')}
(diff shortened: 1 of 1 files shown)


packages/client-core/src/user/menus/ShareMenu.tsx
```diff
@@ -197,7 +197,7 @@ const ShareMenu = (): JSX.Element => {
 
   return (
     <div className="relative z-50 h-fit max-h-[90dvh] min-w-[720px] rounded-2xl bg-surface-4 pb-4 pt-[16.5px] smh:max-h-[60dvh] smh:w-[50vw] smh:max-w-2xl smh:pb-0">
-      <div className="mx-8 grid grid-cols-3 gap-x-3 gap-y-3">
+      <div className="mx-[32px] grid grid-cols-3 gap-x-3 gap-y-3">
```
aditya-mitra: `mx-8` is also 32px right?

packages/client-core/src/user/menus/ShareMenu.tsx
```diff
@@ -212,8 +212,8 @@ const ShareMenu = (): JSX.Element => {
           </div>
 
           <div className="flex w-fit items-center justify-center smh:w-full">
-            <div className="rounded-md bg-white p-4">
-              <QRCodeSVG className="h-[114px] w-[131px] smh:h-[161px] smh:w-[184px]" value={shareLink} />
```
aditya-mitra, line 216: actually i had account for this dimensions by subtracting the padding (146-2*16). same for width

aditya-mitra (commented):
still appears too large on smaller mobile devices (width wise)
image

i think we need to shorten it to 20px
image

packages/client-core/src/user/menus/ShareMenu.tsx
CITIZENDOT: yes, but for some reason, my PC is rendering 34px margin when using rem unit.

packages/client-core/src/user/menus/ShareMenu.tsx
CITIZENDOT: Got it, Original dimensions according to designs are: 163x146. Subtracting 16px padding, we get: 147x130. 

147x130: 
image

Designs:
image

packages/client-core/src/user/menus/ShareMenu.tsx
aditya-mitra: have you set a different font size on your browser? or using some extension?

CITIZENDOT: no, i'm not sure why it happens

DanielBelmes (approved):

PR #1334, opened 2025-03-12
Material drag drop deltas

AidanCaruso:
## Summary
It's a hack. This must be reimplemented ASAP using proper material asset references.

This PR lets scene deltas track the material prototype. Also tentatively also reimplements material drag and drop to replace existing gltf materials for the sake of scene deltas working.

## References
closes https://tsu.atlassian.net/browse/IR-7942


MichaelEstes (commented):
Drag on drop works great on GLTFs, I assume we already know this won't be savable for GLBs. It doesn't seem like values are reset if they're not in material being selected. Like if you have a material that has a normal map and drag a material that doesn't the object retains the normal map texture from the first material. I'm not sure if this is intended.

MichaelEstes (approved):

AidanCaruso:
@MichaelEstes Yeah let's wait on merging until the end of the day. Last minute refactor is a WIP right now and I want to get it in first. This does work with GLBs, tested with sky station. Let me know if there's an asset it failed with in your tests.

PR #1350, opened 2025-03-12
IR-5334 rebased on int

the-reality-engineer:
## Summary
See https://github.com/ir-engine/ir-engine/pull/1254

## References
closes #_insert number here_


HexaField:
https://github.com/ir-engine/ir-engine/pull/1421

PR #1351, opened 2025-03-12
[IR-6344] Graphics settings do not work

MichaelEstes:
## References
closes #_insert number here_


SamMazerIR (approved):

kattsushi (approved):

packages/client-core/src/user/menus/SettingsMenu.tsx
```diff
@@ -185,9 +190,9 @@ function GraphicsTab() {
           max={5}
           min={0}
           step={1}
-          value={rendererState.qualityLevel.value}
-          onChange={handleQualityLevelChange}
-          onRelease={() => {}}
+          value={renderQualityLocal.value}
```
HexaField, line 196: maybe we should make this option disabled if automatic is enabled, and display the true current render quality value always

HexaField (changes requested):

packages/client-core/src/user/menus/SettingsMenu.tsx
MichaelEstes: It displays the true current render quality value, the local value is just so we don't update the render settings multiple times when the slider is dragged and only update it to the final value when it's released. All the render settings should still display their true current values when automatic is selected.

HexaField: ohhhh i see ok

HexaField (approved):

PR #1355, opened 2025-03-12
Update Tooltip position in SceneItem component

SYBIOTE:
Change the Tooltip position to "top" for better visibility in the SceneItem component.


aditya-mitra (approved):
LGTM

PR #1356, opened 2025-03-12
Add mouse event handlers to disable scrolling in MaterialPreviewer

SYBIOTE:
Implement mouse event handlers to prevent scrolling when the mouse enters the MaterialPreviewer component. This enhances user interaction by maintaining focus on the preview.


HexaField (changes requested):
why are we interupting the window events rather than changing the behaviour of the orbit camera for the preview window? this is solving it at the wrong level

SYBIOTE:
the issue noted is that the parent panel scrolls down when the mouse is over the material preview panel and is being moved around

HexaField:
Can we address that with CSS then?

SYBIOTE:
unlikely, i could take a reference to the panel and pause events for that though, i think it should solve the issue of interfering with window level events

HexaField (changes requested):
we should see if it makes sense to implement this in ClientInputHooks for all canvases

packages/editor/src/panels/materials/materialpreviewer.tsx
```diff
@@ -99,8 +99,24 @@ function MaterialPreviewCanvas() {
 export const MaterialPreviewer = () => {
   const selectedMaterial = useHookstate(getMutableState(MaterialSelectionState).selectedMaterial)
   if (!selectedMaterial.value) return null
+  const panel = document.getElementById(MATERIALS_PANEL_ID)!
+  const disableScroll = (event: Event) => {
```
HexaField, line 105: if this is the solution we go with, this needs to be a useCallback or the add/remove calls need to happen in a useEffect, otherwise it will not be adding & removing the same listeners, creating a listener memory leak

SYBIOTE: roger that

HexaField (approved):

PR #1360, opened 2025-03-13
Update FileCard component styles to include secondary text color

SYBIOTE:
Enhance the FileCard component by adding a secondary text color to its styles for improved visual consistency.


aditya-mitra (approved):
works fine

PR #1372, opened 2025-03-13
regression: remove click event listeners on profile menu

CITIZENDOT:
## Summary

[A while ago](https://github.com/ir-engine/ir-engine/pull/1319), I've added event listeners on Profile Menu, so that when username text field is out of focus, username gets updated. It's in accordance with this ticket: https://tsu.atlassian.net/browse/IR-7945.

Those event listeners are conflicting with the Settings Modal, since it is mounted on top of Profile Modal, and all interactions on SettingsModal are captured by Profile Modal's event listeners.

To avoid this complication, This PR removes those event listeners.

## References
closes https://tsu.atlassian.net/browse/IR-8094


aditya-mitra (approved):
works fine

PR #1379, opened 2025-03-13
[MT-Dev.GCP] Studio: Transform component is missing when object is placed using "Place Object at Origin" option

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-7535


aditya-mitra (approved):
works fine!

PR #1388, opened 2025-03-14
Use isIphone check instead for forcing default avatars on ios

AidanCaruso:
## Summary
I have gotten at least one report of default avatars still being visible on iphone so im switching to the isIphone check in hopes that helps. I can't test because I don't have an iphone.

## References
closes #_insert number here_


DanielBelmes:
closing in favor of https://github.com/ir-engine/ir-engine/pull/1389

PR #1395, opened 2025-03-14
IN-MT: Fix(IR-7843): fix adding validation to avoid to publish and update location that name already exists

kattsushi:
## References
closes #_insert number here_


kattsushi:
close in favor of https://github.com/ir-engine/ir-engine/pull/1445

PR #1396, opened 2025-03-14
Add scene delta state key validation for migration on scene name change

AidanCaruso:
## Summary
Currently scene deltas will never be applied due to source component changes on scene save as / rename. This is some tentative logic to detect changes and correct them.

## References
closes https://tsu.atlassian.net/browse/IR-8117


packages/engine/src/scene/systems/SceneDeltaState.tsx
```diff
@@ -79,6 +87,31 @@ export const SceneDeltaState = defineState({
         currentRoot.set(newRoot)
       }
     }, [sceneState.keys])
+
+    // validate deltas by checking for changes in the scene name, then update deltas accordingly
+    const sceneSource = useMutableState(NodesBySourceState)
+    useEffect(() => {
+      if (!sceneSource[sceneSource.keys[0]].value) return
```
HexaField: cant assume keys[0] will be the root scene

HexaField (changes requested):

dtlehrer (approved):

HexaField (approved):
Working fantastic now.

PR #1403, opened 2025-03-15
Use node ids for deltas

AidanCaruso:
## References
closes #_insert number here_


HexaField (approved):

dinomut1 (changes requested):
We can't use node IDs for deltas because you can't uniquely address changes to a specific instance of a file that exists multiple times in the same scene. So if you have multiple copies of the same asset in the scene and only change one you need to be able to keep those changes isolated to that specific asset copy

AidanCaruso:
> We can't use node IDs for deltas because you can't uniquely address changes to a specific instance of a file that exists multiple times in the same scene. So if you have multiple copies of the same asset in the scene and only change one you need to be able to keep those changes isolated to that specific asset copy

This should be okay due to each GLTF asset getting its own unique node ID generated and serialized to the scene GLTF, such that the non unique child nodes would be nested beneath each unique source asset entity node ID in the scene delta registry and avoid being overwritten. Am I missing edge cases here or something about prefabs? It works fine with copies of the same asset in my studio tests

dinomut1 (approved):

PR #1412, opened 2025-03-17
Fix for ir-8112 change that broke desktop studio

derek-m-ir:
## References
closes #_insert number here_

### DIFF
--- a/packages/ui/src/primitives/tailwind/ClickawayListener/index.tsx
+++ b/packages/ui/src/primitives/tailwind/ClickawayListener/index.tsx
@@ -28,6 +28,7 @@ import React from 'react'
 import { PopoverState } from '@ir-engine/client-core/src/common/services/PopoverState'
 
 import { getState } from '@ir-engine/hyperflux'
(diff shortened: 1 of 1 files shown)


packages/ui/src/primitives/tailwind/ClickawayListener/index.tsx
```diff
@@ -57,10 +58,11 @@ const ClickawayListener = (props: { children: React.ReactNode; onClickOutside: V
       }
     }
 
-    document.addEventListener('pointerup', handler)
+    const eventName = isMobile ? 'pointerup' : 'pointerdown'
```
CITIZENDOT, line 61: may be `touchend` (or) `touchstart` is appropriate for mobile?

derek-m-ir: I've never tested with touchend, but pointerup is doing the trick and still remains abstract

CITIZENDOT (approved):

PR #1415, opened 2025-03-17
Fix gltf importer looking for deltas in the wrong places

AidanCaruso:
## References
closes https://tsu.atlassian.net/browse/IR-8228

### DIFF
--- a/packages/engine/src/gltf/GLTFLoaderFunctions.ts
+++ b/packages/engine/src/gltf/GLTFLoaderFunctions.ts
@@ -1524,7 +1524,6 @@ const loadNode = async (options: GLTFParserOptions, nodeIndex: number) => {
   const deltas = getState(SceneDeltaState)?.[getComponent(options.entity, NodeIDComponent)]?.[nodeID]
   if (deltas) {
     for (const [componentName, delta] of Object.entries(deltas)) {
-      console.log(componentName, delta)
       const Component = ComponentJSONIDMap.get(componentName)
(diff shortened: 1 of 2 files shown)


packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -1521,8 +1521,7 @@ const loadNode = async (options: GLTFParserOptions, nodeIndex: number) => {
   await Promise.all(extensionPending)
 
   //apply deltas if they exist in state
-  const hashlessDocumentID = GLTFComponent.removeHashes(options.documentID)
-  const deltas = getState(SceneDeltaState)?.[hashlessDocumentID]?.[nodeID]
+  const deltas = getState(SceneDeltaState)?.[getComponent(options.entity, NodeIDComponent)]?.[nodeID]
```
dinomut1, line 1524: This code only goes one layer deep for nested node IDs. For now that's alright but we need to accommodate deltas that might exist inside doubly nested ECS data

dinomut1 (approved):

packages/engine/src/gltf/GLTFLoaderFunctions.ts
AidanCaruso: Yup I agree. I'd like to account for deeply nested deltas post release.

larryatinfinite:
@DanielBelmes for staging merge

PR #1421, opened 2025-03-17
Proper Resource Caching And Refetching

HexaField:
## Summary

Remake of https://github.com/ir-engine/ir-engine/pull/1254 and https://github.com/ir-engine/ir-engine/pull/1350

## References
closes #_insert number here_


packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -1562,13 +1563,20 @@ const loadScene = async (options: GLTFParserOptions, sceneIndex: number) => {
...
+        if (hasComponent(e, TransformComponent)) {
+          computeTransformMatrix(e)
+          TransformComponent.dirty[e] = 1
+        }
+      })
+    }
 
-  const animationClips = await Promise.all(animationPromises)
-  setAnimationClips(rootEntity, animationClips)
+    /** @todo this is a temporary hack */
```
HexaField, line 1591: is this an accidental reversion?

the-reality-engineer: ah looks like it

packages/engine/src/assets/loaders/texture/TextureMemoryManager.ts
```diff
@@ -43,15 +43,24 @@ const cachePromises = {} as Record<string, Promise<boolean>>
  * @returns True if the texture was offloaded, false otherwise
  */
 export async function offloadTextureData(texture: Texture): Promise<boolean> {
+  // We need a URL to be able to restore the texture later
+  const url = texture.userData?.url
+  if (!url) {
+    console.warn(`Texture does not have a URL, cannot offload: ${url}`)
```
HexaField: no point loggin the URL if it doesn't exist, is there other data we can log?

packages/engine/src/assets/loaders/texture/TextureMemoryManager.ts
```diff
@@ -43,15 +43,24 @@ const cachePromises = {} as Record<string, Promise<boolean>>
...
+    return false
+  }
+
   // Skip if ResourceCache is not available
-  if (!ResourceCache) return false
+  if (!ResourceCache) {
+    console.warn(`ResourceCache not available, cannot offload texture data: ${url}`)
```
HexaField: Can we move this to module scope? we don't need to log this for every texture

the-reality-engineer (approved):

PR #1425, opened 2025-03-18
Audio Settings Are Not Respected In Location

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-8130


packages/engine/src/audio/systems/MediaSystem.ts
```diff
@@ -84,12 +84,11 @@ export class AudioEffectPlayer {
   }
 
   #queue = new Set<string>()
-
+  //
```
HexaField: remove errant comment

HexaField (changes requested):

packages/engine/src/audio/systems/MediaSystem.ts
hzraslan: removed

HexaField (approved):

PR #1427, opened 2025-03-18
IR-8171 add server side validation for media upload filetypes

sfgarza:
## Summary
Adds server side validation for file types allowed by browser upload.

## References
closes #IR-8171
## QA Steps

First: Using Either a client such as Postman or cli utility make POST requests to the `/file-browser-upload`API endpoint using a variety of files such as images, videos, model files. Cross check these files with the ones allowed via the asset upload functionality in the browser on the project dashboard.

```sh
curl -i --location 'https://localhost:3030/file-browser-upload' \
--insecure \
--header 'Authorization: Bearer <insert authtoken>' \
--form 'media=@"/Users/santiagogarza/Downloads/putty.exe"' \ # Change to location of your file
--form 'args="[{\"project\":\"lively-nest/sfg\",\"path\":\"public/putty.exe\"}]"' #change to match to your project and remote path/filename

### CODE REVIEW

File: packages/server-core/src/media/file-browser-upload/file-browser-upload.hooks.ts [hunk 1]
```diff
@@ -49,6 +51,26 @@ import verifyScope from '../../hooks/verify-scope'
 //   return context
 // }
 
+function isValidFileType(fileType: string, fileName: string): boolean {
```
HexaField: We have a bunch of these validation functions all over the place, in the future it would be good to consolidate them into packages/common so they can be consumed isomorphically in both the client and server.

HexaField (changes requested):
Please add unit tests ensuring this functionality works

barankyle (changes requested):
I think we could also use checks like this on the file-browser.patch method. I forget exactly why we have both file-browser and file-browser-upload, but there was a reason for it, even though the latter mostly just calls the former. It might have been so we could do multi-file uploads. But it is possible to call file-browser directly, and that's a hole someone could find. 

Additionally, file-browser.patch accepts URLs for the `body` field on data, and will fetch and download whatever is at that URL (~line 464 in file-browser.class). I think we should apply the same basic checks if it's a URL. Maybe do a crude check of the mimetype before fetching, throw an error if it's not even the right mimetype, then a deeper inspection of the mimetype and content-type once we get the response back, to make sure they didn't try to slip something through?

HexaField:
> I forget exactly why we have both file-browser and file-browser-upload, but there was a reason for it, even though the latter mostly just calls the former. It might have been so we could do multi-file uploads. 

Yes, it needs a custom feathers middleware for file uploads

sfgarza:
> I think we could also use checks like this on the file-browser.patch method. I forget exactly why we have both file-browser and file-browser-upload, but there was a reason for it, even though the latter mostly just calls the former. It might have been so we could do multi-file uploads. But it is possible to call file-browser directly, and that's a hole someone could find.
> 
> Additionally, file-browser.patch accepts URLs for the `body` field on data, and will fetch and download whatever is at that URL (~line 464 in file-browser.class). I think we should apply the same basic checks if it's a URL. Maybe do a crude check of the mimetype before fetching, throw an error if it's not even the right mimetype, then a deeper inspection of the mimetype and content-type once we get the response back, to make sure they didn't try to slip something through?

@barankyle I have created a separate ticket [[8407]](https://tsu.atlassian.net/browse/IR-8407) to add similar checks to `/file-browser` and `/upload-asset` endpoints since the parameters and implementation for them differ quite a bit from `/file-browser-upload`

barankyle (approved):

jonathan-casarrubias (approved):

packages/common/src/utils/fileUploadUtils.ts
```diff
@@ -0,0 +1,24 @@
+/*
+CPAL-1.0 License
```
HexaField: Seems to be an empty file

HexaField (dismissed):

packages/common/src/utils/fileUploadUtils.ts
sfgarza: Ah my bad. Was originally going to place the isValidFileType function here but opted to place it in `packages/server-core/src/media/FileUtil.ts` instead. File has been removed

HexaField (commented):
Have we tested that multi-file upload successfully uploads just the valid files, or will a .gltf folder with an errant file in it (like a .txt or .md as is common) fail to upload?

sfgarza:
> Have we tested that multi-file upload successfully uploads just the valid files, or will a .gltf folder with an errant file in it (like a .txt or .md as is common) fail to upload?

Tested folder upload with errant files, and confirmed that only valid files will get uploaded, and errant files will get skipped . 

However I did find an issue with generated thumbnails failing to upload do to the mimetype not being set therefore defaulting the mime to application/octet-stream for the thumbnails

Proper fix would be to add the mimetype based off of the file extension here:
https://github.com/ir-engine/ir-engine/blob/680f5303b5bdc7ed8f9ebf2dca1df9f6db439897/packages/client-core/src/common/services/FileThumbnailJobState.tsx#L122

Working on making this update

barankyle (approved):

PR #1439, opened 2025-03-19
Mute video/audio when autoplaying before user interaction

CITIZENDOT:
## Summary

Due to browser policies, videos with an audio track cannot autoplay unless muted. Attempting to play such videos without muting will result in a `NotAllowedError`. As a workaround, the video is initially played on mute, and the audio is unmuted once the user interacts (click/touch) with the page.

## References
closes #_insert number here_


the-reality-engineer:
Looks great. Any reason we are using click and touch events, rather than pointer events?

CITIZENDOT:
> Looks great. Any reason we are using click and touch events, rather than pointer events?

Updated to use `pointerdown` event.

the-reality-engineer (approved):

PR #1443, opened 2025-03-19
speed up server-core tests

aditya-mitra:
## Summary

speed up the server-core tests by partitioning only the `server-core` test into 2 sets.

1 set runs the specific file name tests while the other runs the remaining.

## QA Steps

| earlier | now |
|---|---|
|  image | image |


HexaField:
awesome!!!!

HexaField:
I might need to add these steps to the repo's configuration for expected tests for it to pass properly, I will handle that tomorrow if this PR is ready by then, otherwise it might interfere with other PRs.

aditya-mitra:
we are using 2 cores in our runners. adding 4 cored cpu will speed this up further.

we can reduce this further. since eslint check (`check-errors`) takes around 5 minutes, i think this limit (of 4.5 minutes) for both the server-core tests is good.

HexaField:
yeah i am following up about using more resources. it should also make a huge difference in times and the time based failures we have

CITIZENDOT:
how about skipping `server-core` tests when change is only in `ui` package? Following is a list we can use.

|   When change is in  |             Skip             |
|:--------------------:|:----------------------------:|
|      packages/ui     | server, server-core, spatial |
|    packages/client   | server, server-core, spatial |
| packages/client-core | server, server-core, spatial |
|    packages/server   |            spatial           |
| packages/server-core |            spatial           |

More on this, here: https://theinfinitereality.slack.com/archives/C070ZGB7Q9E/p1742812084057389

HexaField:
> how about skipping `server-core` tests when change is only in `ui` package? Following is a list we can use.
> 
> When change is in	Skip
> packages/ui	server, server-core, spatial
> packages/client	server, server-core, spatial
> packages/client-core	server, server-core, spatial
> packages/server	spatial
> packages/server-core	spatial
> More on this, here: https://theinfinitereality.slack.com/archives/C070ZGB7Q9E/p1742812084057389

This seems like overkill for now

PR #1444, opened 2025-03-19
Ir 5408 remove old gltf loader and replace with gltf component

HexaField:
## References
closes #_insert number here_


AidanCaruso (dismissed):
Only problem I'm seeing is with drag and drop of newly exported materials. The exported gltf data looks correct but it might be a problem with the logic in addMediaNode now. Hard to investigate logs due to system error spam caused by the files panel on navigation to the material for drag drop.

MichaelEstes (approved):

HexaField:
found a problem with scene saving

PR #1445, opened 2025-03-19
Fix(IR-7843): fix avoid to publish and update location that name already exists

kattsushi:
Fix(IR-7843): fix avoid to publish and update location that name already exists

## References
fixes Jira ticket [IR-7843](https://tsu.atlassian.net/browse/IR-7843)

## QA Steps

[IR-7843]: https://tsu.atlassian.net/browse/IR-7843?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/client-core/i18n/en/admin.json
+++ b/packages/client-core/i18n/en/admin.json
@@ -37,7 +37,8 @@
       "confirmLocationDelete": "Do you want to delete location",
       "lobby": "Lobby",
       "featured": "Featured",
-      "locations": "Locations"
+      "locations": "Locations",
+      "nameExists": "Location name already exists"
     },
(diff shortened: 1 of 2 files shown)


MoizAdnan (changes requested):
This was being achieved through location.hooks in MT, seems like that code is broken. If those hooks can be fixed then this PR won't be needed

kattsushi:
Screenshot 2025-03-20 at 10 31 58
i will mark this one as won't do, i can reproduce it at the moment in dev and mt-int

PR #1450, opened 2025-03-19
Hotfix material drag drop

AidanCaruso:
## References
closes https://tsu.atlassian.net/browse/IR-8333

### DIFF
--- a/packages/engine/src/gltf/GLTFLoaderFunctions.ts
+++ b/packages/engine/src/gltf/GLTFLoaderFunctions.ts
@@ -812,9 +812,10 @@ const loadMaterial = async (options: GLTFParserOptions, materialIndex: number) =
         materialConstructor = prototype.prototypeConstructor
         // optionally serializing the uuid to determine if we need to replace the material -
         // this is insanely brittle but will do for now
-        if (materialDelta.uuid) materialConstructorParameters = {}
+        if (materialDelta.uuid || materialPrototype) materialConstructorParameters = {}
 
         for (const key in materialDelta) {
+          if (!materialDelta[key]) continue
           switch (prototype.arguments[key]?.type) {
             case 'color':
               materialConstructorParameters[key] = new Color(materialDelta[key])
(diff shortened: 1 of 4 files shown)


packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -812,9 +812,10 @@ const loadMaterial = async (options: GLTFParserOptions, materialIndex: number) =
         materialConstructor = prototype.prototypeConstructor
         // optionally serializing the uuid to determine if we need to replace the material -
         // this is insanely brittle but will do for now
-        if (materialDelta.uuid) materialConstructorParameters = {}
+        if (materialDelta.uuid || materialPrototype) materialConstructorParameters = {}
 
         for (const key in materialDelta) {
+          if (!materialDelta[key]) continue
```
HexaField: redundant line

AidanCaruso: @HexaField It's not, there's a serialization issue that I'm only seeing with mesh physical materials that causes some keys (i've seen only one instance so far) to be created with a value of null. Passing these into the material parameters breaks the material entirely. So I added this as a catchall for any time a value doesn't serialize properly for whatever reason.

I would like to look further into addressing this properly in the upcoming cleanup work

AidanCaruso: Yeah that can happen any time we try to serialize a value like "Infinity". But seeing as these are coming from a threejs material, there are likely other edge cases I haven't caught, for the release a quick falsey check is probably best at preventing cases of corrupted materials.

HexaField: oh very interesting
did not know that was possible

HexaField: definitely need to assert typeof or something here, since `false`, `0` and `''` will not be deserialized with this nullish check

AidanCaruso: Me neither, that's why it's being added so last minute :P

AidanCaruso: Oh good catch yeah. I'll check explicitly for null

HexaField (changes requested):
This works for the first time, but if i drag a second material from the asset panel it does not overwrite, but i see the new material populate in the material library

HexaField (approved):

PR #1452, opened 2025-03-19
[IR-8338] Unit Tests - GLTFComponent

heysokam:
## References
https://tsu.atlassian.net/browse/IR-8338


packages/engine/src/gltf/GLTFComponent.test.tsx
```diff
@@ -279,37 +296,249 @@ describe('GLTFComponent', () => {
...
+        const root = startReactor(Reactor)
+        // 4. Check the result (output)
+        expect(result).toBe(Expected)
+      })
     })
   }) //:: useInstanceID
 
   describe('removeHashes', () => {
-    it.todo('should return `@param url` with all its `?hash=...` instances removed', () => {})
+    /** @todo The Regex of this function looks completely wrong */
```
HexaField: removeHashes is unused and can be removed

heysokam: Fixed in b98e55c

PR #1460, opened 2025-03-20
IR-6992 cleanup input handling

the-reality-engineer:
## References
closes #_insert number here_


packages/ecs/src/QueryFunctions.tsx
```diff
@@ -80,7 +80,7 @@ export function removeQuery(queryOrTerms: ReturnType<typeof defineQuery> | bitEC
   }
 }
 
-export const query = (queryTerms: bitECS.QueryTerm[]) => bitECS.query(HyperFlux.store, queryTerms)
+export const query = (queryTerms: bitECS.QueryTerm[]) => bitECS.query(HyperFlux.store, queryTerms) as readonly Entity[]
```
HexaField, line 87: can we add a second optional param here that defualts to the simulation layer, and include it in the query?

the-reality-engineer: Can do that in a separate PR

HexaField (approved):
Working amazingly!

HexaField (changes requested):
Huh, found some problems in the dev test suite

- interactables not working (tested link, chair)
- always orbiting, even when not clicking

We probably want to also test WebXR

HexaField:
- still some regressions with the multiple scene example, though maybe unrelated `/examples?example=scene_multiple`
- multi avatar test has camera locked
- Minor camera glitchiness when lerping and flying at the same time (see video)

https://github.com/user-attachments/assets/cf4f6ea8-c129-49e4-89b8-bec6deaf740b

But honestly, these are minor and can be addressed later on

HexaField (approved):

the-reality-engineer:
the camera glitch is annoying, I'll find a way to fix that in a new PR. It's important for transitions to be smoothly interruptible by user input.

PR #1465, opened 2025-03-21
remove hookstate patch

HexaField:
## References
closes #_insert number here_


HexaField:
force merging without review because it's bricked CI/CD and checks pass here

PR #1471, opened 2025-03-21
MT-STG Interactable component breaks on preview, target selection

hzraslan:
## References
closes


aditya-mitra (approved):
works nice!

`useMemo` is better than `useEffect + useState`

PR #1473, opened 2025-03-21
add default project and userid to analytics logger

infinitelos:
## References
closes #_insert number here_

### DIFF
--- a/packages/common/src/logger.ts
+++ b/packages/common/src/logger.ts
@@ -43,6 +43,10 @@ import { ServiceTypes } from '../declarations'
 import config from './config'
 import { logsApiPath } from './schema.type.module'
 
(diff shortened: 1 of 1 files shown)


packages/common/src/logger.ts
```diff
@@ -43,6 +43,10 @@ import { ServiceTypes } from '../declarations'
 import config from './config'
 import { logsApiPath } from './schema.type.module'
 
+import { EditorState } from '@ir-engine/editor/src/services/EditorServices'
```
HexaField, line 46: this will cause import issues, the logger has a context state that can be injected by various routes `packages/client-core/src/util/ClientContextState.ts`

HexaField (changes requested):

PR #1478, opened 2025-03-22
[IR-7631] Fix meshes spawning at origin

MichaelEstes:
## Summary
Fixes race condition with physics system and GLTF loader transform fighting
Adds buffers to GLTF loader cache
Add typings to GLTF loader functions
Remove async promise executors so GLTF loading errors don't end up being swallowed
## References
closes #_insert number here_


packages/spatial/src/physics/systems/PhysicsPreTransformSystem.ts
```diff
@@ -168,6 +168,7 @@ const rigidbodyQuery = defineQuery([TransformComponent, RigidBodyComponent, Enti
 const colliderQuery = defineQuery([TransformComponent, ColliderComponent, EntityTreeComponent]) // @todo maybe add Not(RigidBodyComponent) to this query
 
 const filterAwakeCleanRigidbodies = (entity: Entity) => {
+  if (!RigidBodyComponent.initialized[entity]) return false
```
HexaField: aha looks like it was indeed a race condition with physics. i reckon rather than adding a new propery on the rigidbody component, we should check for the rigidbody in the physics world, which it kind of looks like `return !Physics.isSleeping(world, entity)` is kind of obscuring

packages/ecs/src/ComponentFunctions.ts
```diff
@@ -148,6 +148,14 @@ export interface ComponentPartial<
...
   /** @todo Explain ComponentPartial.onRemove(...) */
   onRemove?: (entity: Entity, component: State<ComponentType>) => void | Promise<void>
+  /**
+   * If present, this function is called after this component is created by an extension in a GLTF node
+   *
+   * @param options GLTFParserOptions
+   * @param nodeIndex index of the node that has this component as an extension
+   * @returns
+   */
+  loadNode?: (options: any, nodeIndex: number) => Promise<void>
```
HexaField: we dont want to make this an official API in the ecs, since components can have anything assigned to it's definition. the way the gltf extensions work was kind of a hack regression back towards the original threejs gltf loader as the asynchronicity of reactors made it too hard to load things deterministically. at some point we will figure it out, but we should not be overloading the ecs with concepts from higher levels

HexaField (changes requested):

packages/spatial/src/physics/systems/PhysicsPreTransformSystem.ts
```diff
@@ -168,13 +168,14 @@ const rigidbodyQuery = defineQuery([TransformComponent, RigidBodyComponent, Enti
 const colliderQuery = defineQuery([TransformComponent, ColliderComponent, EntityTreeComponent]) // @todo maybe add Not(RigidBodyComponent) to this query
 
 const filterAwakeCleanRigidbodies = (entity: Entity) => {
+  const world = Physics.getWorld(entity)
```
HexaField, line 171: these should be moved after transform dirty checks as these conditions are much less frequently met

HexaField (changes requested):

packages/spatial/src/physics/systems/PhysicsPreTransformSystem.ts
MichaelEstes: That's the race condition though, we're setting the transform position to the rigidbody position before the rigidbody has been created and initialized

HexaField: ohhhh i see. missed the return true. yeah thats annoying, we will have to optimize this another time then

packages/engine/src/gltf/MaterialExtensionComponents.tsx
```diff
@@ -819,32 +817,22 @@ export const EEMaterialComponent = defineComponent({
...
-          pending.push(
-            GLTFLoaderFunctions.assignTexture(options, v.contents).then((texture) => {
-              if (!texture) return
-              if (k === 'map') texture.colorSpace = SRGBColorSpace
-              materialParams[k] = texture
-            })
-          )
-        } else {
-          resultProperties[k] = null
+          const texture = await GLTFLoaderFunctions.assignTexture(options, v.contents)
```
HexaField: unfortunately this will slow things down as it will load textures in series not parallel, can we revert this but still solve whatever problem was occuring?

HexaField (changes requested):

packages/engine/src/gltf/MaterialExtensionComponents.tsx
MichaelEstes: Sure we can revert it, I'm interested in where you're seeing that await and .then are functionally different though

HexaField (approved):

PR #1479, opened 2025-03-22
Trigger editor panel rerender on selection update in ColliderComponentEditor

SYBIOTE:
Update the selection state to trigger a rerender of the editor panel when entities are selected in the ColliderComponentEditor.


aditya-mitra (approved):
works fine!

PR #1484, opened 2025-03-24
Fix VariantComponent not changing GLTF in Editor Preview

glenn-gartner-ir:
## Summary
Fix VariantComponent not swapping GLTF sources when previewing in the editor

## References
[closes IR-7231 ](https://tsu.atlassian.net/browse/IR-7231?atlOrigin=eyJpIjoiNzdlYjc3OGVhMmUwNDkyODgwMGRhMTg5ZTM0ZGU2NDYiLCJwIjoiaiJ9)

## QA Steps
* Add a Model Variant entity in the editor
* Add 2 levels to the Variant Component. 
* Add one GLTF mesh source to each level source. Make the meshes unique, so you can see them change. 
* Adjust the distance for each level to
   * Level 0
      * min: 0
      * max: 5
   * Level 1 
      * min: 5
      * max: 10
* Preview the level in the editor. 
* Zoom the camera close to the avatar (or within the avatar - FPS mode)
* Walk close to the Model Variant mesh, observe the mesh source change. 
* Walk away from the Model Variant mesh, observe the mesh source change again.


packages/engine/src/scene/components/VariantComponent.tsx
```diff
@@ -123,8 +124,18 @@ export const VariantComponent = defineComponent({
   reactor: () => {
     const entity = useEntityContext()
     const variantComponent = useComponent(entity, VariantComponent)
-
     const instancingComponent = useOptionalComponent(entity, InstancingComponent)
+    let childEntity = getChildrenWithComponents(entity, [GLTFComponent]).at(0)
+
+    useEffect(() => {
+      childEntity = createEntity()
```
HexaField: this useEffect should be dependent on not having an instancing component, since it handles loading models separately, and needs to remove the entity on it's unmount. essentially, this useEffect can be merged with the next one, and the childEntity can be scoped entirely to the next useEffect, cleaned up with removeEntity in the return function

packages/engine/src/scene/components/VariantComponent.tsx
```diff
@@ -123,8 +124,18 @@ export const VariantComponent = defineComponent({
   reactor: () => {
     const entity = useEntityContext()
     const variantComponent = useComponent(entity, VariantComponent)
-
     const instancingComponent = useOptionalComponent(entity, InstancingComponent)
+    let childEntity = getChildrenWithComponents(entity, [GLTFComponent]).at(0)
```
HexaField: this should be a useHookstate storing the child entity, set when it is created, rather than calculating it each time react rerenders. `let` is essentially an anti-pattern in react

HexaField (changes requested):

HexaField (approved):

PR #1486, opened 2025-03-24
Audio Settings Are Not Respected In Location

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-8130


packages/engine/src/audio/systems/MediaSystem.ts
```diff
@@ -110,9 +109,13 @@ export class AudioEffectPlayer {
     el.volume = getState(AudioState).masterVolume * volumeMultiplier
     if (el.src !== sound) el.src = sound
     el.currentTime = 0
-    if (volumeMultiplier === 0) return
-    source.start()
-    source.connect(audioContext.destination)
+    if (volumeMultiplier === 0) {
```
HexaField: what if we just escape this logic entirely at the very start of the `play` function if the volume multipler or the master volume is 0?

HexaField (changes requested):

packages/engine/src/audio/systems/MediaSystem.ts
hzraslan: yes that also works

HexaField (approved):

PR #1490, opened 2025-03-24
MT-STG Interactable component breaks on preview, target selection

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-8274


packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -273,6 +273,7 @@ export const InteractableComponent = defineComponent({
     const interactableComponent = useComponent(entity, InteractableComponent)
     const isEditing = useMutableState(EngineState).isEditing
     const modalState = useXRUIState<InteractiveModalState>()
+    const xrui = getOptionalComponent(interactableComponent.uiEntity.value, XRUIComponent)
```
HexaField: should use useHasComponent and use getComponent inside the effect

packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -319,6 +320,17 @@ export const InteractableComponent = defineComponent({
       }
     }, [isEditing.value])
 
+    useEffect(() => {
+      if (!isEditing.value && xrui) {
```
HexaField: rather than nesting conditions, escape the effect if conditions are not met, such that readability and potential race conditions can be avoided
```ts
  useEffect(() => {
    const xrui = getOptionalComponent(interactableComponent.uiEntity.value, XRUIComponent)
    if (isEditing.value || !xrui) return
    xrui.containerElement.addEventListener('click', () => callInteractCallbacks(entity))
    return () => {
      xrui.containerElement.removeEventListener('click', () => callInteractCallbacks(entity))  
    }
  }, [isEditing.value, xrui, entity])
 ```

**Review by HexaField — CHANGES_REQUESTED** [review]

File: packages/engine/src/interaction/components/InteractableComponent.ts (hunk 2 above)
Comment (hzraslan) [reply packages/engine/src/interaction/components/InteractableComponent.ts]: thanks for the review @HexaField . Updated the branch

File: packages/engine/src/interaction/components/InteractableComponent.ts [hunk 3]
```diff
@@ -273,7 +273,7 @@ export const InteractableComponent = defineComponent({
     const interactableComponent = useComponent(entity, InteractableComponent)
     const isEditing = useMutableState(EngineState).isEditing
     const modalState = useXRUIState<InteractiveModalState>()
-    const xrui = getOptionalComponent(interactableComponent.uiEntity.value, XRUIComponent)
+    const isXruiExist = hasComponent(interactableComponent.uiEntity.value, XRUIComponent)
```
HexaField: this needs to be `useHasComponent`

packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -321,15 +321,13 @@ export const InteractableComponent = defineComponent({
...
-        xrui?.containerElement.addEventListener('click', () => callInteractCallbacks(entity))
-      }
+      const xrui = getOptionalComponent(interactableComponent.uiEntity.value, XRUIComponent)
+      if (isEditing.value || !xrui) return
+      xrui.containerElement.addEventListener('click', () => callInteractCallbacks(entity))
       return () => {
-        if (xrui) {
-          xrui?.containerElement.removeEventListener('click', () => callInteractCallbacks(entity))
-        }
+        xrui.containerElement.removeEventListener('click', () => callInteractCallbacks(entity))
```
HexaField: this is not how event listeners work - the same function instance needs to be passed into both the add and remove calls. so you need to create a function definition inside this useEffect and pass that to both the addEventListener and removeEventListener

HexaField (changes requested):

packages/engine/src/interaction/components/InteractableComponent.ts
hzraslan: True, totally missed that part.  pushed the fixed version

packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -319,6 +321,16 @@ export const InteractableComponent = defineComponent({
       }
     }, [isEditing.value])
 
+    useEffect(() => {
```
HexaField, line 324: Oh I'm not thinking straight.

XRUI already has its own API for consuming DOM events. This solution is the wrong level of abstraction. I will get back to you soon with pointers as to the correct solution.

HexaField (changes requested):

DanielBelmes:
Was exploring this with hizir off dev. I somewhat agree that the label also being intractable is good UX. But there's still issues such as click interact not being settable on hover. Not positive what the difference between hover and proximity are. But we have no way of allowing click on hover activation. CC: @SamMazerIR @MbfloydIR  who have more commits in this territory.

PR #1492, opened 2025-03-25
remove unused components in the editor package

aditya-mitra:
## Summary

remove unused component and move directories and files into appropriate folder names

## References
closes https://tsu.atlassian.net/browse/IR-8474


SYBIOTE (commented):
Error pop up was being used right, did we have to delete it? Do we have an alternative error pop-up we can use across the code base?

aditya-mitra:
> Error pop up was being used right, did we have to delete it? Do we have an alternative error pop-up we can use across the code base?

It was used in a single place. It just used to show a type of error text. Currently replaced with a text with color of `ui-error` matching closer to designs.

PR #1497, opened 2025-03-26
release1.0.1  hotfix - ir 8211 video updates

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-8242

## References
closes #_insert number here_


SamMazerIR (dismissed):

packages/ui/src/components/editor/properties/transform/index.tsx
```diff
@@ -67,12 +68,21 @@ export const TransformPropertyGroup: EditorComponentType = (props) => {
...
-  position.copy(transformComponent.position.value)
-  rotation.copy(transformComponent.rotation.value)
-  scale.copy(transformComponent.scale.value)
+  const position = useHookstate(transformComponent.position.get(NO_PROXY))
+  const rotation = useHookstate(transformComponent.rotation.get(NO_PROXY))
+  const scale = useHookstate(transformComponent.scale.get(NO_PROXY))
+
+  useExecute(
+    () => {
+      position.set(transformComponent.position.get(NO_PROXY))
```
SamMazerIR: this feels gross but apparently this was Josh's suggestion for a workaround

MbfloydIR: I know, but the Transform Editor Node isn't reacting the changes in the Transform component.
The useExecute was under the recommendation of @HexaField

HexaField: as it is currently, this will cause a re-render every frame. if we compare the value and only set if the values change, we can ensure it only re-renders when the values change.

MbfloydIR: @HexaField 
updated with conditions

SamMazerIR (approved):

SYBIOTE (approved):

DanielBelmes (approved):

PR #1499, opened 2025-03-26
Update ir-engine-deploy.yml for sonarcloud

larryatinfinite:
## References
closes #_insert number here_


lucas3900:
@larryatinfinite Code looks good. Only thing is this workflow runs when you push code to one of the deployment branches. Do we want to do this on pull request prior to merge? 

Also depending on the output of this, we may want to move this to our closed source repo, to hide any sensitive info.

PR #1501, opened 2025-03-26
CLONE - [Bug] Cannot drag/reorder entities in hierarchy

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-8455


SamMazerIR:
I pulled your changes to test with, I'm noticing what looks like a horizontal scrollbar now when hovering over a dropzone for reordering. Can you look into this?

PR #1503, opened 2025-03-26
Static code analysis tools

DanielBelmes:
## References
closes #_insert number here_


barankyle (approved):

DanielBelmes:
Had to exclude WebXrManager..ts from coverage as the coverage tools are somehow breaking its build @heysokam

vitest.server.config.ts
```diff
@@ -37,8 +37,9 @@ export default defineConfig({
     testTimeout: 10000,
     maxConcurrency: 1,
     coverage: {
-      reporter: ['text', 'lcov'],
-      reportsDirectory: './coverage',
```
heysokam: The folder was needed to merge the reports for all modules without having to setup a vitest workspace. Setting up the workspace is wanted, but hasn't been done yet.

vitest.client.config.ts
```diff
@@ -43,22 +41,10 @@ export default defineConfig({
     reporters: reporters,
     slowTestThreshold: 1000,
     coverage: {
-      reporter: ['html', 'lcov'],
-      reportsDirectory: './coverage',
```
heysokam, line 47: Same comment as on the .server.config file

packages/spatial/package.json
```diff
@@ -8,7 +8,7 @@
   "description": "Core spatial system for the Infinite Reality Engine",
   "scripts": {
     "check-errors": "tsc --noemit",
-    "test": "cross-env TEST=true vitest run --config=../../vitest.client.config.ts",
```
heysokam: Mentioned on DM
Lets use `enabled: true` on the config file instead.

This is also only good if we want to -always- generate coverage no matter whats running the tests. That includes the usual unit testing workflow that requires fast iteration and continuously runnning the tests after each modification on a .test file.

With the current setup, there are specific/separate package.json scripts for coverage generation on each package. Spatial has them, for reference.

I kept this option not enabled by default because coverage generation takes time, and its better to separate the two.

Lets try to figure out a way to not blanket enable them, if thats possible with the sonar setup.

heysokam:
> @DanielBelmes:
> Had to exclude WebXrManager..ts from coverage as the coverage tools are somehow breaking its build

Maybe its because of the custom WebXR polyfill? :thinking:

DanielBelmes:
> > @DanielBelmes:
> > Had to exclude WebXrManager..ts from coverage as the coverage tools are somehow breaking its build
> 
> Maybe its because of the custom WebXR polyfill? 🤔

Josh also pointed out 
```
const FunctionTypeName = 'function'
const VoidFunctionString = 'function() {\n  }'
const EventFunctionString = 'function(event) {\n  }'
const TypeListenerFunctionString = 'function(type, listener) {\n  }'
```

vitest.server.config.ts
DanielBelmes: I think it's more interesting to keep coverage on a monorepo level

vitest.client.config.ts
DanielBelmes: I think it's more interesting to keep coverage on a monorepo level

larryatinfinite (approved):

PR #1505, opened 2025-03-26
Ir 3467 add dimensions

JT00y:
https://tsu.atlassian.net/browse/IR-3467


packages/server-core/src/media/static-resource/migrations/20241107185830_static-resource-dimensions.ts
```diff
@@ -0,0 +1,55 @@
...
+ * @returns { Promise<void> }
+ */
+export async function down(knex: Knex): Promise<void> {
+  const dimensionsColumnExists = await knex.schema.hasColumn(staticResourcePath, 'dimensions')
+  if (dimensionsColumnExists) {
+    await knex.schema.alterTable(staticResourcePath, async (table) => {
+      table.dropColumn('dimensions')
+    })
+  }
+}
```
dinomut1: We might want to make 3 columns for "width", "height", "depth" so we can do ranged queries. I don't think MySQL supports ranged queries on json columns

dinomut1 (changes requested):
let's also add logic that kicks off a thumbnail job either if the asset is missing a thumbnail OR is missing dimensions

dinomut1 (changes requested):
I uploaded an asset, and the dimensions were calculated to be 0,0,0 when the asset contains geometry

To reproduce:
in studio, upload rings.glb from the default project
observe the thumbnail generation occur
open the file properties modal
you will see the dimensions are still 0,0,0

packages/client-core/src/common/services/FileThumbnailJobState.tsx
```diff
@@ -197,62 +299,21 @@ export const FileThumbnailJobState = defineState({
...
-      // If there are more files left to be processed in the list we have specified, refetch the query
-      if (resourceQuery.total > resourceQuery.data.length) resourceQuery.refetch()
-    }, [resourceQuery.data])
+  },
+  useGenerateDimensions: async (files: readonly FileBrowserContentType[]) => {
+    useGenerateHelper(
+      files,
+      (file) => (file.type === 'glf' || file.type === 'glb' ? file.key : undefined),
+      {
+        $and: [{ width: '0' }, { height: '0' }, { depth: '0' }]
```
dinomut1: If someone uploads a gltf that has no geometry in it (like a .material.gltf or .lookdev.gltf) the system will repeatedly try to calculate its dimensions. We should probably make these fields nullable so that they start as null, then if they are calculated to be 0 we know that we don't need to recalculate them again

dinomut1 (changes requested):

packages/client-core/src/common/services/FileThumbnailJobState.tsx
```diff
@@ -170,89 +212,103 @@ const uploadThumbnail = async (src: string, projectName: string, blob: Blob | nu
...
-      }
-
-      // If there are more files left to be processed in the list we have specified, refetch the query
-      if (resourceQuery.total > resourceQuery.data.length) resourceQuery.refetch()
-    }, [resourceQuery.data])
+  },
+  useGenerateDimensions: async (files: readonly FileBrowserContentType[]) => {
+    useGenerateHelper(
+      files,
+      (file) => (file.type === 'glf' || file.type === 'glb' ? file.key : undefined),
```
dinomut1: typo, need file.type === 'gltf' not 'glf'

packages/client-core/src/common/services/FileThumbnailJobState.tsx
```diff
@@ -170,89 +212,103 @@ const uploadThumbnail = async (src: string, projectName: string, blob: Blob | nu
...
+  filterKey: (file: FileBrowserContentType) => string | undefined,
+  queryConditions: Record<string, any>,
+  jobType: 'thumbnail' | 'dimension' = 'thumbnail'
+) => {
+  const jobState = useMutableState(FileThumbnailJobState)
+  const seenResources = jobState.seenResources[jobType]
+  const fileList = files
+    .map(filterKey)
+    .filter((key): key is string => key !== undefined)
+    .filter((key) => !seenResources.value.includes(key))
```
dinomut1, line 235: both thumbnail and dimension jobs are sharing `seenResources`. They need to have different seenResources objects for each job type so they don't get confused

dinomut1 (changes requested):

dinomut1 (approved):

Rezmason (dismissed):
Echoing what David's said, we need to either separate the processes of FileThumbnailJobState and dimension evaluation, or, if there's a good reason for them to occur simultaneously, then to rename the shared elements and encapsulate the task-specific elements.

JT00y:
> Echoing what David's said, we need to either separate the processes of FileThumbnailJobState and dimension evaluation, or, if there's a good reason for them to occur simultaneously, then to rename the shared elements and encapsulate the task-specific elements.
dimension basically have the same way as FileThumbnailJobState, new JobState will introduce many duplicated work

Rezmason (approved):

PR #1507, opened 2025-03-26
RELEASE Updated the generated thumbnail name to use a hashed value

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-7534

## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/src/common/services/FileThumbnailJobState.tsx
+++ b/packages/client-core/src/common/services/FileThumbnailJobState.tsx
@@ -82,14 +82,13 @@ import mime from 'mime-types'
 import { uploadToFeathersService } from '../../util/upload'
 import { getCanvasBlob } from '../utils'
 
-const getFilenameWithoutExtension = (url) => {
+const getFilename = (url) => {
   const path = new URL(url).pathname // Get the path part of the URL
-  const filename = path.substring(path.lastIndexOf('/') + 1) // Get the filename after the last "/"
-  return filename.substring(0, filename.lastIndexOf('.')) // Remove the file extension
+  return path.substring(path.lastIndexOf('/') + 1) // Get the filename after the last "/"
 }
 
 export function generateThumbnailKey(src: string, projectName: string) {
-  const uniqueFileName = `${projectName}-${getFilenameWithoutExtension(src)}-${Date.now()}`
+  const uniqueFileName = `${projectName}-${getFilename(src)}-${Date.now()}`
   const encoder = new TextEncoder()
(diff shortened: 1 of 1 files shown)


packages/client-core/src/common/services/FileThumbnailJobState.tsx
```diff
@@ -77,15 +77,24 @@ import { SkyboxComponent } from '@ir-engine/engine/src/scene/components/SkyboxCo
...
+  const filename = path.substring(path.lastIndexOf('/') + 1) // Get the filename after the last "/"
+  return filename.substring(0, filename.lastIndexOf('.')) // Remove the file extension
+}
+
 export function generateThumbnailKey(src: string, projectName: string) {
-  return `${decodeURI(stripSearchFromURL(src).replace(/^.*?\/projects\//, ''))
-    .replace(projectName + '/', '')
-    .replaceAll(/[^a-zA-Z0-9\.\-_\s]/g, '_')
-    .replaceAll(/\s/g, '-')}-thumbnail.png`
+  const uniqueFileName = `${projectName}-${getFilenameWithoutExtension(src)}-${Date.now()}`
```
SamMazerIR: does removing the extension for the hashing function make sense? why not just leave it and avoid issues with files that have the same name/path but different extensions?

packages/client-core/src/common/services/FileThumbnailJobState.tsx
```diff
@@ -77,15 +77,24 @@ import { SkyboxComponent } from '@ir-engine/engine/src/scene/components/SkyboxCo
...
+  const encoder = new TextEncoder()
+  const buffer = encoder.encode(uniqueFileName)
+  let hash = createHash('sha256').update(buffer).digest('hex')
+  hash = hash.slice(0, 46) // Ensuring max length constraint with VALID_FILENAME_REGEX
+  return `${hash}.png`
```
SamMazerIR, line 96: double-checking..do we convert all images to .png? we support multiple formats but I'm not sure if we convert on upload

packages/client-core/src/common/services/FileThumbnailJobState.tsx
MbfloydIR: updated

SamMazerIR (approved):

packages/client-core/src/common/services/FileThumbnailJobState.tsx
HexaField: only .png for now

aditya-mitra (approved):

PR #1510, opened 2025-03-26
update the importing of the glTF files to have the "Occlude Camera" feature of the glTF component set to True

MbfloydIR:
update the importing of the glTF files to have the "Occlude Camera" feature of the glTF component set to True

## Summary
https://tsu.atlassian.net/browse/IR-8543
## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/functions/addMediaNode.ts
+++ b/packages/editor/src/functions/addMediaNode.ts
@@ -218,7 +218,7 @@ export async function addMediaNode(
     } else {
       const { entityUUID } = EditorControlFunctions.createObjectFromSceneElement(
         [
-          { name: GLTFComponent.jsonID, props: { src: url, cameraOcclusion: true } },
+          { name: GLTFComponent.jsonID, props: { src: url } },
           { name: ShadowComponent.jsonID },
           { name: EnvMapComponent.jsonID },
           ...extraComponentJson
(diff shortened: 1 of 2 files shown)


packages/editor/src/functions/addMediaNode.ts
```diff
@@ -218,7 +218,7 @@ export async function addMediaNode(
     } else {
       const { entityUUID } = EditorControlFunctions.createObjectFromSceneElement(
         [
-          { name: GLTFComponent.jsonID, props: { src: url } },
+          { name: GLTFComponent.jsonID, props: { src: url, cameraOcclusion: true } },
```
HexaField: i reckon we should set it to true by default on the component itself honestly

HexaField (changes requested):

packages/editor/src/functions/addMediaNode.ts
SamMazerIR: @HexaField its being asked for by content, as there is no way to save this to the prefab itself - only to the scene. Already went through Ross

HexaField: yeah what i am saying is that the `cameraOcclusion` property of the GLTFComponent is false by default, i think it sohuld be true by default

HexaField (approved):

SamMazerIR (approved):

MichaelEstes (approved):

PR #1513, opened 2025-03-27
Classify iPads as mobiles

CITIZENDOT:
## Summary

iPad Air & iPad Pro uses the same user agent as MacOS, so they're classified as desktops now.
This PR adjusts `isMobile` check to classify iPad Air & iPad Pro as mobile devices.

## References
closes https://tsu.atlassian.net/browse/IR-8418


the-reality-engineer (changes requested):
What problem is this fixing? Overly relying on device classification is an anti-pattern. If we need to detect touch-screen capability, then we should do that. `mobile` vs `desktop` doesn't really mean anything. Any device type can potentially have a touch-screen.

CITIZENDOT:
> What problem is this fixing

- iPads are not detected through isMobile check.
- Currently, TouchPad is shown to mobiles only (devices with isMobile as true). But since iPads are not passing that check, so TouchPad is not visible. This PR solves that.

> If we need to detect touch-screen capability, then we should do that

Makes sense, We can create a new conditional (`hasTouchScreen`), that checks only for `maxTouchPoints` >= 1, and use it for touch pad.

packages/spatial/src/input/components/InputSourceComponent.tsx
```diff
@@ -70,7 +70,7 @@ export const InputSourceComponent = defineComponent({
             hapticActuators: [],
             id: 'emulated-gamepad-' + entity,
             index: 0,
-            mapping: isMobile ? 'xr-standard' : '',
+            mapping: isTouchAvailable ? 'xr-standard' : '',
```
the-reality-engineer, line 73: Are we using this mapping to control whether or not the touchpad appears? Seems kind of arbitrary. Can we keep it set to "xr-standard" and manage the touchpad separately?

CITIZENDOT: > Are we using this mapping to control whether or not the touchpad appears

No. Touchpad appears on iPad pro, but it doesn't work (at least that's what I saw on emulated iPad). Dragging touch pad controls does not move the avatar.

I believe it's happening because `gamepad.mapping` controls how the buttons behave. In this case, iPad's mapping is not set, hence Touchpad controls are not working.

https://github.com/ir-engine/ir-engine/blob/a32ea5060e8fbf825fba2c3a803c5f523bd12e04/packages/spatial/src/input/state/ButtonState.ts#L265-L276


This solution works on emulated iPad. Although, I need to confirm whether this is working on real device.

DanielBelmes (approved):

DanielBelmes:
Sadly breaking unit tests due to the assumptuion of a default value of ''

DanielBelmes (changes requested):
unit tests

DanielBelmes (approved):

DanielBelmes:
Annoyingly didn't make it to dev done 1.0.1. Please port to 1.0.2

CITIZENDOT:
Closing in favour of https://github.com/ir-engine/ir-engine/pull/1750

PR #1514, opened 2025-03-27
hotfix(release1.0.2): prevent modal from being closed on click outside

aditya-mitra:
## Summary

prevent modal from closing when clicked outside

## References

https://tsu.atlassian.net/browse/IR-8366


CITIZENDOT (approved):

DanielBelmes (changes requested):
This makes a lot of modals unable to be exited out of. For example the profile modal in location. Which does need a close button for mobile. I think the best bet is to get some close buttons for all of these modals. Share, Settings, Profile, etc...

aditya-mitra:
> This makes a lot of modals unable to be exited out of. For example the profile modal in location. Which does need a close button for mobile. I think the best bet is to get some close buttons for all of these modals. Share, Settings, Profile, etc...

I think only the profile modal required a close button because it left minimal space to click outside (on smaller screens). You are right - adding this functionality does not allow some menus to be dismissed.

Importantly we cannot place a close icon on the **circular emote menu**.

Sam added a close button in profile menu in https://github.com/ir-engine/ir-engine/pull/1598. Hence, I think we can close this PR and ticket.

@DanielBelmes what are your thoughts?

DanielBelmes (changes requested):
I think we should find a way to just make the publish flow not have a click away option. Instead of applying to all modals.

aditya-mitra:
Fixed with the new changes!

https://github.com/user-attachments/assets/846f459c-4d24-458d-8dce-7369f1bc1600

Also requires https://github.com/theinfinitereality/irpro-multitenancy/pull/939 to be merged

SYBIOTE (approved):

kattsushi (approved):

PR #1529, opened 2025-03-28
Use useExecute in entity properties to keep components list updated

HexaField:
## References
closes #_insert number here_


aditya-mitra (approved):
works fine

PR #1542, opened 2025-03-29
Remove TransformGizmoTagComponent to prevent interference with heuristic access

SYBIOTE:
Eliminate the TransformGizmoTagComponent to avoid conflicts with heuristic


SYBIOTE:
wrong commit to wrong branch

PR #1543, opened 2025-03-29
Remove TransformGizmoTag from gizmoPlane to avoid heuristic interference

SYBIOTE:
Eliminate the TransformGizmoTag component from the gizmoPlane to prevent conflicts with heuristics.


SYBIOTE:
only commented out with the reason and have not removed it so its not added in the future

PR #1547, opened 2025-03-31
Hoist AssetsQueryProvider

Tortillaguy:
## Summary
 Because the AssetsQueryProvider was mounted so low in the DOM tree, only the assets panel had reference to it. 
 The SavePrefabDialog is mounted by another part of the DOM tree, and could not call the `refreshResources` method for updating state.
 
 The solution here is to move the AssetsQueryProvider up the DOM tree to a common ancestor between the assets panel and the hierarchy viewer.

## References
closes [IR-8558]

## QA Steps

[IR-8558]: https://tsu.atlassian.net/browse/IR-8558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/editor/src/components/EditorContainer.tsx
+++ b/packages/editor/src/components/EditorContainer.tsx
@@ -56,6 +56,7 @@ import { useTranslation } from 'react-i18next'
 import { IoHelpCircleOutline } from 'react-icons/io5'
 import { onSaveScene, setCurrentEditorScene } from '../functions/sceneFunctions'
 import { AssetsPanelTab } from '../panels/assets'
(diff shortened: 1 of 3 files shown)


HexaField (dismissed):
i'm wondering why we are using providers at all when hypeflux states can do the same pattern without tying it to the DOM

cc @aditya-mitra @CITIZENDOT @DanielBelmes

aditya-mitra (approved):
LGTM

PR #1552, opened 2025-04-01
Remove TransformGizmoTagComponent from gizmo plane heuristic consideration

SYBIOTE:
Eliminate the consideration of TransformGizmoTagComponent in the gizmo plane heuristic to streamline entity queries for gizmo entities.


SamMazerIR (approved):

SamMazerIR:
Just wanted to point out that this may affect the ClientInputSystem from being able to interact with the gizmo plane...not sure if that is the intention though

SYBIOTE:
yeah, i dont think we even want the client Input system to interact with the gizmo plane, in the end its a gizmo helper, and it runs based on the raycast used for the gizmo, should not interfere with existing gizmo functions

PR #1553, opened 2025-04-01
[IR-7095] Replace usage of is editor and is editing with layer api

MichaelEstes:
## References
closes #_insert number here_

### DIFF
--- a/packages/ecs/src/ComponentFunctions.ts
+++ b/packages/ecs/src/ComponentFunctions.ts
@@ -1292,6 +1292,10 @@ export function getAuthoringCounterpart(entity: Entity) {
   return LayerComponents[Layers.Simulation].refs[entity] ?? UndefinedEntity
 }
 
+export function hasAuthoring(entity: Entity) {
+  return !!getAuthoringCounterpart(entity)
+}
+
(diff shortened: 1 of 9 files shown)


packages/ecs/src/ComponentFunctions.ts
```diff
@@ -1271,6 +1274,16 @@ export const LayerComponent = defineComponent({
   }
 })
 
+export function useLayerCounterpart(entity: Entity, layer: LayerID) {
```
HexaField, line 1277: layers are not mutable, so there is no need to store this in reactive state, and since all entities will have a layer, this does get a little expensive in terms of memory. we can assume layers are always static.

HexaField (changes requested):

packages/ecs/src/ComponentFunctions.ts
MichaelEstes: Oh going in and out of playmode doesn't add/remove the authoring layer?

HexaField: nope :) and if it did, the entity layer won't change, it'd destroy/re-create the entities

PR #1554, opened 2025-04-01
add opaque types to ObjectLayers and ObjectLayerMasks

glenn-gartner-ir:
## Summary
Opaque type is a more strict type used within the engine. This makes it so that something not passed with this type will throw a typescript error. This helps with bugs and troubleshooting the development experience. 

## References
closes [IR-7904](https://tsu.atlassian.net/browse/IR-7904)

## QA Steps

[IR-7904]: https://tsu.atlassian.net/browse/IR-7904?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/spatial/src/renderer/constants/ObjectLayers.ts
```diff
@@ -22,55 +22,59 @@ Original Code is the Infinite Reality Engine team.
 All portions of the code written by the Infinite Reality Engine team are Copyright © 2021-2023 
 Infinite Reality Engine. All Rights Reserved.
 */
+import { Opaque } from '@ir-engine/common/src/interfaces/OpaqueType.ts'
```
HexaField: we actually have a duplicate of the opaque type definition, and we need to use the one from hyperflux

HexaField: also, looks like the import has the file extension which can cause problems

HexaField (changes requested):

HexaField:
Looks like LineSegmentComponent needs to be adjusted to this change still, as per the `check-errors` check

packages/spatial/src/renderer/components/LineSegmentComponent.ts
```diff
@@ -72,7 +72,7 @@ export const LineSegmentComponent = defineComponent({
     }, [component.name])
 
     useEffect(() => {
-      ObjectLayerMaskComponent.setMask(entity, component.layerMask.value)
+      ObjectLayerMaskComponent.setMask(entity, component.layerMask.value as ObjectLayerMask)
```
HexaField: rather than casting here, we should change the type of the schema to reflect this. i dont know if we have a number schema definition for this, but i think an enum might work?

HexaField (changes requested):

packages/spatial/src/renderer/components/LineSegmentComponent.ts
glenn-gartner-ir: agreed. I looked into a variation of TypedString, something like TypedNumber. It got a little abstract, but I'll push a potential solution. Perhaps enum will be simpler.

HexaField: might make less sense to be as strict with a mask, since the individual masks are 2^x but any mask usage would be any integer as a combination of individual masks, so types become more of a sanity check than actually definiting bounds of the value

glenn-gartner-ir: Sure. What about making the layerMask schema an 
```ts
S.Required(S.Type<ObjectLayerMask>())? 
```
Or, alternatively, ```S.Optional...```, if it's truly optional?

packages/spatial/src/renderer/components/LineSegmentComponent.ts
```diff
@@ -45,7 +45,7 @@ export const LineSegmentComponent = defineComponent({
     geometry: S.Required(S.Type<BufferGeometry>()),
     material: S.Class(() => new LineBasicMaterial() as Material),
     color: S.Optional(T.Color()),
-    layerMask: S.Number(ObjectLayers.NodeHelper)
+    layerMask: S.Required(S.Type<ObjectLayerMask>())
```
HexaField: should still default to the node helper ideally

glenn-gartner-ir: Just specifying the type with a default value seems good
```ts 
S.Type<ObjectLayerMask>(ObjectLayerMasks.NodeHelper) 
```
The S.Required() breaks setting the component in other areas of the code that aren't passing a layerMask. Making it optional, while passing a default object seems strange. So, S.Type<> seems alright. 
Sanity check that my using ObjectLayerMasks.NodeHelper (the mask) instead of the original ObjectLayers.NodeHelper (the layer) is what's intended.

HexaField (approved):
awesome

glenn-gartner-ir:
@HexaField failing the 30% code coverage for tests. Should we test for types?

HexaField:
> @HexaField failing the 30% code coverage for tests. Should we test for types? 

Nah, sonarqube is a bit over the top. This is fine as it is.

PR #1558, opened 2025-04-01
Tcr/ir 8618 add test i ds for studio editor

TCRodriguez:
## Summary
This PR brings in/updates `data-testid` attributes in the studio editor. It also adds a few more unit test files to preserve the majority of these data-testids.

## Subtasks Checklist
N/A

## Breaking Changes
N/A

## References
closes https://tsu.atlassian.net/browse/IR-8618

## QA Steps
N/A

### DIFF
--- a/packages/editor/package.json
+++ b/packages/editor/package.json
@@ -17,7 +17,7 @@
   "npmClient": "npm",
   "scripts": {
     "check-errors": "tsc --noemit",
-    "test": "cross-env TEST=true vitest run --config=../../vitest.client.config.ts"
+    "test": "cross-env TEST=true vitest run --config=./vitest.editor.config.ts"
   },
   "resolutions": {
     "@types/react": "18.2.0",
--- a/packages/editor/src/panels/assets/CategoriesList.test.tsx
+++ b/packages/editor/src/panels/assets/CategoriesList.test.tsx
@@ -1,0 +1,175 @@
+/*
+CPAL-1.0 License
+
+The contents of this file are subject to the Common Public Attribution License
+Version 1.0. (the "License"); you may not use this file except in compliance
+with the License. You may obtain a copy of the License at
+https://github.com/ir-engine/ir-engine/blob/dev/LICENSE.
+The License is based on the Mozilla Public License Version 1.1, but Sections 14
+and 15 have been added to cover use of software over a computer network and 
+provide for limited attribution for the Original Developer. In addition, 
+Exhibit A has been modified to be consistent with Exhibit B.
+
+Software distributed under the License is distributed on an "AS IS" basis,
+WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for the
+specific language governing rights and limitations under the License.
+
+The Original Code is Infinite Reality Engine.
+
+The Original Developer is the Initial Developer. The Initial Developer of the
+Original Code is the Infinite Reality Engine team.
+
+All portions of the code written by the Infinite Reality Engine team are Copyright © 2021-2023 
+Infinite Reality Engine. All Rights Reserved.
+*/
+
+import { cleanup, fireEvent, render, screen } from '@testing-library/react'
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import React from 'react'
+import { EditorState } from '../../services/EditorServices'
+import { FilesState } from '../../services/FilesState'
+import CategoriesList from './categories'
+
+function CategoriesListTestWrapper() {
+  const [selected, setSelected] = React.useState<string | undefined>(undefined)
+
+  const handleClick = (category?: string) => {
+    setSelected(category)
+  }
+
+  return <CategoriesList selected={selected} onClick={handleClick} />
+}
+
+describe('CategoriesList component', () => {
+  beforeAll(() => {
+    vi.mock('@ir-engine/hyperflux', async (importOriginal) => {
+      const actual = await importOriginal()
+      return {
+        // @ts-ignore
+        ...actual,
+        useHookstate: vi.fn().mockReturnValue({
+          value: false,
+          set: vi.fn()
+        }),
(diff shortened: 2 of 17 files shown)


aditya-mitra (approved):
Looks good to me!

heysokam (commented):
Leaving some post-merge review notes.

Some of them are just questions.
But I'm posting my internal notes as they are just for awareness, since the PR is already merged.
```md
- [ ] `@ts-ignore` should be avoided
      `@ts-expect-error` will warn/error when the tag is no longer needed (which is more desirable)

- [ ] Readability warning: `vi.mock` is hoisted. It affects the -entire- file.
      Nesting calls to `vi.mock` is confusing for readability,
      it makes it unclear what the call is affecting.
      _(in reality, it affects **all imports** from that file_
      _and any symbol used by that file accessed by the mocked import **anywhere** in the project)_
      @see The Mockery and The Local Hero Anti-Patterns
           https://stackoverflow.com/questions/333682/unit-testing-anti-patterns-catalogue

- [ ] Mocking the entire hyperflux import is questionable. Are we actually testing correct behavior this way?
      Specially when the mocking code is duplicated inside each describe.
      @see The Mockery Anti-Pattern
           https://stackoverflow.com/questions/333682/unit-testing-anti-patterns-catalogue
- [ ] Patching hyperflux for tests is questionable. Can it be avoided?

- [ ] Monolithic tests : Tests should be more granular
      They should test one thing each, not all of the cases all at once.
      _eg: `EditorDropdownItem.test.tsx` should have 4 unit tests, not 1_
      If a test fails, it should be because its one single condition changed.
      Multi condition tests could be hiding failures as "1 test failed, its fine",
      when in reality multiple conditions in the chain should all have failed instead.
      @see The Free Ride / Piggyback Anti-Pattern
           https://stackoverflow.com/questions/333682/unit-testing-anti-patterns-catalogue

- [ ] Why `vitest.setup.ts` -> `import '@testing-library/jest-dom/vitest'`
- [ ] Why a separate `vitest.editor.config.ts`?
```

packages/editor/src/panels/assets/CategoriesList.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const handleClick = (category?: string) => {
+    setSelected(category)
+  }
+
+  return <CategoriesList selected={selected} onClick={handleClick} />
+}
+
+describe('CategoriesList component', () => {
+  beforeAll(() => {
+    vi.mock('@ir-engine/hyperflux', async (importOriginal) => {
```
HexaField, line 46: this is a very bad idea - we want our tests to mimic functionality as directly as possible, and mocking entire modules (especially ours) is only going to lead to a divergence between the tested code and the testing code. we have very well established patterns for setting up the engine.

packages/editor/package.json
```diff
@@ -17,7 +17,7 @@
   "npmClient": "npm",
   "scripts": {
     "check-errors": "tsc --noemit",
-    "test": "cross-env TEST=true vitest run --config=../../vitest.client.config.ts"
+    "test": "cross-env TEST=true vitest run --config=./vitest.editor.config.ts"
```
HexaField, line 20: the client vitest config is supposed to be common across all client packages, creating a new one changes the behaviour between client packages and means we need to maintain multiple

packages/ui/src/components/tailwind/MoreOptionsMenu/MoreOptionsMenu.test.tsx
```diff
@@ -61,6 +61,7 @@ describe('MoreOptionsMenu component', () => {
 
   it('should render a button with data-testid "more-options-button"', () => {
     const moreOptionsButton = screen.getByTestId('more-options-button')
+    // @ts-ignore
```
HexaField, line 64: why use @ts-ignore? If it is because screen.getByTestId returns a potentially falsy value we should instead use the falsy supression syntax (`!`), otherwise @ts-expect-error if we expect an error (which does not seem to be the case here)

packages/editor/package.json
TCRodriguez: @HexaField This was to address an issue where we were getting errors in other packages from the `import '@testing-library/jest-dom/vitest'`. Having it go through the root `vitest.client.config.ts` seemed to lead to these issues. It was something @DanielBelmes had me do in order to isolate that import to just the relevant packages.

packages/ui/src/components/tailwind/MoreOptionsMenu/MoreOptionsMenu.test.tsx
TCRodriguez: @HexaField This was actually because of something that only started happening in this particular PR, and not my previous one with the console/dashboard work.

I was getting this error:
```
Property 'toBeInTheDocument' does not exist on type 'Assertion<HTMLElement>'.ts(2339)
```

At lines like this:
```
    expect(moreOptionsButton).toBeInTheDocument()
```

All the tests were passing when ran locally, but when I tried pushing up in a PR it was causing some of the branch tests to fail. I went with `ts-ignore` since I saw it is being used in plenty of other files in `ir-engine`. The `import '@testing-library/jest-dom/vitest'` module is what "extends" the `expect` that we get from `vitest` itself.

That being said, I can make that change over to `ts-expect-error` if necessary.

packages/editor/src/panels/assets/CategoriesList.test.tsx
TCRodriguez: @HexaField I see 🤔

All I was focused on was getting the component in question to render so that I could assert the existence of the relevant `data-testid` attributes, not any deeper functionality of the module. Admittedly I was working at the very edges of my understanding of how to bring in the component dependencies to get the component to render properly.

That being said, I can see that if this is a concern, it'd need to be changed to accommodate any new `it` blocks/tests that are actually focused on testing deeper module functionality.

PR #1564, opened 2025-04-02
add loadMaterialGLTF function

JT00y:
https://tsu.atlassian.net/browse/IR-8626


HexaField (changes requested):
doing this at the GLTF loader level is not a good idea. this should a higher level functionality that picks out the material once the whole document is loaded

packages/engine/src/gltf/GLTFLoaderFunctions.ts
dinomut1, line 1: don't adjust the gltf loader to handle a document that is an array. That is an invalid GLTF file. Instead, fix the gltf file so it is not an array

packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -1545,7 +1545,21 @@ const loadNode = async (options: GLTFParserOptions, nodeIndex: number) => {
 
   return nodeEntity
 }
+const loadMaterialGLTF = async (options: GLTFParserOptions, sceneIndex: number) => {
+  const json = Array.isArray(options.document) ? options.document[0] : options.document
```
dinomut1: Same thing here. Arrays are not valid GLTF files.

packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -1545,7 +1545,21 @@ const loadNode = async (options: GLTFParserOptions, nodeIndex: number) => {
...
 
+  // load deltas into state before anything else
+  const deltas = json.extensions?.[SCENE_DELTA_EXTENSION_NAME] as SceneDeltaRegistry | null
+  if (deltas) {
+    const parsedDeltas = parseStorageProviderURLs(deltas)
+    getMutableState(SceneDeltaState).merge(parsedDeltas)
+  }
```
dinomut1: Don't support deltas on material assets, as they will never have deltas saved into them

packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -1545,7 +1545,21 @@ const loadNode = async (options: GLTFParserOptions, nodeIndex: number) => {
...
+  DependencyCache.set(options.url, new Map())
+  const pending = [] as Promise<Entity>[]
+  pending.push(getDependency(options, 'material', 0) as Promise<Entity>)
+  await Promise.all(pending)
+  getComponent(options.entity, GLTFComponent).body = null
```
dinomut1: Don't call getDependency. Just loop through the gltf document's "materials" field and call "loadMaterial" for each index

dinomut1 (changes requested):

dinomut1:
> doing this at the GLTF loader level is not a good idea. this should a higher level functionality that picks out the material once the whole document is loaded

The gltf loader does not load materials if they are not referenced by a node in the file. Rather than add a dud node, with a dud mesh, with a dud primitive, referencing the material (very sloppy), we just need to load the materials directly, using a simpler load function than loadGLTFScene

HexaField:
> > doing this at the GLTF loader level is not a good idea. this should a higher level functionality that picks out the material once the whole document is loaded
> 
> The gltf loader does not load materials if they are not referenced by a node in the file. Rather than add a dud node, with a dud mesh, with a dud primitive, referencing the material (very sloppy), we just need to load the materials directly, using a simpler load function than loadGLTFScene

we should change it to do that then, rather than adding a whole other workflow in the gltf loader. i've already changed this for the exporter

dinomut1:
> > > doing this at the GLTF loader level is not a good idea. this should a higher level functionality that picks out the material once the whole document is loaded
> > 
> > 
> > The gltf loader does not load materials if they are not referenced by a node in the file. Rather than add a dud node, with a dud mesh, with a dud primitive, referencing the material (very sloppy), we just need to load the materials directly, using a simpler load function than loadGLTFScene
> 
> we should change it to do that then, rather than adding a whole other workflow in the gltf loader. i've already changed this for the exporter

This seems like a much more complex solution than just loading materials directly using a smaller, simpler function when we know there are only materials in the file. I'm not seeing where this introduces more complexity, it's not a whole other workflow. It's a much simpler subset of the normal load function, using the exact same helper functions, constrained to the "materials" field of the gltf

HexaField:
> This seems like a much more complex solution than just loading materials directly using a smaller, simpler function when we know there are only materials in the file. I'm not seeing where this introduces more complexity, it's not a whole other workflow. It's a much simpler subset of the normal load function, using the exact same helper functions, constrained to the "materials" field of the gltf

adding a special condition for our custom material gltf files i would categorize as "more complexity". it's an edge case - one for a very specific use case, that can be mainlined into the main GLTF loader. i am literally suggesting a simpler solution here...

dinomut1:
> > This seems like a much more complex solution than just loading materials directly using a smaller, simpler function when we know there are only materials in the file. I'm not seeing where this introduces more complexity, it's not a whole other workflow. It's a much simpler subset of the normal load function, using the exact same helper functions, constrained to the "materials" field of the gltf
> 
> adding a special condition for our custom material gltf files i would categorize as "more complexity". it's an edge case - one for a very specific use case, that can be mainlined into the main GLTF loader. i am literally suggesting a simpler solution here...

Well this is a special condition. We're not following the GLTF spec by just loading the materials. You're suggesting we expand "loadScene" to load parts of the GLTF file that are not in the scene. That would make our GLTF loader behave differently than the GLTF spec, which allows multiple scenes to exist in a single file. When each scene is loaded, we're only supposed load the contents of that scene.

dinomut1:
We should keep loadScene the way it is, so that it only loads the dependency tree of assets that are linked to the nodes included in the scene we are loading. We introduce our own, strictly material loading function, which loads all the materials in a file without them being referenced by any scene. This is the simplest solution, and preserves the behavior of loadScene

the-reality-engineer:
I can see the value of both perspectives here; my two cents is that a single api surface is easier to maintain than an introducing a new api to handle an edge case. 

What does "loading a gltf material" imply here? It just means the material becomes referencable in our ECS, right? That doesn't imply that unnecessary resources will be downloaded or uploaded to the GPU, but I can imagine it being very useful to have all materials in a GLTF be referencable so that it's easy to swap materials or whatever without having to load up the GLTF again through a separate loading function.

HexaField:
> Well this is a special condition. We're not following the GLTF spec by just loading the materials. You're suggesting we expand "loadScene" to load parts of the GLTF file that are not in the scene. That would make our GLTF loader behave differently than the GLTF spec, which allows multiple scenes to exist in a single file. When each scene is loaded, we're only supposed load the contents of that scene.

But all of this is only required because you are wanting to load gltf files that only have materials, which by your logic is already going against the gltf spec? So my point is, if this is not the behaviour we want for the GLTF loader, it should be an extension or something and we shouldn't overload the responsibility of the gltf loader with a custom file extension.

HexaField:
After speaking with Gheric it seems like the whole GLTF file should always be `loaded` and treated as a library file, it's just that the 'scene' property specifies what should be `rendered`. This should solve all these problems and allow us to avoid conditional logic based on the file extensions.

https://kcoley.github.io/glTF/specification/2.0/#scenes

In the future, we may be able to optimize this further to only load resources if we deem them necessary to be loaded, but for now the GLTF spec seems to imply that everything should be loaded, just not rendered. This will make it easier to do LODding of resources too.

dinomut1:
If we want to load all of the contents of a GLTF file, that is itself a different function from our "loadScene" function that we have right now. loadScene behaves exactly how it is supposed to, only instantiating ECS data relating to what is in the specified scene. Also, in our engine, the act of loading meshes and materials into the ECS immediately uploads it to the GPU anyways.

This isn't "overloading" our GLTF loader. We don't have a GLTF loader. We have a very modular set of functions which parse JSON data from GLTF files. This PR introduces nothing other than a specific function to just load materials, consisting of a for loop that calls only the functions to load the materials. It will not make anything harder to maintain, and does nothing to the existing flow, other than add a conditional handler for .material.gltf files, which we already need anyways to handle material assets.

A function that loads the entirety of the GLTF file immediately has repercussions beyond the scope of this PR. I'm inclined to merge the implementation that is already here, and if we want to refactor our system to load the entire file, that should be a separate PR that introduces a new loadGLTF function that loads every scene, and every resource, even if it is not in a scene.

packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -1545,7 +1545,18 @@ const loadNode = async (options: GLTFParserOptions, nodeIndex: number) => {
 
   return nodeEntity
 }
+const loadMaterialGLTF = async (options: GLTFParserOptions, sceneIndex: number) => {
```
dinomut1: don't need the scene index

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -250,18 +250,34 @@ export const GLTFComponentReactor = () => {
...
+    const hashUrl = new URL(url)
+    hashUrl.search = ''
+    const unhashUrl = hashUrl.href
+    if (unhashUrl.endsWith('.material.gltf')) {
+      GLTFLoaderFunctions.loadMaterialGLTF(options, sceneIndex).then(() => {
+        documentLoaded.set(true)
+
+        // force transform update for all entities in the model.
+        // required to propagate dirty update auth to sim layers
+        TransformComponent.dirty[entity] = 1
```
dinomut1: don't need to set dirty transforms

dinomut1 (changes requested):

HexaField:
> If we want to load all of the contents of a GLTF file, that is itself a different function from our "loadScene" function that we have right now. loadScene behaves exactly how it is supposed to, only instantiating ECS data relating to what is in the specified scene. 

Yes, that was what my incorrect assumption had been previously, we just established this is not what GLTF specifies.

> Also, in our engine, the act of loading meshes and materials into the ECS immediately uploads it to the GPU anyways.
That's not how it works, they are uploaded by the threejs renderer once they are rendered. If they are loaded into the ECS but not in a hierarchy that is specified by the RendererComponent's `scenes` property, then they will not get uploaded. We do occasionally use `initTexture` but that is for specific use cases.

> This isn't "overloading" our GLTF loader. We don't have a GLTF loader. We have a very modular set of functions which parse JSON data from GLTF files. This PR introduces nothing other than a specific function to just load materials, consisting of a for loop that calls only the functions to load the materials. It will not make anything harder to maintain, and does nothing to the existing flow, other than add a conditional handler for .material.gltf files, which we already need anyways to handle material assets.

I disagree, this is a system that loads a GLTF file - hence GLTF loader. What I am saying that conditional logic for just loading materials is already part of the GLTF spec, and we should adhere to it, which also removes the constraint of requiring the '.material.gltf' extension.

> A function that loads the entirety of the GLTF file immediately has repercussions beyond the scope of this PR. I'm inclined to merge the implementation that is already here, and if we want to refactor our system to load the entire file, that should be a separate PR that introduces a new loadGLTF function that loads every scene, and every resource, even if it is not in a scene.

Yes, which is why I am saying this solution is adding tech debt unnecessarily, and we should do that other work instead, which will have a better result overall.

If you look at the threejs loader, it does exactly this - it loads in ALL scenes, and gives back the 'main' scene as the `scene` property.

dinomut1:
> > If we want to load all of the contents of a GLTF file, that is itself a different function from our "loadScene" function that we have right now. loadScene behaves exactly how it is supposed to, only instantiating ECS data relating to what is in the specified scene.
> 
> Yes, that was what my incorrect assumption had been previously, we just established this is not what GLTF specifies.
> 
> > Also, in our engine, the act of loading meshes and materials into the ECS immediately uploads it to the GPU anyways.
> > That's not how it works, they are uploaded by the threejs renderer once they are rendered. If they are loaded into the ECS but not in a hierarchy that is specified by the RendererComponent's `scenes` property, then they will not get uploaded. We do occasionally use `initTexture` but that is for specific use cases.
> 
> > This isn't "overloading" our GLTF loader. We don't have a GLTF loader. We have a very modular set of functions which parse JSON data from GLTF files. This PR introduces nothing other than a specific function to just load materials, consisting of a for loop that calls only the functions to load the materials. It will not make anything harder to maintain, and does nothing to the existing flow, other than add a conditional handler for .material.gltf files, which we already need anyways to handle material assets.
> 
> I disagree, this is a system that loads a GLTF file - hence GLTF loader. What I am saying that conditional logic for just loading materials is already part of the GLTF spec, and we should adhere to it, which also removes the constraint of requiring the '.material.gltf' extension.
> 
> > A function that loads the entirety of the GLTF file immediately has repercussions beyond the scope of this PR. I'm inclined to merge the implementation that is already here, and if we want to refactor our system to load the entire file, that should be a separate PR that introduces a new loadGLTF function that loads every scene, and every resource, even if it is not in a scene.
> 
> Yes, which is why I am saying this solution is adding tech debt unnecessarily, and we should do that other work instead, which will have a better result overall.
> 
> If you look at the threejs loader, it does exactly this - it loads in ALL scenes, and gives back the 'main' scene as the `scene` property.

loadScene loads the scene, that is not an incorrect implementation. That gltf spec does not say anything about loading a scene requiring you to load everything in the file. You're requesting a different function, which, again, introduces complications beyond the scope of this PR. If we load everything in the file, that doesn't just mean materials, it includes animations, buffers, meshes etc. that we have no way of defining without nodes. I could easily make a valid argument that we should have specific functions for loading each segment of a gltf file for specialized purposes (loading geometry, animations, materials, images, etc.), and it would not be tech debt. The difference in control flow and where the conditional logic goes is frankly arbitrary.

This is 30 lines of code that unblocks Tong from continuing to build our material library, using simple logic that has no side effects anywhere. She is now continuing with her original task of using this to create additional materials. If we want to further discuss this larger request you are making, which introduces other considerations that have no immediate answer and require additional implementations, we can do it in another PR

dinomut1 (approved):

PR #1567, opened 2025-04-02
Inherit layer from parent in createSceneEntity

dinomut1:
Dependency for https://github.com/theinfinitereality/irpro-builder-frontend/pull/9


junhua-ir (approved):
Tested with World Builder and Frontend changes and confirmed working.

PR #1568, opened 2025-04-02
Michael/ir 8293 trigger particles 1.02

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-8293
## References
closes #_insert number here_


packages/engine/src/scene/components/ParticleSystemComponent.ts
```diff
@@ -1013,6 +1013,7 @@ export const ParticleSystemComponent = defineComponent({
 
     const sceneEntity = useAncestorWithComponents(entity, [SceneComponent])
     const visible = useHasComponent(entity, VisibleComponent)
+    const empty3dObj = new Object3D()
```
HexaField: this shouldn't be declared in this scope, doing this will create a new Object3D every time the reactor rerenders, adding more memory. it can just be passed in directly to the setComponent

HexaField (dismissed):

packages/engine/src/scene/components/ParticleSystemComponent.ts
MbfloydIR: implemented

packages/engine/src/scene/components/ParticleSystemComponent.ts
```diff
@@ -1066,6 +1066,8 @@ export const ParticleSystemComponent = defineComponent({
           }
         }
         removeComponent(entity, ObjectComponent)
+        setComponent(entity, ObjectComponent, new Object3D())
```
SamMazerIR, line 1069: should this be in a mount rather than a return?

MbfloydIR: no.
This is to re-add the ObjectComponent, since it was removed to clear the particle's 3D object reference.
But, the ObjectComponent is needed for it's visible/invisible callback functions.
So it is being added back in, with an empty new Object3D

SamMazerIR (approved):

PR #1573, opened 2025-04-03
IR-7206 user vol system fix 1.0.2 (hotfix)

SamMazerIR:
## Summary
User volume now adjusts volume of other user voice chat
Microphone volume now adjusts your outgoing voice volume for other users

added tooltips for each volume setting to better explain

## References
[https://tsu.atlassian.net/browse/IR-7206](https://tsu.atlassian.net/browse/IR-7206)


packages/client-core/src/systems/PositionalAudioSystem.tsx
```diff
@@ -95,7 +95,7 @@ const execute = () => {
    */
   const networkedAvatarAudioEntities = networkedAvatarAudioQuery()
   for (const entity of networkedAvatarAudioEntities) {
-    if (!network?.peers) continue
+    if (!network) continue
```
HexaField: ?.peers is necessary as peers actually is nullish sometimes, even though it isn't typed as such

HexaField (dismissed):

packages/client-core/src/systems/PositionalAudioSystem.tsx
SamMazerIR: oh interesting, i may have removed it in error thinking it was breaking the logic before. Let me undo this and test

MbfloydIR (approved):

PR #1526, opened 2025-03-27
Debounce thumbnail checks and delay for server latency release 1.01

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-7830
## References
closes #_insert number here_

PR #1574, opened 2025-04-03
fix(IR-8529): avoiding refreshing constantly when thumbnails are being generated

kattsushi:
fix(IR-8529): avoiding refreshing constantly meanwhitle thumbnails are being generated
also prevent ui shifting on file browser loaders and improve ux

## Summary

https://github.com/user-attachments/assets/8223a8d6-4d63-43c2-9772-b372f6c68a73

## References
related to Jira ticket [IR-8529](https://tsu.atlassian.net/issues/IR-8529?filter=-1)

## QA Steps

[IR-8529]: https://tsu.atlassian.net/browse/IR-8529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/editor/src/panels/files/filebrowser.tsx
+++ b/packages/editor/src/panels/files/filebrowser.tsx
@@ -71,7 +71,11 @@ export function Browser() {
   }
 
   useEffect(() => {
-    refreshDirectory()
+    const timeoutId = setTimeout(() => {
+      refreshDirectory()
+    }, 3500) // Debounce refresh to prevent rapid re-renders
+
+    return () => clearTimeout(timeoutId)
   }, [thumbnailJobState.jobs.length])
 
   const staticResourceDataQuery = useFind(staticResourcePath, {
--- a/packages/editor/src/panels/files/loaders.tsx
+++ b/packages/editor/src/panels/files/loaders.tsx
@@ -118,7 +118,7 @@ export function ProjectDownloadProgress() {
   return (
     <>
       {isDownloading && (
-        <div className="flex h-auto w-full justify-center pb-2 pt-2">
+        <div className="flex h-auto w-full justify-center pb-1 pt-1">
           <div className="flex w-1/2">
             <span className="inline-block pr-2 text-xs font-normal leading-none ">
               {t('editor:layout.filebrowser.downloadingProject', { completed, total })}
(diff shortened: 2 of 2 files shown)


packages/editor/src/panels/files/filebrowser.tsx
```diff
@@ -71,7 +71,11 @@ export function Browser() {
   }
 
   useEffect(() => {
-    refreshDirectory()
+    const timeoutId = setTimeout(() => {
+      refreshDirectory()
+    }, 3500) // Debounce refresh to prevent rapid re-renders
+
+    return () => clearTimeout(timeoutId)
```
aditya-mitra, line 78: we need to clear the previous when a new timeout is entered.

this will refresh the file browser still but after 3.5 seconds.

packages/editor/src/panels/files/loaders.tsx
```diff
@@ -181,17 +181,22 @@ function FilesLoading() {
...
+      containerClassName="flex-row mt-1"
+      className="mx-2 my-auto h-6 w-6"
+    />
   ) : null
 }
 
 export default function Loaders() {
   return (
-    <>
+    <div className="flex h-[60px] flex-col py-1">
```
aditya-mitra, line 195: can you add _before_ and _after_ screenshots to show the differences?

kattsushi: sure this is the behavior before, was no only the loading flickering also having layout shifting in the ui of the progress bar


https://github.com/user-attachments/assets/0b0a3683-d3d5-4cdf-84a7-cbf61c75609b

packages/editor/src/panels/files/filebrowser.tsx
kattsushi: Probably we should check the number of jobs in the queue? For example, after processing 30 files, we could consider using the timeout what do you think ? as reference we are doing something similar in the release1.01 branch here https://github.com/ir-engine/ir-engine/pull/1526/files#diff-044f9e4bf8d5e7fcd596b1922afe0930a37b9de5b81d6c4948a3421b16e66f27R423

jose-galvan:
@MbfloydIR opened a PR for a similar issue, and it was merged against release1.0.1
https://github.com/ir-engine/ir-engine/pull/1526

kattsushi:
closed by https://github.com/ir-engine/ir-engine/pull/1526

PR #1575, opened 2025-04-03
[IR-8730] Unit Tests - GLTF Loading: Accessors

heysokam:
## References
https://tsu.atlassian.net/browse/IR-8730


packages/engine/src/gltf/loader/GLTFLoader.accessors.test.tsx
```diff
@@ -19,33 +19,126 @@ The Original Code is Infinite Reality Engine.
...
+   * @description Creates the most minimal gltf with one accessor possible, as required by spec
+   * */
+  function mockGLTFMinimal() {
+    const result = mockGLTF()
+    result.accessors = [
+      {
+        componentType: 5126, // FLOAT
+        count: 1,
+        type: 'SCALAR',
+        bufferView: undefined,
```
HexaField: be careful with specifying undefined properties, rather than not including them.

`const myobj = { prop: undefined }` will return true for  `'prop' in myobj`

heysokam: Fixed in 5837b09

heysokam:
@HexaField I not sure if the leftover `it.todo` tests for `accessor.min` and `accessor.max` are worth testing for now. But let me know if they are.
Other than that, it should be good to go when CI passes :writing_hand:

MichaelEstes (approved):

PR #1577, opened 2025-04-03
IR-8472: Populated context params from state for analytics events

hanzlamateen:
## Summary
It seems like in some analytics logging when creating logger object we were missing assigning modifier property which is used to populate values from state.

## References
https://tsu.atlassian.net/browse/IR-8427


aditya-mitra (approved):
LGTM

PR #1578, opened 2025-04-03
Avoid hideModal when dropdown is clicked

CITIZENDOT:
## References
closes https://tsu.atlassian.net/browse/IR-7449

### DIFF
--- a/packages/ui/src/primitives/tailwind/Select/index.tsx
+++ b/packages/ui/src/primitives/tailwind/Select/index.tsx
@@ -447,10 +447,6 @@ const Select = ({
               onMouseLeave={() => {
                 setActiveIndex(-1)
               }}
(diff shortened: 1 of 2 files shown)


packages/ui/src/primitives/tailwind/Select/index.tsx
```diff
@@ -462,6 +458,10 @@ const Select = ({
                 }
                 setTouchedMoved(false)
               }}
+              onPointerUp={(e) => {
+                e.stopPropagation()
+                e.preventDefault()
+              }}
```
aditya-mitra, line 464: imo, this is the only change that we were missing

(events were being propagated to clickawaylistener)

aditya-mitra (approved):
Works fine on mobile and desktop!

PR #1582, opened 2025-04-04
Tcr/ir 8724 add unit tests for hierarchy panel component

TCRodriguez:
## Summary
This PR adds a unit test file for the `HierarchyPanelTab` component, as well as implements the [comments left on previous PR](https://github.com/ir-engine/ir-engine/pull/1558) by @heysokam and @HexaField .

## Subtasks Checklist
N/A
## Breaking Changes
N/A
## References
closes https://tsu.atlassian.net/browse/IR-8724

## QA Steps
N/A


packages/editor/src/panels/assets/Resources.test.tsx
```diff
@@ -24,167 +24,106 @@ Infinite Reality Engine. All Rights Reserved.
...
-            }
-          }
-        })
-      }
-    })
-
-    vi.mock('@ir-engine/engine/src/assets/classes/AssetLoader', () => ({
-      AssetLoader: {
-        getAssetType: vi.fn().mockReturnValue('image')
+vi.mock('react-i18next', () => ({
```
heysokam, line 34: I vaguely remember we removing this on a different file, after adding createEngine, and the file working ok :thinking: But my memory might be failing me. Was this an oversight, or did you find it required for this file?

packages/editor/src/panels/files/FilesToolbar.test.tsx
```diff
@@ -23,201 +23,174 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
-}
-
-const dynamicViewMode = {
-  _value: 'icons',
-  get value() {
-    return this._value
-  },
-  set: vi.fn((newVal) => {
-    dynamicViewMode._value = newVal
+vi.mock('react-i18next', () => ({
```
heysokam, line 35: Same as the other comment. Pointing to it just in case of oversight.

packages/editor/src/panels/hierarchy/HierarchyPanelTab.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+import { SceneComponent } from '@ir-engine/spatial/src/renderer/components/SceneComponents'
+import { mockSpatialEngine } from '@ir-engine/spatial/tests/util/mockSpatialEngine'
+import { act, cleanup, fireEvent, render, screen } from '@testing-library/react'
+import React from 'react'
+import { Cache } from 'three'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { DndWrapper } from '../../../../editor/src/components/dnd/DndWrapper'
+import { HierarchyPanelTab } from './index'
+
+vi.mock('react-i18next', () => ({
```
heysokam, line 54: Same as before. Pointing to it just in case.

packages/editor/src/panels/hierarchy/HierarchyPanelTab.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+  useTranslation: () => ({
+    t: (key: string) => key
+  })
+}))
+
+vi.mock('@ir-engine/common/src/utils/FeathersHooks', () => ({
```
heysokam, line 60: Is there a way we could initialize feathers using the expected api, without having to mock its import?

This mock seems small, but feathers is very foundational to the stack, so I'm worried it might affect things that are not easy to find _(just like mocking hyperflux would have)_.

heysokam (commented):
Looks much better in general with the changes. Ty :writing_hand:
Left a few comments, in case we can improve those.

packages/editor/src/panels/files/FilesToolbar.test.tsx
TCRodriguez: Ah yes, I'm able to remove them and the tests run and pass just fine but I do get this `sdterr` in the console:
```
react-i18next:: You will need to pass in an i18next instance by using initReactI18next
```

If it's something we can just ignore in this case, then I can remove them. But that's what I had left them in @heysokam

packages/editor/src/panels/hierarchy/HierarchyPanelTab.test.tsx
TCRodriguez: @heysokam I see your point, but unless there is something like `createEngine` that we could use here I don't think there is?

That being said, I'm also hoping that we can go with this for now since these tests I've introduced aren't concerned with deeper functionality but just with the markup that gets rendered and that the `data-testid` attributes remain in place. It would take me much more time to figure out how to pull in all the requisite pieces instead of mocking them (i.e., more codebase/engine knowledge than I currently have). @larryatinfinite only tasked me with the `data-testid` assertion, and I don't believe I can currently take much more time to set this all up without the mocks.

But just to be clear, I do see your point as I did with mocking `hyperflux` -- definitely not something we should do when get to testing deeper functionality due to the divergence @HexaField mentioned in the last PR.

CITIZENDOT (approved):

HexaField:
Just noticed that this PR had disabled unit tests for the editor package :(

PR #1586, opened 2025-04-04
Skyxbox Component Unit Test

Tortillaguy:
## References
closes [IR-6390]

## QA Steps

[IR-6390]: https://tsu.atlassian.net/browse/IR-6390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/engine/src/scene/components/SkyboxComponent.test.ts
```diff
@@ -144,25 +144,33 @@ describe('SkyboxComponent', () => {
       expect(result.backgroundType).equals(SkyTypeEnum.equirectangular)
       expect(result.equirectangularPath).equals('https://picsum.photos/200.jpg')
 
-      const background = getOptionalComponent(entity, BackgroundComponent) as Texture
-      assert.exists(background)
+      let background
+      await vi.waitFor(() => {
```
HexaField: this can be simplified to `const background = await vi.waitUntil(() => getOptionalComponent(entity, BackgroundComponent) as Texture)`

HexaField (dismissed):

packages/engine/src/scene/components/SkyboxComponent.test.ts
Tortillaguy: good to know, that's a lot easier

packages/engine/src/scene/components/SkyboxComponent.ts
```diff
@@ -81,7 +81,7 @@ export const SkyboxComponent = defineComponent({
 
   reactor: function () {
     const entity = useEntityContext()
-    if (!isClient) return null
```
DanielBelmes, line 84: I tried looking into why this was here. Thought maybe it was some optimization for instance servers. Nope it was done during our initial MUI implementation 2 years ago. 🤷🏼‍♂️

DanielBelmes (approved):

packages/engine/src/scene/components/SkyboxComponent.ts
HexaField: It is an optimization for instance servers

Tortillaguy: @HexaField Should I set up an env var for this?

HexaField: no we don't need to worry about this for now

PR #1592, opened 2025-04-07
Texture Filtering on tileable meshes shows grey grid based on Camera Distance

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-8280


packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -1002,6 +1003,8 @@ const loadTextureImage = async (
   loader: ImageLoader | ImageBitmapLoader | TextureLoader | KTX2Loader | Loader
 ) => {
   const json = options.document
+  const rendererComponent = getComponent(getState(ReferenceSpaceState).viewerEntity, RendererComponent)
```
HexaField: we cannot assume the viewer renderer is always available (such as avatar or resource preview in the admin panel), we should have a store separate from the threejs renderer that holds these capabilities

HexaField (changes requested):

packages/engine/src/gltf/GLTFLoaderFunctions.ts
hzraslan: hi Josh, I created a helper function to get max anisotropy without using viewer renderer.

packages/engine/src/gltf/gltfUtils.ts
```diff
@@ -355,3 +355,22 @@ export function gltfReplaceUUIDsReferences(gltf: GLTF.IGLTF, UUIDs: [EntityUUID,
     }
   }
 }
+export function getMaxAnisotropyWithoutRenderer(): number {
```
HexaField: we can cache this result to a module level variable that starts off as uninitialized, so we don't compute it every time a texture loads, and we also should automatically return false if `isClient` is false.

HexaField (dismissed):

packages/engine/src/gltf/gltfUtils.ts
hzraslan: updated the branch, added a module level variable and isClient check

HexaField: looks good, let's rename the variable `maxAnisotropySupported`

hzraslan: renamed to `maxAnisotropySupported ` and pushed

HexaField: `getMaxAnisotropyWithoutRenderer` should also just be `getMaxAnisotropy`

hzraslan: changed and pushed

HexaField:
I want to make sure this isn't going to mess with anything else. @AidanCaruso @MichaelEstes is it OK to force generate mipmaps always like this?

AidanCaruso:
@HexaField the threejs loader only generates mipmaps if the texture does not use linear/nearest filtering, and is uncompressed. I'm not sure if any of these cases will actually cause problems when mipmapping is used, though not checking for compressed textures does seem potentially problematic

https://github.com/mrdoob/three.js/blob/dev/examples/jsm/loaders/GLTFLoader.js#L3345C4-L3345C137

hzraslan:
I don't think it's necessary to force mipmap generation for this issue. texture.generateMipmaps is true by default, so there's no need to explicitly set it.
Reference: [Three.js Texture.generateMipmaps](https://threejs.org/docs/#api/en/textures/Texture.generateMipmaps)
So I’ve removed that part from the PR.

DanielBelmes (approved):

PR #1599, opened 2025-04-08
Add Storybook support and new components for the editor

SYBIOTE:
Introduce Storybook support along with initial stories for various components, enhancing the editor's development experience. Update dependencies and improve existing component stories. Remove unused components and add necessary configuration files.


aditya-mitra:
Do we need to get storybooks for every editor, admin table, panel and studio components?
Most of these components work together or have just a minor variation of columns or headings.
We should get the scope of what the designers want to have stories of.
Because having so many scopes will be harder to maintain and will delay adding even minor changes.

DanielBelmes:
> Do we need to get storybooks for every editor, admin table, panel and studio components? Most of these components work together or have just a minor variation of columns or headings. We should get the scope of what the designers want to have stories of. Because having so many scopes will be harder to maintain and will delay adding even minor changes.

I'm not positive. I know they want ways to be able to verify design changes in general. So any surface they can view is good. I am not positive if this is the right surface but it would work.

SYBIOTE:
right now its covering almost everything, so we will only be narrowing the scope, which is still easier than adding it

Tortillaguy:
I see there's multiple storybooks. How do we run them together or what's the plan for combining them?

SYBIOTE:
yes there are three, the OG idea was to use storybook compose, but that still requires us to deploy three other storybooks
ideally we should just deploy 3 storybooks to chromatic

MbfloydIR (approved):

PR #1600, opened 2025-04-08
chore(ir-engine): analytics logs flow optimization

jonathan-casarrubias:
## Summary
- Optimized analytics logging functionality by removing manual payload object's mapping 
- Organized auto populated properties within a resolver
- Implemented internal analytics log schema and big query analytics log schema
- Implemented schema validators as a contracts for internal processes and clients' calls to comply with the required schema for analytics to correctly work


sfgarza (commented):
Code Looks good to me, I think unit tests just need to be added to meet the code coverage requirements

PR #1602, opened 2025-04-08
SpawnPointComponent Unit Tests

Tortillaguy:
## References
closes [IR-6390]

## QA Steps

[IR-6390]: https://tsu.atlassian.net/browse/IR-6390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/engine/src/scene/util/testUtil.ts
```diff
@@ -0,0 +1,35 @@
...
+Original Code is the Infinite Reality Engine team.
+All portions of the code written by the Infinite Reality Engine team are Copyright © 2021-2023 
+Infinite Reality Engine. All Rights Reserved.
+*/
+
+import { createEngine, createEntity, destroyEngine, Entity, removeEntity } from '@ir-engine/ecs'
+import { mockSpatialEngine } from '@ir-engine/spatial/tests/util/mockSpatialEngine'
+import { test as base } from 'vitest'
+
+export const it = base.extend<{ entity: Entity }>({
```
HexaField: i would prefer not to abstract things behind APIs like this, the test file should specify all of it's setup and configuration for clarity

HexaField (dismissed):

packages/engine/src/scene/util/testUtil.ts
Tortillaguy: I found this often repeated in component class tests

packages/engine/src/scene/components/SpawnPointComponent.ts
```diff
@@ -56,18 +56,21 @@ export const SpawnPointComponent = defineComponent({
   }),
 
   reactor: function () {
+    console.log('reactor')
```
DanielBelmes: Can you remove the console logs please

DanielBelmes (changes requested):

DanielBelmes (approved):

PR #1612, opened 2025-04-09
[HOTFIX 1.0.2] IR-7895: filter out invalid target options for interactable component

jose-galvan:
## Summary

Interactable component breaks due to some entities without `NodeIDComponent` being listed as target options

- Updated query to get entities with `CallbackComponent` and `NodeIDComponent`
- Filter out options with invalid properties in Select Component

[IR-7895
](https://tsu.atlassian.net/browse/IR-7895)

## References
closes #_insert number here_

## QA Steps

- Open a scene in the editor 
- Add a primitive geometry to the scene 
- Add an interactable component 
- Save changes 
- Select the entity and go properties, scroll down all the way to the interactable component 
- Enter preview mode 

**Expected result:** Interactable component should not break when entering preview mode


SamMazerIR (approved):
lgtm

PR #1613, opened 2025-04-09
Video visible / invisible callbacks missing

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-8906

The setVisible / setInvisible callbacks are missing because, are missing because there is no ObjectComponent on the video’s Entity. 
The  ObjectComponent is needed to access the  setVisible / setInvisible callbacks

set the ObjectComponent on the auth layer when creating the mesh for a video, so its setVisible, setInvisible callbacks will be accessible for triggers

## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/scene/components/VideoComponent.tsx
+++ b/packages/engine/src/scene/components/VideoComponent.tsx
@@ -31,7 +31,6 @@ import {
   LinearFilter,
   Mesh,
   MirroredRepeatWrapping,
-  Object3D,
   PlaneGeometry,
   RepeatWrapping,
   ShaderMaterial,
@@ -257,7 +256,8 @@ function VideoReactor() {
     )
     const authEntity = getAuthoringCounterpart(entity)
     if (authEntity !== UndefinedEntity) {
-      setComponent(authEntity, ObjectComponent, new Object3D())
+      const mesh = getComponent(videoMeshEntity, MeshComponent)
+      setComponent(authEntity, ObjectComponent, mesh)
     }
     return videoMeshEntity
   }).value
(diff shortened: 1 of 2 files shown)


packages/engine/src/scene/components/VideoComponent.tsx
```diff
@@ -252,6 +255,10 @@ function VideoReactor() {
         })
       )
     )
+    const authEntity = getAuthoringCounterpart(entity)
+    if (authEntity !== UndefinedEntity) {
+      setComponent(authEntity, ObjectComponent, new Object3D())
```
DanielBelmes: I think the video component already has an object3D for the plane we create to display the texture. We should use that when setting the ObjectComponent

DanielBelmes (changes requested):

packages/engine/src/scene/components/VideoComponent.tsx
MbfloydIR: updated

HexaField: This isn't a great solution. The visible component callbacks shouldn't be tied to the object component's existence. And honestly, the whole callback system is not a great pattern. I'm wanting to refactor it completely, and I don't know what side effects setting the object component on the authoring layer will have.

DanielBelmes: This is release1.0.2 It might be hard to refactor callbacks for the release sadly.

HexaField: if the callbacks exist on the simulation entity, instead we should make the editor look for the callbacks on the simulation entity, if this is what the problem is. Rather than messing with data across the simulation and authoring layer. I just don't see this being a good solution.

DanielBelmes (approved):

PR #1627, opened 2025-04-10
Improve individual workflow actions to properly utilize cache and ensure cache misses recover

DanielBelmes:
### Problem Statement
We are creating a cache for almost every single PR change. This is inefficient as the only times we really want to reinstall node_modules if if package.json(but really package lock) changes. We were hitting situations where a person's step would fail and the next day they would attempt to rerun it and the cache would no longer be valid and they would hit odd errors since we were relying on the cache to pass the entire repo files around. 

### Work Done
- setup-node already has caching features available to it that can ensure we don't overuse the cache functionality leading to failing tests from cache misses.
- Added fallback for if cache restoring does not work. 
- Instead of caching entire repo. Only cache node_modules and checkout repo. This should speed up new PR checks once this runs on 'dev'
- Furthermore I have to do a lot of work to make package.json work. It's highly expected we utilize package-lock.json for a multitude of reasons they list in their documentation. 
  - https://github.com/actions/setup-node?tab=readme-ov-file#checking-in-lockfiles
  - https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#working-with-lockfiles
  - https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md

## References
IR-9030 - https://tsu.atlassian.net/browse/IR-9030


DanielBelmes:
Note. I wanted to see if improving some of this would improve sonarQube randomly 403ing. It has not. I can not figure out why either the action or gh is being weird with that. We might have to contact sonarQube. But about 40% of the time the sonar cube analysis step fails.

barankyle (approved):

PR #1628, opened 2025-04-10
Deep nested deltas

AidanCaruso:
## Summary

SourceComponent,
NodeIDComponent,
UUIDComponent

are now composed in a single UUIDComponent, with properties

EntitySourceID
EntityID

EntitySourceID is the UUID of the entity's source
EntityID is the entity's ID

This gives access to the entity source, the entity's ID, and the entity's UUID when the two are concatenated.

Entity UUIDs are now just concatenated entity IDs - without any URLs. 

This architectural change allows nested scene deltas, among other things, to be serialized based on concatenaed uuids, rather than node id chain.

## References
closes #_insert number here_


packages/network/src/functions/WorldNetworkAction.ts
```diff
@@ -33,7 +33,7 @@ import { matchesUserID } from './matchesUserID'
 export class WorldNetworkAction {
   static spawnEntity = defineAction({
     type: 'ee.network.SPAWN_ENTITY',
-    entityUUID: matchesEntityUUID,
+    entityUUIDArray: matchesEntityUUIDArray,
```
HexaField: I'd recommend passing this as two separate parameters, which will be safer

packages/ecs/src/UUIDComponent.ts
```diff
@@ -91,11 +92,14 @@ export const UUIDComponent = defineComponent({
     return UUIDComponentFunctions._getUUIDState(uuid, layer).get(NO_PROXY_STEALTH)
   },
 
-  getOrCreateEntityByUUID(uuid: EntityUUID, layer = Layers.Simulation as LayerID) {
+  getUUID: (idArray: EntityUUIDArray) => idArray.toString() as EntityUUID,
```
HexaField: I'd prefer that we pass in an entity rather than an `EntityUUIDArray`, and replace all calls to `toString()`, such that if we need to update how we are stringifying we have one place for it - will also make strict typing more compliant

HexaField (changes requested):

packages/ecs/src/UUIDComponent.ts
AidanCaruso: I agree - any remaining .toString() calls are holdovers from previous testing that I haven't yet replaced with the component centralized component function

packages/network/src/functions/WorldNetworkAction.ts
```diff
@@ -33,7 +33,7 @@ import { matchesUserID } from './matchesUserID'
 export class WorldNetworkAction {
   static spawnEntity = defineAction({
     type: 'ee.network.SPAWN_ENTITY',
-    entityUUID: matchesEntityUUID,
+    entityUUIDPair: matchesEntityUUIDPair,
```
HexaField: Been thinking about this. For networking, we require that there is a parentUUID to attach the entity to, and we know that the parent is already in a context, thus I believe we can just supply the second aspect of the entityUUID here, and get the first aspect from the parent. I believe this will be safer and simple, essentially ensuring there is only one source of truth.

AidanCaruso: Yeah good thinking

AidanCaruso: In practice, this works great except for cases where we break the pattern of uuid determined by parent entities. 

This means avatars, ik targets, etc, where the instance ID needs to be the user ID, we can't rely on the parent entity UUID. I've experimentally added an optional parameter to the spawn action for these cases that overrides the use of the parent entity UUID.

HexaField (approved):

PR #1630, opened 2025-04-11
IR-8715 Add reference number to moderation schema and update related components

mshafiqmk:
## References
[Update 'Report ID' in Moderation section to be a simpler number](https://tsu.atlassian.net/browse/IR-8715)


packages/server-core/src/moderation/moderation/migrations/20250411143123_add-reference-number-col.ts
```diff
@@ -0,0 +1,63 @@
...
+      .where('id', records[i].id)
+      .update({ referenceNumber: i + 1 })
+  }
+
+  // 3. Modify the column to be AUTO_INCREMENT and UNIQUE using raw SQL
+  const tableName = moderationPath.replace(/[^a-zA-Z0-9_]/g, '') // sanitize if needed
+  await knex.raw(`
+    ALTER TABLE \`${tableName}\`
+    MODIFY COLUMN referenceNumber INT UNSIGNED NOT NULL AUTO_INCREMENT UNIQUE
+  `)
```
hanzlamateen: Do this using knex.

hanzlamateen (changes requested):

packages/server-core/src/moderation/moderation/migrations/20250411143123_add-reference-number-col.ts
mshafiqmk: okay

packages/common/src/schemas/moderation/moderation.schema.ts
```diff
@@ -112,6 +114,9 @@ export const moderationQuerySchema = Type.Intersect(
     querySyntax(moderationQueryProperties, {
       id: {
         $like: Type.String()
+      },
```
hanzlamateen: This can be removed i think

mshafiqmk: done

hanzlamateen (approved):

PR #1636, opened 2025-04-11
[IR-8919] Add Info icon linking to component docs

likethemammal:
## Summary

image

- Added an "i" icon next to each ComponentDropdown name that links to that Component's documentation.
- Added a `documentationList` array with the list of documented components
- Added a `documentationMap` object which maps some of the component types that dont match the syntax in the docs routing
- Had to use custom regex instead of slugify cus slugify assumes spaces between words

## QA Steps

- Select an Entity
- In the Properties Panel the components within the Entity should have an "i" icon next to the components with documentation
- Clicking the "i" icon should navigate to the appropriate documentation page
- There should be no icon next to components that don't have documentation

### DIFF
--- a/packages/editor/src/panels/properties/common/NodeEditor.tsx
+++ b/packages/editor/src/panels/properties/common/NodeEditor.tsx
@@ -73,6 +73,41 @@ class NodeEditorErrorBoundary extends React.Component<INodeErrorProps, INodeErro
   }
 }
 
+const documentationMap = {
+  GLTFComponent: 'ModelComponent',
+  RigidBodyComponent: 'RigidbodyComponent',
+  TriggerCallbackComponent: 'TriggerComponent'
+}
+
+const documentationList = [
+  'ModelComponent',
+  'AudioComponent',
+  'VideoComponent',
+  'ImageComponent',
+  'PrimitiveGeometryComponent',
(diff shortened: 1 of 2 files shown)


packages/editor/src/panels/properties/common/NodeEditor.tsx
```diff
@@ -73,6 +73,41 @@ class NodeEditorErrorBoundary extends React.Component<INodeErrorProps, INodeErro
...
 }
 
+const documentationMap = {
+  GLTFComponent: 'ModelComponent',
+  RigidBodyComponent: 'RigidbodyComponent',
+  TriggerCallbackComponent: 'TriggerComponent'
+}
+
+const documentationList = [
+  'ModelComponent',
```
DanielBelmes, line 83: I do wish we could somehow source this from a dynamic source of truth but I think it's fine for now and renders faster.

DanielBelmes (approved):

packages/editor/src/panels/properties/common/NodeEditor.tsx
likethemammal: I agree. When I started I assumed the slugs would be available on the `component`s directly. Maybe we can add something to the build step for the docs

PR #1637, opened 2025-04-11
[Release 1.0.2] IR-7891: Improve url error handling in GLTFComponent

jose-galvan:
## Summary

This change prevents the INVALID_SOURCE error from showing when the component is just added to the scene and doesn't have a value. 

Show both type of errors as inline errors to consolidate the UI 

Screen Recording 2025-04-14 at 3 32 08 p m

## References
closes #_insert number here_

## QA Steps

 - Open an scene in the studio
 - Add an entity
 - Add a model component to the entity 
 - Observe there's no errors shown after the component was added 
 - Enter an invalid url 
 - Observe the errors added to the component 
 - Enter a valid url 
 - Observe the errors dissappear from the component

### DIFF
--- a/packages/ui/src/components/editor/properties/gltf/loader/index.tsx
+++ b/packages/ui/src/components/editor/properties/gltf/loader/index.tsx
@@ -81,7 +81,6 @@ const GLTFNodeEditor: EditorComponentType = (props) => {
   const validRootMesh = hasComponent(props.entity, MeshComponent)
   const validChildMeshes = childMeshEntities.length !== 0
   const showMeshError = isMeshOrConvexHull && !(validChildMeshes || validRootMesh)
-  const urlInputTouched = useHookstate(false)
 
   const errors = ErrorComponent.useComponentErrors(props.entity, GLTFComponent)?.value
   const srcProject = useHookstate(() => {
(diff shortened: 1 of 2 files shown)


MbfloydIR (approved):

HexaField (changes requested):
this seems like a bug with the way errors are handled in components, rather than needing this additional login in the UI

HexaField:
Or perhaps, the solution is to no show the error if the asset is still loading or has no source

HexaField:
Okay yeah - the solution here is to remove the addError call in the GLTFComponent L#501

```ts
    if (!url) {
      addError(entity, GLTFComponent, 'INVALID_SOURCE', 'Invalid URL') // remove this, still return
      return
    }
```

jose-galvan:
> Okay yeah - the solution here is to remove the addError call in the GLTFComponent L#501
> 
> ```ts
>     if (!url) {
>       addError(entity, GLTFComponent, 'INVALID_SOURCE', 'Invalid URL') // remove this, still return
>       return
>     }
> ```
@HexaField can you review again, please? I removed the call to AddError on the GLTF component but 
but I also kept the changes to the Editor component that handle styling the errors and cleaning up when the URL changes. This was necessary for cases where a user inputs an invalid URL and then corrects it (see attached GIF).

Screen Recording 2025-04-21 at 7 37 21 p m

packages/ui/src/components/editor/properties/gltf/loader/index.tsx
```diff
@@ -147,15 +147,21 @@ const GLTFNodeEditor: EditorComponentType = (props) => {
       Icon={GLTFNodeEditor.iconComponent}
       {...props}
     >
-      <InputGroup name="Model Url" label={t('editor:properties.model.lbl-modelurl')}>
+      <InputGroup name="Model Url" className="flex flex-col gap-y-2" label={t('editor:properties.model.lbl-modelurl')}>
         <ModelInput
           value={gltfComponent.src.value}
           onRelease={(src) => {
+            if (src != gltfComponent.src.value) {
+              removeError(props.entity, GLTFComponent, 'LOADING_ERROR')
```
HexaField: removing errors like this should happen in the gltf component logic, not the UI

HexaField (dismissed):

packages/ui/src/components/editor/properties/gltf/loader/index.tsx
jose-galvan: Fixed!

DanielBelmes (approved):

PR #1642, opened 2025-04-14
[Wizard Enhancement] Add loading Bar/Spinner while the template is loading into the viewport

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-8893


HexaField (changes requested):
this is not the right approach. this will create race conditions and problems where we are collapsing possibly multiple states of the loading spinner into one boolean.

Rather than using the spinners themselves, we can just use `GLTF.useSceneLoaded(entity)` where ever we need to update UI.

hzraslan:
Thanks @HexaField. In that case I think we don't need that pr. Closing it

PR #1644, opened 2025-04-15
 [IR-9042]: Fix iOS fullscreen bars with viewport-fit

likethemammal:
## Summary

On iOS the html element doesnt scale to the full width of the viewport, showing white bars instead.

By adding `viewport-fit=cover` to the meta tags, iOS will scale the content to the viewport

image

## References

- https://webkit.org/blog/7929/designing-websites-for-iphone-x/

## QA Steps

- On an iPhone, open Safari
- Go to a published project link
- Turn phone to landscape
- The app should stretch to both screen edges without white bars on either side.

### DIFF
--- a/packages/client/index.html
+++ b/packages/client/index.html
@@ -11,7 +11,7 @@
     </style>
     <base href="/" />
     <meta charset="UTF-8" />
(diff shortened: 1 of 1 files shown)


aditya-mitra (approved):
works nicely!

PR #1645, opened 2025-04-15
IR-6490 Migrate authentication-setting to engine-setting

mshafiqmk:
## Summary of PR

1. Migrate the authentication settings into the engine settings.  
2. Update the config file to reference the `engine_settings` table instead of `engine-setting`.  
3. Modify the admin authentication component to use the engine settings.   

## References
Related PR in MT [IR-6490 Update authentication settings path and refactor data retrieval in AccountDetails and SignInPage components #907](https://github.com/theinfinitereality/irpro-multitenancy/pull/907)
[Migrate authentication-setting to engine-setting](https://tsu.atlassian.net/browse/IR-6490)


barankyle (changes requested):
Mostly looks good, though I want to test it on mt-qat after these issues have been addressed:

1. The new settings `authStrategies.[XX].<service_name>`, e.g. `authStrategies.[7].google`, need to be public. These indicate whether the respective auth strategy is enabled, and that's used on the login modal at /location/<location>, at least, to determine which icons to show guest users. If they're all private, then when you do a FIND on engine-settings on type 'authentication', nothing gets returned, and you can't log in.
2. At /admin/settings#Authentication, if I click the toggle to disable an auth type, then click save, that is not getting saved. It doesn't even look like any request is getting made to the engine-setting endpoint, unlike if I change a public or private key for some auth strategy and then click Save.

Also please remove any console.log's.

barankyle (changes requested):
I'm noticing that ProfileMenu is now re-rendering endlessly, which is bad for performance. If you open the menu from a location, you'll see that all of the console.log's in there just keep printing over and over. I don't see this behavior on the dev branch, so it seems it's caused by something in this PR.

All of those debugging console.log's should be removed, but first the endless re-renders should be addressed. Otherwise, this looks good.

mshafiqmk:
> I'm noticing that ProfileMenu is now re-rendering endlessly, which is bad for performance. If you open the menu from a location, you'll see that all of the console.log's in there just keep printing over and over. I don't see this behavior on the dev branch, so it seems it's caused by something in this PR.
> 
> All of those debugging console.log's should be removed, but first the endless re-renders should be addressed. Otherwise, this looks good.

Thanks , I'll take a look into it.

hanzlamateen (approved):

PR #1655, opened 2025-04-16
Ir 9016 replace editor history full history with json patch operations

HexaField:
## References
closes #_insert number here_


MichaelEstes (commented):
There seem to be some errors with undoing adding an object to a scene in the hierarchy panel

packages/engine/src/authoring/AuthoringState.tsx
```diff
@@ -0,0 +1,432 @@
...
+  return null
+}
+
+/**
+ * Given a list of commands, compute the final state of the history.
+ * @param commands
+ * @returns
+ */
+export const computeCommands = (commands: HistoryCommand[], sourceID?: SourceID) => {
+  const commandLength = commands.length
```
MichaelEstes, line 297: commands is undefined here if there's nothing to undo, other than that and the failing test, looks good. I'll approve now and you can merge in when you fix the test.

MichaelEstes (approved):

PR #1658, opened 2025-04-16
IR-8716 Allow moderators to filter by user name in the search input of moderation

mshafiqmk:
## References
[Update moderation search field to search by name](https://tsu.atlassian.net/browse/IR-8716)


packages/client-core/src/admin/components/moderation/ModerationTable.tsx
```diff
@@ -62,6 +62,15 @@ export default function ModerationTable({ search }) {
   const { t } = useTranslation()
   const [statusFilter, setStatusFilter] = useState(ModerationFilterStatus.All)
   const [selectedReport, setSelectedReport] = useState<ModerationType>()
+  const [searchQuery, setSearchQuery] = useState('')
+
+  useEffect(() => {
+    const searchDelayTimer = setTimeout(() => {
```
hanzlamateen: It seems something like debounce is being used all across our admin panel code to delay the search. Can you use same pattern to make code look more coherant.

image

packages/client-core/src/admin/components/moderation/ModerationTable.tsx
```diff
@@ -72,43 +81,68 @@ export default function ModerationTable({ search }) {
   const handleReportResolve = (report: ModerationType) => {
     setSelectedReport(userReportsQuery.data[userReportsQuery.data.findIndex((r) => r.id === report.id) + 1])
   }
+
+  const userIds = useFind(userPath, {
```
hanzlamateen: You shouldnt do it like this, this should be done in backend as some users might not have `user:read` scope and this api request will fail for them.

Search for `context.service.createQuery(` in our code based to get an idea and then you can have a join in place with where clause.

hanzlamateen (changes requested):

packages/client-core/src/admin/components/moderation/ModerationTable.tsx
mshafiqmk: I believe there is already a debounce implementation in the` <Search/>` component, so it's unnecessary here. I'll remove it.

packages/client-core/src/admin/components/moderation/ModerationTable.tsx
mshafiqmk: Done

packages/server-core/src/moderation/moderation/moderation.hooks.ts
```diff
@@ -53,17 +56,57 @@ const validateModeration = async (context: HookContext) => {
...
     all: [schemaHooks.validateQuery(moderationQueryValidator), schemaHooks.resolveQuery(moderationQueryResolver)],
-    find: [iff(isProvider('external'), verifyScope('moderation', 'read'))],
+    find: [iff(isProvider('external'), verifyScope('moderation', 'read')), persistQuery, handleModerationSearch],
     get: [
       iff(isProvider('external'), verifyScope('moderation', 'read')),
       schemaHooks.validateQuery(moderationQueryValidator),
-      schemaHooks.resolveQuery(moderationQueryResolver)
+      schemaHooks.resolveQuery(moderationQueryResolver),
+      persistQuery,
+      handleModerationSearch
```
hanzlamateen: We dont need this in get. Only find method is used in client for search.

hanzlamateen (changes requested):

packages/server-core/src/moderation/moderation/moderation.hooks.ts
mshafiqmk: done

hanzlamateen (approved):

PR #1665, opened 2025-04-16
[IR-9031] Fix VideoComponent autoplay on iOS

likethemammal:
## Summary

The VideoComponent doesnt show a first frame and doesn't autoplay on iOS because of how iOS handles video autoplay works. iOS requires `muted` to be set to allow autoplay on mobile.

image

## References
- https://developer.apple.com/documentation/webkit/delivering-video-content-for-safari#Enable-Video-Autoplay

## QA Steps

- Open safari on iPhone
- Go to a published project with an autoplaying video in it
- Video should autoplay on iOS

### DIFF
--- a/packages/engine/src/scene/components/VideoComponent.tsx
+++ b/packages/engine/src/scene/components/VideoComponent.tsx
@@ -460,8 +460,30 @@ function VideoReactor() {
     const sourceTexture = sourceVideoComponent?.texture
 
     if (video.texture.value) {
+      const videoEl = mediaElement.element as HTMLVideoElement
+
+      const resetMuted = () => {
+        videoEl.muted = false
+        document.removeEventListener('pointerdown', resetMuted)
+      }
+
+      if (videoEl.paused) {
+        videoEl.pause()
+      } else {
+        videoEl.play().catch((error) => {
+          if (error.name === 'NotAllowedError') {
+            videoEl.muted = true
+            videoEl.play()
+
+            document.addEventListener('pointerdown', resetMuted)
+          } else {
+            console.error(error)
+          }
+        })
+      }
+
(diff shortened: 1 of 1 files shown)


packages/engine/src/scene/components/VideoComponent.tsx
```diff
@@ -460,8 +460,11 @@ function VideoReactor() {
     const sourceTexture = sourceVideoComponent?.texture
 
     if (video.texture.value) {
+      const videoEl = mediaElement.element as HTMLVideoElement
+      videoEl.muted = iOS
```
CITIZENDOT: with this, video is always muted on iOS right?

to avoid this, start off the video with `muted` set to `true`. when user interacts with the page (clicks), video can play with audio. Similar logic is applied to MediaComponent here: https://github.com/ir-engine/ir-engine/pull/1439

CITIZENDOT (dismissed):

packages/engine/src/scene/components/VideoComponent.tsx
likethemammal: I thought to do this, but I wasnt sure where to add the click logic (and it ended up being opened up to the UX team). I didnt think adding a generic `document` listener just for this case was appropriate (I figured there was a place we should put this listener and then use state to handle the `muted` toggle), but it does seem clean enough in ur PR.

Question: Wouldnt #1439, that just got merged in yesterday, fix this issue, making this PR redundant? I'm new to the codebase, but its my understanding that `VideoComponent` would inherit the autoplay fix from `MediaComponent`

likethemammal: Updated

DanielBelmes (approved):

PR #1666, opened 2025-04-16
[IR-9036] Load all resources from a gltf into the ecs rather than just the ones specified by the node hierarchy

MichaelEstes:
## References
closes #_insert number here_


packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -82,6 +83,8 @@ export const GLTFComponent = defineComponent({
 
   schema: S.Object({
     src: S.String(''),
+    library: S.Bool(false),
```
HexaField: this shouldn't be necessary. there should only be one loader and it should take all the descriptions in the gltf and load it into the ecs, regardless if it is part of a node hierarchy or not. we don't want different kinds of loaders and different kinds of loading methodologies.

HexaField (changes requested):

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -137,6 +139,12 @@ export const GLTFComponent = defineComponent({
   },
   removeHashes: <T extends EntityUUID | SourceID | NodeID | string>(url: T) => {
     return url.replaceAll(/\?hash=[^-]+/g, '') as T
+  },
```
HexaField: are these helper functions needed?

HexaField: just need these two `isMaterialGLTF` and `isLookdevGLTF` functions removed and it looks good to me!

HexaField (changes requested):

HexaField (changes requested):
We should also add a test info loading a GLTF file with a material but no meshes/nodes

HexaField (approved):

PR #1668, opened 2025-04-16
[MT-STG, UAT] - Compress on Publish error "object null is not iterable"

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-8294


packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
```diff
@@ -223,6 +224,8 @@ export default function AddEditLocationModal(props: AddEditLocationModalProps) {
     if (!isValid) {
       return
     }
+    let saveScenePath: string | undefined
```
SYBIOTE: use hookstate instead of local variables

SYBIOTE (approved):

SYBIOTE (changes requested):
change local variable to hookstates

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
hzraslan: updated the branch

hzraslan:
@SYBIOTE changed and updated the branch

SYBIOTE (approved):

PR #1684, opened 2025-04-18
[IR-8732] Unit Tests - GLTF Loading: Asset

heysokam:
## References
https://tsu.atlassian.net/browse/IR-8732


packages/engine/src/gltf/loader/GLTFLoader.asset.test.tsx
```diff
@@ -28,41 +28,116 @@ Infinite Reality Engine. All Rights Reserved.
...
-    it.todo('MUST be a `string` type when defined', () => {})
+    it('MAY be undefined', () => {
+      const asset = mockAssetMinimal()
+      expect(() => GLTFValidate.asset(asset)).not.toThrow()
+    })
+
+    it('MUST be a `string` type when defined', () => {
+      const asset = mockAssetMinimal()
+      // @ts-expect-error Set the copyright property to a number to provoke the error
+      asset.copyright = 2023 // Not a string
```
HexaField, line 80: do we want a positive case here too? for when it IS a string?

heysokam: Oh, good point. I didn't notice that these optional properties don't have any other cases for checking the valid case, since the spec has no requirement other than being a string. Will add that :writing_hand:

heysokam: Added them on fb4893b

packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -1753,3 +1754,61 @@ export type GLTFParserOptions = {
   path: string
   requestHeader: Record<string, string>
 }
+
+const validateVersionFormat = (vers: string): boolean => {
+  assert(vers.length < 1024, 'version string too long')
```
HexaField: do we want to use assert here? or throw an error?

HexaField (approved):

PR #1691, opened 2025-04-18
Support for GIFs to be displayed in the image component

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-8289

### DIFF
--- a/packages/server-core/src/media/ffmpeg/ffmpeg.hooks.ts
+++ b/packages/server-core/src/media/ffmpeg/ffmpeg.hooks.ts
@@ -1,0 +1,104 @@
+/*
+CPAL-1.0 License
+
+The contents of this file are subject to the Common Public Attribution License
+Version 1.0. (the "License"); you may not use this file except in compliance
+with the License. You may obtain a copy of the License at
+https://github.com/ir-engine/ir-engine/blob/dev/LICENSE.
+The License is based on the Mozilla Public License Version 1.1, but Sections 14
+and 15 have been added to cover use of software over a computer network and 
+provide for limited attribution for the Original Developer. In addition, 
+Exhibit A has been modified to be consistent with Exhibit B.
+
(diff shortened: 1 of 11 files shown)


packages/server-core/src/media/ffmpeg/ffmpeg.hooks.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  if (!isValidFileType(file.mimetype, file.originalname)) {
+    throw new BadRequest('Unsupported file type')
+  }
+}
+
+export default {
+  before: {
+    all: [iff(isProvider('external'), verifyScope('editor', 'write'))],
+    find: [],
```
barankyle: If there's no plans to use a method, such as find, you can do
```
import { disallow } from 'feathers-hooks-common'
```
and then add the disallow hook to that method's before array here, like
```
get: [disallow()]
```
project-build.hooks.ts is one example of this. This probably isn't strictly necessary with a service like this that isn't backed by a database table, but it probably helps security at the margins to not expose methods that aren't used.

packages/server-core/src/media/ffmpeg/ffmpeg.hooks.ts
```diff
@@ -0,0 +1,83 @@
...
+    get: [],
+    create: [
+      (context) => {
+        context[SYNC] = false
+        return context
+      },
+      validateFile
+    ],
+    update: [],
+    patch: [
```
barankyle: I don't see a handler for the patch method in ffmpeg.class.ts. If this endpoint won't be supporting patches, then it should be converted to just run the `disallow()` hook as well.

hzraslan: Hi @barankyle thanks for reviewing. I made the change and updated the branch

DanielBelmes (approved):

PR #1703, opened 2025-04-21
MixerComponent: a component for interpolating properties of other components

Rezmason:
## Summary
MixerComponent implements the "Sequencer" concept. The MixerComponent's coordinate is externally driven; there isn't yet a timer that plugs into it, as that falls outside the component's scope:

> Create a component + system that allows generic driven interpolation of component values based upon some driver. Initially, we will create a 1D interpolation track, driven by time. This track should be able to be “subscribed” to by properties on other components in the scene, which will then have their values controlled by the track. 
> 
> Implement an efficient method for determining interpolation (<=O(log(N)) complexity)
> 
> Keep the implementation general enough that we can extend to 2D and 3D drivers.

## References
Tracked by https://tsu.atlassian.net/browse/IR-8719 .

### DIFF
--- a/packages/engine/src/scene/components/MixerComponent.tsx
+++ b/packages/engine/src/scene/components/MixerComponent.tsx
@@ -147,6 +147,7 @@ type MixerState = {
   properties: Map<string, Property>
   entriesByCoord: Map<number, [Entry, number]>
   sortedEntries: [number, Entry][]
+  needsUpdate: boolean
 }
 
 const schema = S.Object({
@@ -161,8 +162,7 @@ export const MixerComponent = defineComponent({
   jsonID: 'IR_mixer',
   schema,
 
-  reactor: (props: { mixerEntity: Entity }) => {
-    const entity = props.mixerEntity
+  reactor: ({ entity }: { entity: Entity }) => {
     const mixerComp = useComponent(entity, MixerComponent)
 
     useEffect(() => {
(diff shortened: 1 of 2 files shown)


the-reality-engineer:
This seems to overlap a bit with the BehaviorComponent that @HexaField is working on, as well as the Transitionable api.

HexaField:
I agree with @speigg. I'd like us to consider including this functionality in the Transition API, as a continuous transition type as opposed to the currently implemented static transition. Seems like all we need is some way to optionally reference a datastructure that is changing, rather than a static target value.

The BehaviorComponent already sits on top of the Transition API, so it would naturally be able to take advantage of this, rather than requiring a more custom integration.

Rezmason:
Flatterers.

packages/engine/src/scene/components/MixerComponent.tsx
```diff
@@ -0,0 +1,367 @@
...
+  propertyPath: string
+): string => `${toEntityUUID(targetEntity)}::${toComponentID(targetComponent)}::${propertyPath}`
+
+const unpackAddress = (packedAddress: string): PropertyAddress =>
+  packedAddress.split('::') as [EntityUUID, string, string]
+
+type Entry = Record<string, number[]>
+type Property = { type: MixableType; address: PropertyAddress }
+
+const createProperty = (
```
Rezmason, line 263: @dinomut1 Two important things not yet implemented in the component's properties are support for properties nestled in their schema, and deriving the type of a property from the schema.

packages/engine/src/scene/components/MixerComponent.tsx
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  const propertySchema = Component.schema.properties[propertyPath] // TODO: support properties nestled in schema
+
+  if (propertySchema == null) return null
+  const entity = toEntity(targetEntity)
+  if (entity == null) return null
+  const component = getComponent(entity, Component)
+  if (component == null) return null
+  const type = MixableType.Number // TODO: pull from propertySchema
```
dinomut1: We can't both not support nested properties and not infer type from the schema. This means we can't animate anything but top level scalar fields, leaving all transform properties off limits. We should be able to support both of these things in the mvp

dinomut1 (changes requested):

packages/engine/src/scene/components/MixerComponent.tsx
Rezmason: Now implemented. 👍

packages/engine/src/scene/components/MixerComponent.tsx
Rezmason: There's now support for discerning data type from schema kind, and getting and setting nested properties. @dinomut1 I'd like the nested property "setter" to receive special scrutiny— if it doesn't do what we need it to, then I should iterate on it. Ping me if you'd like me to elaborate or talk about it.

dinomut1 (approved):

PR #1707, opened 2025-04-21
[IR-8731] Unit Tests - GLTF Loading: Animations

heysokam:
## References
https://tsu.atlassian.net/browse/IR-8731


heysokam:
@HexaField Should be good to go for now.
There are some missing buffer setup issues, but I will come back to those all at once with the buffers task instead :writing_hand:

HexaField (approved):

PR #1717, opened 2025-04-22
[IR-5371] Replace non type json schema declarations with option metadata

MichaelEstes:
## References
closes #_insert number here_


packages/ecs/src/schemas/JSONSchemas.ts
```diff
@@ -262,74 +223,21 @@ export const S = {
...
-   * S.NonSerialized(S.Object({ test: S.Number(0) })) serializes to null
-   *
-   */
-  NonSerialized: <T extends Schema, Initial>(schema: T, init?: Initial, options?: TNonSerializedSchema<T>['options']) =>
-    ({
-      [Kind]: 'NonSerialized',
-      options: buildOptions(init, options),
-      properties: schema
-    }) as TNonSerializedSchema<T>,
+  Optional: <T extends Schema>(schema: T, options?: TUnionSchema<[T, TUndefinedSchema]>['options']) =>
```
HexaField, line 239: would we be able to remove the S.Undefined type and use an `{ optional: true }` option field?

HexaField (approved):

HexaField (changes requested):
Looks like layer propagation is broken, scenes load fine in location but not studio. Tested apartment and sky station, neither work.

HexaField:
I'm getting nasty errors and crashing just loading up the sky station in the studio.

HexaField:
> I'm getting nasty errors and crashing just loading up the sky station in the studio. In a freshly installed repo and reinited.

OK no i'm wrong, this is happening on dev too. Something is messed up with dev unfortunately it seems

HexaField (changes requested):
If I open the sky station in the studio, and then translate the skybase model with the transform tool, the whole model unloads.

HexaField:
> If I open the sky station in the studio, and then translate the skybase model with the transform tool, the whole model unloads.

That was an easy fix, was serializing data we shouldn't have

HexaField (approved):

PR #1719, opened 2025-04-22
[IR-9162] Add horizontal scroll to Files panel breadcrumb

likethemammal:
## Summary

The breadcrumb in the toolbar for the Assets and Files panel would break down with long breadcrumb lists. This change rearranges some of the toolbar elements and uses CSS Grid so everything in the toolbar stays visible regardless of the breadcrumb width. It also adds a horizontal scroll when the breadcrumb becomes to large, preventing the design from breaking.

image

## References

Closes https://tsu.atlassian.net/browse/IR-9162

## QA Steps
- In Asset panel go to Files
- Add or upload a folder with a deep folder structure
- Click into each folder
- The breadcrumb should update with the list of folder, as the breadcrumb gets too long for the design, a scroll bar should appear to horizontally scroll the breadcrumb list

### DIFF
--- a/packages/editor/src/panels/assets/topbar.tsx
+++ b/packages/editor/src/panels/assets/topbar.tsx
@@ -87,20 +87,22 @@ export function AssetsBreadcrumbs() {
   }
 
   return (
-    <div className="flex items-center gap-2" data-testid="assets-panel-breadcrumbs">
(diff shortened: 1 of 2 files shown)


CITIZENDOT (approved):
Works nicely!

https://github.com/user-attachments/assets/0eceb6dc-fded-436d-b61e-ffa880e2a050

PR #1728, opened 2025-04-22
[IR-8734] Unit Tests - GLTF Loading: BufferViews

heysokam:
## References
https://tsu.atlassian.net/browse/IR-8734


heysokam:
@HexaField Same as with the GLTFLoader.animations tests, these should be good to go until the Buffers data setup pass is done for all tests at once.

HexaField (approved):

PR #1733, opened 2025-04-22
Stop animations when "None" is selected or component is removed

koktavy:
## Summary
The Loop Animation Component now stops animating when "None" is selected from the clip dropdown, and when the component is removed.

## Subtasks Checklist
N/A

## Breaking Changes
N/A

## References
closes [#7892](https://tsu.atlassian.net/browse/IR-7892)

## QA Steps
1. Open Dev env ([IR Engine](https://ir-engine-mt-dev.theinfinitereality.io/dashboard) )
2. Create a scene

3. Place a gltf/glb with baked animation into scene (Ex: transforms.gltf)

4. Add Loop Animation Component to object

5. Select animation

6. Observe object animating

7. Select None for Loop Animation
8. Observe object animation stopped (FIXED)
9. Select animation again

10. Remove Loop animation component

11. Observe object animation stopped (FIXED)


packages/engine/src/avatar/components/LoopAnimationComponent.ts
```diff
@@ -187,7 +215,19 @@ export const LoopAnimationComponent = defineComponent({
...
+          loopAnimationComponent._action.value.stop()
+        }
+
+        // Also stop all actions in the mixer to be thorough
+        if (animComponent?.mixer?.value) {
+          animComponent.mixer.value.stopAllAction()
+        }
+
+        // Set activeClipIndex to -1 (None) to indicate no clip is selected
+        loopAnimationComponent.activeClipIndex.set(-1)
```
HexaField: I don't think we want to change the activeClipIndex, that is what specifies which animation should run and should only be set via the editor UI

HexaField (changes requested):
Let's also add some unit tests for positive and negative cases to catch regressions for this behavior

HexaField:
Please clean up the code comments, we only need ones that specify unintuitive things, todos or problems

HexaField (approved):

PR #1737, opened 2025-04-23
[IR-9183] [IR-9185] Fix overlapping styles in Hierarchy panel

likethemammal:
## Summary

The Hierarchy panel has some issues where the styling would breakdown if the panel was smaller width than text length of the children within.

- By changing the layout to use `grid`, the design stays intact when the text overflows and doesnt cause any overlap with the other elements.
- Added horizontal scroll to the text labels so that they can be scrolled when the panel is small

image

## QA Steps

- Select a Scene
- In the Hierarchy panel, select an entity
- Resize the panel to a small width
- You should be able horizontal scroll the names of each entity


aditya-mitra (approved):
Works fine!

PR #1739, opened 2025-04-23
Fix Loading Spinner Rel1.0.2

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-7399?linkSource=email

### DIFF
--- a/packages/editor/src/systems/ModelLoadingSpinnerSystem.tsx
+++ b/packages/editor/src/systems/ModelLoadingSpinnerSystem.tsx
@@ -53,7 +53,7 @@ const LoadingSpinnerReactor = (props: { entity: Entity }) => {
 
   useEffect(() => {
     if (!shouldHaveSpinned) return
-    spinnerEntity.current = createLoadingSpinner(`loading ${gltfComponent.src.value}`)
+    spinnerEntity.current = createLoadingSpinner(`loading ${gltfComponent.src.value}`, entity)
     return () => {
       removeEntityNodeRecursively(spinnerEntity?.current!)
       spinnerEntity.current = null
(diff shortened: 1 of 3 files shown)


DanielBelmes:
The functionality is mostly looking good however the feature is expected to spawn the spinner on top of the entity that is being loaded. So if I drag a model in I want the spinner where the model will show up and that has been lost. I know it was buggy before. I sadly have no video of it working before all these bugs.

hzraslan:
@DanielBelmes thanks for session again, it should be good to review again

DanielBelmes (approved):

packages/editor/src/systems/ModelLoadingSpinnerSystem.tsx
```diff
@@ -23,80 +23,50 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
-    const spinnerEntity = createLoadingSpinner(`loading ${gltfComponent.src.value}`, entity)
-    loadingEntity.set(spinnerEntity)
-  }
-
-  const removeLoadingGeo = () => {
-    if (!loadingEntity.value) return
-    removeEntityNodeRecursively(loadingEntity.value)
-    loadingEntity.set(UndefinedEntity)
-  }
+  const spinnerEntity = useRef<Entity | null>(null)
```
SYBIOTE, line 51: any specific reason we use useRef here? i think we use hook state across the codebase, use ref is rarely used, can this implementation be done with use hookstate, its fine other wise

SYBIOTE (approved):
looks good

DanielBelmes:
Was not supposed to be merged yet reverting and reopening

packages/editor/src/systems/ModelLoadingSpinnerSystem.tsx
```diff
@@ -23,80 +23,50 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
...
-      ))}
-    </>
-  )
-}
-
 export const ModelLoadingSpinnerSystem = defineSystem({
   uuid: 'ee.editor.ModelLoadingSpinnerSystem',
   insert: { before: PresentationSystemGroup },
-  reactor
+  reactor: () => (
```
HexaField, line 69: this is highly problematic. we cannot put this on the simulation layer, it is causing all sorts of issues loading avatars in

PR #1741, opened 2025-04-23
[IR-9087] Fix Files panel not displaying all files

likethemammal:
## Summary

Not all the files that should be present in the Files panel are present. This is caused by the `InfiniteScroll` component not retrieving all the `files` immediately. This change updates the Files panel to 

## QA Steps

- Upload a large set of images
- Open Files panel, navigate to large set of images
- Some of the expected images should only partially show until you scroll and the rest of the results should load.


aditya-mitra (approved):
Looks good!

PR #1752, opened 2025-04-24
IR-9227: Inject user id in GA4 GTM upon login

hanzlamateen:
## Summary
Added changes which should push userId on login to GTM.

Configurations on GTM can be done based on http://developers.google.com/analytics/devguides/collection/ga4/user-id?client_type=gtm

## References
https://tsu.atlassian.net/browse/IR-9227


hanzlamateen:
Discussed this with Muhammad Hussain and going to merge the PR to test it on dev.

PR #1756, opened 2025-04-25
fix bug with engine canvas crashing on builder startup

dinomut1:
disable csm due to shader errors in builder

## References
closes #_insert number here_


HexaField:
Happy to help debug whatever the core problem is here :)

dinomut1:
Yes please. There's some weird behavior relating to the initialization of the builder frontend UI. Specifically around changes to useEngineCanvas that have emerged in the last 3-4 weeks. So I've identified three different bugs:

1. The initialization of the builder frontend introduces a race condition in the handling of the engine render canvas. It looks like the useEngineCanvas being used in EditorContainer, and the useEngineCanvas being used in BuilderPanel, and the act of disabling the default editor UI, are all fighting with each other, and somehow that leads to the viewer entity being in the wrong state when the use effect in the useEngineCanvas in BuilderPanel runs (ie missing a RenderSettingsComponent, or not existing at all). So that's why I added the check for the RenderSettingsComponent. You can breakpoint the useEffect in useEngineCanvas to repro this when loading a builder scene
2. Once the scene does load, for some reason the CSM shader code is throwing index errors, saying that the shader code is trying to access more cascades than actually exist. I worked around this by just disabling the shadow system
3. Also once the scene loads, if you resize the window (like trying to drag the debug console to make it larger) the gl context crashes

HexaField:
@dinomut1 `EditorState` has a `canvasRef` property that was added recently which may be exactly what you're looking for. it persists the canvas and it's respective entities as the viewport is moved around different areas in the dock.

PR #1760, opened 2025-04-25
Avatars Now Roll Around

dinomut1:
Changed avatars so they roll around


dinomut1:
Cleaning up. This was for hackathon

PR #1772, opened 2025-04-29
Replace Material Prototypes with Components

HexaField:
## Summary

- Introduce new `defineMaterialPlugin` API that automatically synchronizes material uniforms via json schema with shader override callback and update loop
- Replaces material prototypes & EE_material with plugins
- Removes material specific authoring behavior with the general component solution

## References
closes #_insert number here_


HexaField:
Will wait to merge this in until https://github.com/ir-engine/ir-engine/pull/1628 is in

HexaField:
- need to ensure texture support still works
- need to use appropriate gltf plugins in place of prototypes

packages/engine/src/material/defineMaterialPlugin.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+          onApply(shader, renderer)
+        }
+        setPlugin(material, callback)
+        return () => {
+          removePlugin(material, callback)
+        }
+      }, [material])
+
+      useExecute(
```
AidanCaruso, line 159: Is it worth creating a system for every plugin instance, even if it does not need one? That's a lot of extra systems doing nothing in most cases. Currently elsewhere we already established the pattern of querying for entities in a single system to update plugin uniforms. 

In every current use case I see for this update loop (noise offset, lava, holographic), it's just updating a time value. Let me know if I'm missing one.

With that in mind, I would suggest including the time value implicitly with plugin components (be it via uniform object reference, single system setting the time values, etc). Then there may no longer be a reason to create a useExecute/update loop for every plugin instance to begin with.

packages/engine/src/gltf/MaterialExtensionComponents.tsx
```diff
@@ -1038,52 +997,3 @@ const invertGlossinessMap = async (glossinessMap: Texture) => {
 }
 
 export type MaterialExtensionPluginType = { id: string; uniforms: { [key: string]: any } }
-const MaterialExtensionPluginTypeSchema = S.Object({ id: S.String(), uniforms: S.Record(S.String(), S.Any()) })
-
-export const EEMaterialComponent = defineComponent({
```
AidanCaruso, line 1043: Just to be 100% sure, do we have PRs ready in the project(s) that define custom materials that have run them through the migrater?

AidanCaruso (changes requested):
looks good, however I am seeing some issues with uniform flickering with positional values set by the trasnparency system - but only in the studio so far. Odd

https://github.com/user-attachments/assets/4479c2fc-b3b4-4c08-9716-3aeac1a3a80c


Material authoring seems to not be working properly, can't set material colors, etc.
Also getting webgl errors and webgl susbequently not rendering anything any time I try to change a material prototype
image

Speaking of do we still even need any of these besides standard/physical/lambert to be defined here?
image

packages/engine/src/material/defineMaterialPlugin.tsx
HexaField: There is no performance concern here. Systems are essentially free when just setting values like this.

packages/engine/src/gltf/MaterialExtensionComponents.tsx
HexaField: Yes, it's just the irpro advanced materials project

AidanCaruso:
Material drag/drop from the advanced materials repo on the companion branch does not seem to be working. I'm not sure if this is because those gltf assets have not been migrated, or something in addMediaNode not loading it in. Or both.

HexaField:
> Material drag/drop from the advanced materials repo on the companion branch does not seem to be working. I'm not sure if this is because those gltf assets have not been migrated, or something in addMediaNode not loading it in. Or both.

So the reason this is not working is actually entirely unrelated, it seems like when you are drag and dropping something over the viewport, the input pointer position is not being updated

AidanCaruso:
Looks good. Approved. Does not seem to be causing any issues, but noting that I still see some mismatching between how colors are serialized
image

AidanCaruso (approved):

HexaField:
> Looks good. Approved. Does not seem to be causing any issues, but noting that I still see some mismatching between how colors are serialized

odd, not critical but i will look into it

PR #1775, opened 2025-04-29
Fix studio and modal images

DanielBelmes:
## Summary

Vite does not support importing images from /public. Moved images we use to a /assets dir in src as expected.

has an MT counterpart PR

## References
[IR-9236
](https://tsu.atlassian.net/browse/IR-9236)


aditya-mitra (approved):
Looks good!

PR #1783, opened 2025-04-29
[IR-9203] Fix Download container shaking while progress

likethemammal:
## Summary

The way the Download Project progress container was structured caused the entire container to shake every time the progress text was updated. This PR restructures the elements so that the changing text element doesnt affect the others.

By giving the text a `min-width` based on text size the progress text doesnt force the layout to change.

https://github.com/user-attachments/assets/9b69e9a0-225c-41fe-a19f-85968b15a15a

## QA Steps

- Open a scene
- Open Files or Assets panel
- OPTIONAL: DevTools throttling
  - Open DevTools ("Inspect Element")
  - Go to Performance tab
  - On the right side, under "Environment Settings", set CPU and set Network throttling (4x or more)
- Press Download button in Files/Assets panel
- The "Downloading Project" progress should appear. The progress text should be the only element that resizes

### DIFF
--- a/packages/client-core/i18n/en/editor.json
+++ b/packages/client-core/i18n/en/editor.json
@@ -1366,7 +1366,7 @@
       "uploadFolder": "Upload Folder",
       "uploadingFiles": "Uploading Files ({{completed}}/{{total}})",
       "fetchingProject": "Fetching Project",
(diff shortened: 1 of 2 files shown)


aditya-mitra (approved):
awesome1

PR #1785, opened 2025-04-29
[IR-9435] Fix Files list view table styling

likethemammal:
## Summary

The `InfiniteScroll` component adds an extra div element to the list of items, which breaks the styling in tables. This breaks the styling in the Files List view

By removing the extra div and using a React fragment `<></>` this problem is avoided.

image

## QA Steps

- Open Files panel
- Select List view
- Each list item should span the full width of the table


aditya-mitra (approved):
looks good!

PR #1787, opened 2025-04-30
fix text overflow on assets filebrowser

aditya-mitra:
## Summary

fix the misalignment issue caused by text overflow in my assets file browser resources

## References
closes https://tsu.atlassian.net/browse/IR-9082

## QA Steps

image

### DIFF
--- a/packages/editor/src/panels/assets/resources.tsx
+++ b/packages/editor/src/panels/assets/resources.tsx
@@ -205,7 +205,7 @@ export function FileCard({
             {name}
           </Text>
         </Tooltip>
-        <span className="max-w-24 overflow-hidden text-ellipsis whitespace-nowrap text-xs text-[#375DAF]">{info}</span>
+        <span className="w-24 overflow-hidden text-ellipsis whitespace-nowrap text-xs text-[#375DAF]">{info}</span>
       </div>
     </>
   )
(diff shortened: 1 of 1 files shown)


jose-galvan (approved):

DanielBelmes (approved):

packages/editor/src/panels/assets/resources.tsx
```diff
@@ -205,7 +205,7 @@ export function FileCard({
             {name}
           </Text>
         </Tooltip>
-        <span className="text-xs text-[#375DAF]">{info}</span>
+        <span className="max-w-24 overflow-hidden text-ellipsis whitespace-nowrap text-xs text-[#375DAF]">{info}</span>
```
likethemammal: In this case, using a hard width `w-24` would prob be more appropriate instead of `max-width` because otherwise, if the text is especially small, the misalignment issue would persist. This current code assumes the filename is always big/wide enough to force the span to be `w-24`

aditya-mitra: makes sense. implemented!

likethemammal (approved):

MbfloydIR (approved):

kattsushi (approved):

Tortillaguy (approved):

PR #1804, opened 2025-05-01
[IR-9743] Connect color scheme to system settings

likethemammal:
## Summary

The color scheme options arent connected to the user's system settings. By using `window.matchMedia`, the preferred color scheme can be inherited directly into our theme service. This gives the user the correct color scheme initially, but also the color scheme stays in sync when the system settings change.

https://github.com/user-attachments/assets/43a18601-5e19-464a-815f-a64bb5faa70d

## QA Steps

- Open editor
- Open system settings for changing color scheme (On Mac: Settings -> Appearance)
- Toggle between Dark or Light mode in system settings. The editor should swap the UI colors.

### DIFF
--- a/packages/client-core/src/common/services/ThemeService.tsx
+++ b/packages/client-core/src/common/services/ThemeService.tsx
@@ -299,10 +299,16 @@ export const themes: Record<string, Partial<CSSClasses>> = {
   dark: darkTheme
 }
 
+export const checkDarkMode = () => window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
+
 export const ThemeState = defineState({
   name: 'ThemeState',
-  initial: {
-    theme: 'dark' as 'light' | 'dark'
+  initial: () => {
+    const initialDarkMode = checkDarkMode()
+
+    return {
+      theme: (initialDarkMode ? 'dark' : 'light') as 'light' | 'dark'
+    }
   },
 
   setTheme: (theme: 'light' | 'dark') => {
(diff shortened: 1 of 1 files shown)


aditya-mitra (approved):
Looks fine on mac!

@CITIZENDOT can you please check this on linux?

packages/client-core/src/common/services/ThemeService.tsx
```diff
@@ -299,10 +299,14 @@ export const themes: Record<string, Partial<CSSClasses>> = {
...
 
+export const checkDarkMode = () => window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
+
+const initialDarkMode = checkDarkMode()
+
 export const ThemeState = defineState({
   name: 'ThemeState',
   initial: {
-    theme: 'dark' as 'light' | 'dark'
+    theme: (initialDarkMode ? 'dark' : 'light') as 'light' | 'dark'
```
HexaField: prefer an initializer function rather than top level declaration:

```ts

export const ThemeState = defineState({
  name: 'ThemeState',
  initial: () => ({
    theme: (initialDarkMode ? 'dark' : 'light') as 'light' | 'dark'
  }))
})

HexaField (dismissed):

packages/client-core/src/common/services/ThemeService.tsx
HexaField: sorry i mean that we should call checkDarkMode such that we aren't accessing the window at module scope, as it is this will make it hard to write tests

likethemammal: Ah I figured you meant something like that

PR #1825, opened 2025-05-05
[IR-9438] Seperate scene and render triangle count in debug info

MichaelEstes:
## References
closes #_insert number here_


HexaField (approved):
awesome

PR #1834, opened 2025-05-06
[IR-6500] Handle component serialization and deserialization to be objects instead of primitive values to be GLTF compliant

MichaelEstes:
## References
closes #_insert number here_


HexaField (changes requested):
i don't think we should do this at the loader/exporter scope, but in the components themselves with custom onSet and toJSON functions

packages/ecs/src/schemas/JSONSchemas.ts
```diff
@@ -279,5 +279,36 @@ export const S = {
     ({
       ...buildSchema('Proxy', { create: proxy, serialized: true }),
       properties: schema
-    }) as TProxySchema<T>
+    }) as TProxySchema<T>,
+
+  GLTFExtensionSchema: <T extends Schema>(schema: T): T => {
```
HexaField: i don't think we need this, the specific component serializers/deserializers alone should suffice. i don't anticipate we'll have many more than uuid and visible, virtually everything else will be an object

HexaField (changes requested):

packages/spatial/src/renderer/components/VisibleComponent.ts
```diff
@@ -28,7 +28,17 @@ import { Entity } from '@ir-engine/ecs/src/Entity'
 
 export const VisibleComponent = defineComponent({
   name: 'VisibleComponent',
-  jsonID: 'EE_visible'
+  jsonID: 'EE_visible',
+
+  onSet(entity, component, json) {
```
HexaField: i think the onSet here is redundant, we only need a toJSON

packages/engine/src/gltf/exportGLTFScene.ts
```diff
@@ -1257,7 +1257,8 @@ const exportEntity = async (
     } else {
       const compData = serializeComponent(entity, component)
       // Do we not want to serialize tag components?
-      if (!compData) continue
+      if (compData == null) continue
```
HexaField, line 1263: does serializeComponent ever return `undefined`?

packages/engine/src/gltf/NodeIDComponent.ts
```diff
@@ -44,7 +44,27 @@ export const NodeIDComponent = defineComponent({
   name: 'NodeIDComponent',
   jsonID: 'EE_uuid',
 
-  schema: S.EntityID(),
+  schema: S.EntityID({
```
HexaField: NodeIDComponent is getting merged with UUIDComponent, so this will need to be moved to that component

HexaField (changes requested):

packages/spatial/src/renderer/components/VisibleComponent.ts
HexaField: in fact, i think the visible component should return `{}` instead of `{ value: true }`, same as what other gltf extensions wih no properties do https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_unlit

HexaField: we should make `serializeComponent` return `{}` if there is no schema on the component

packages/engine/src/gltf/exportGLTFScene.ts
MichaelEstes: Ya, if the serialize flag is set to false

packages/engine/src/gltf/NodeIDComponent.ts
MichaelEstes: Looks like the UUIDComponent is now an object schema already, so it's just visible component

AidanCaruso (approved):

HexaField (approved):

PR #1842, opened 2025-05-07
IR-9742: Create Inspector panel component

svyoung:
## Summary
Create an Inspector panel component next to Properties

## References
https://tsu.atlassian.net/browse/IR-9742

## QA Steps
Go to studio editor where Properties panel is at: 

image


packages/editor/src/panels/inspector/inspectoreditor.tsx
```diff
@@ -0,0 +1,101 @@
...
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import { FiEdit2 } from 'react-icons/fi'
+import { ClickPlacementState } from '../../systems/ClickPlacementSystem'
+
+const InspectorEditor = () => {
+  const { t } = useTranslation()
+  const { metadata } = useHookstate(getMutableState(ClickPlacementState)).value
+  const { thumbnail, name, type, author, dateCreated, fileSize, dimensions, mesh, resources, tags } = metadata
+  const leftTableCellCss = `table-cell w-2/5 text-left font-bold py-0.5`
```
likethemammal: ### Table styles vs CSS Grid:
- I normally avoid using table styles, especially because they can have unintended styling results and we have better ways to style nowadays (`flex` and `grid`)
- At the very least, if youre going to use table styles like this to create a table, its better Accessiblilty, SEO, and best practice to use actual `table` elements (`table-row`, etc)
- But generally I try to use `flex` and `grid` to solve most problems like this. In this case you want the cells to match sizes, so you'd be fighting  `flex` a bit to accomplish this. I'd use `grid` and `grid-template-columns: auto auto` on the parent to create the same table behavior. You can then remove the exact width on the table cells and use the `gap-y` for spacing instead of the `py-0.5` which is messier and less clear. And you should be able to something like `justify-items: center` (or even `text-align`) to keep the content aligned to the center, like a table would.

likethemammal: 🐑 In general, exact or percentage widths are a signal the styling will break down. In a case like this I'd use `width: 100%` and `min-width` to create the same effect while also allowing for some responsive scaling. But this width property becomes redundant if we're styling with CSS grid instead (like mentioned above)

packages/editor/src/panels/inspector/inspectoreditor.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const rightTableCellCss = `table-cell text-left py-0.5`
```
likethemammal: 🐑 I'd avoid using a vertical padding on both top and bottom to space out elements. This adds extra padding on first and last element that arent nessecary. This type of styling can be achieve with a `gap-y-1` on the parent. It doesnt add extra spacing, its only a single property on 1 element, and its more obvious what the styling is doing. (Addressed with Grid styling changes mentioned above)

likethemammal:
🐑 UX: I noticed that this inspector panel doesn't stay in sync with the properties panel. Is this an expected behavior? It feels like the inspector panel should match whats being viewed in the properties panel

packages/editor/src/panels/inspector/inspectoreditor.tsx
```diff
@@ -0,0 +1,101 @@
...
+          <div className={leftTableCellCss}>{t('editor:inspector.assetMeshComplexity')}</div>
+          <div className={rightTableCellCss}>{'N/A'}</div>
+        </div>
+      </div>
+      <div className="flex flex-col gap-3">
+        <div className="font-bold">{t('editor:inspector.assetTags')}</div>
+        <div className="text-lg">
+          {tags.length > 0
+            ? tags.map((tag) => (
+                <span className="rounded-[30px] border border-[#42454D] bg-surface-2 px-3 py-1">{tag}</span>
```
likethemammal: Use `rounded-full` instead of a custom px. 1) because its cleaner, but 2) `px` dont scale with `rem`, so in some cases the design would break down and 30px wouldn't create a completely rounded pill shape

packages/editor/src/panels/inspector/inspectoreditor.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const { thumbnail, name, type, author, dateCreated, fileSize, dimensions, mesh, resources, tags } = metadata
+  const leftTableCellCss = `table-cell w-2/5 text-left font-bold py-0.5`
+  const rightTableCellCss = `table-cell text-left py-0.5`
+
+  return Object.keys(metadata).length > 0 ? (
+    <div className="flex h-full flex-col gap-6 overflow-y-auto bg-surface-3 p-3 px-10 text-text-secondary">
+      <div className="align-center flex justify-center">
+        <img src={thumbnail} alt={name} className="m-3 text-center" />
+      </div>
+      <div className="flex flex-row items-center justify-between">
```
likethemammal: This could use a `gap-x-1` to create space between the text and icon

image

packages/editor/src/panels/inspector/inspectoreditor.tsx
likethemammal: With the above Grid changes, you should be able to add a `gap-x` to put some horizontal space between the metadata. They could use some spacing when the panel is small

image

packages/editor/src/panels/inspector/inspectoreditor.tsx
```diff
@@ -0,0 +1,101 @@
...
+          <div className={leftTableCellCss}>{t('editor:inspector.assetAuthor')}</div>
+          <div className={rightTableCellCss}>{author || 'N/A'}</div>
+        </div>
+        <div className="table-row">
+          <div className={leftTableCellCss}>{t('editor:inspector.assetDateCreated')}</div>
+          <div className={rightTableCellCss}>{toDisplayDateTime(dateCreated) || 'N/A'}</div>
+        </div>
+        <div className="table-row">
+          <div className={leftTableCellCss}>{t('editor:inspector.assetFileSize')}</div>
+          <div className={rightTableCellCss}>{'N/A'}</div>
```
likethemammal: 🐑 Is there a plan for the hardcoded "N/A"s?

likethemammal (changes requested):
Sorry for all the change requests. I've noticed we dont have much standards around styling and I'd like to make sure we're more purposeful with how we style.

🐑 = I'm saying something redundant or not critical

packages/editor/src/panels/inspector/inspectoreditor.tsx
DanielBelmes: Not yet. We have dimensions done. But the rest we have to rewrite the dimensions setup to collect more metadata for our files. In general I don't super think this needs to be on the client either and would love a better ingest mechanism for assets into the CSM.

svyoung:
> 🐑 UX: I noticed that this inspector panel doesn't stay in sync with the properties panel. Is this an expected behavior? It feels like the inspector panel should match whats being viewed in the properties panel

What do you mean by in sync? Currently, the user flow is still in the works. Check out the Figma design https://www.figma.com/design/DXwFHGa5g0peqqSjyjpxQz/Studio-Design?node-id=13011-21452&p=f&m=dev

packages/editor/src/panels/inspector/inspectoreditor.tsx
svyoung: @DanielBelmes @likethemammal should we not show at all until we get the data?

DanielBelmes (changes requested):
I'd like a storybook so that the designers can review the changes in chromatic.

packages/editor/src/panels/inspector/inspectoreditor.tsx
likethemammal: N/A is fine for me. I think a better UX would prob be "--". Just cus N/A suggests that the information is actually not relevant, even tho the user could think differently.

likethemammal:
> > 🐑 UX: I noticed that this inspector panel doesn't stay in sync with the properties panel. Is this an expected behavior? It feels like the inspector panel should match whats being viewed in the properties panel
> 
> What do you mean by in sync? Currently, the user flow is still in the works. Check out the Figma design https://www.figma.com/design/DXwFHGa5g0peqqSjyjpxQz/Studio-Design?node-id=13011-21452&p=f&m=dev

Ah okay. I see this all relates to Assets/Files. I was just noticing I would click in the scene on different elements and was expecting the Inspector panel to update with the relevant metadata, like how the Properties panel updates with the relevant components.

But it may be complicated to do and it doesnt look like its part of the UX plan

packages/editor/src/panels/inspector/inspectoreditor.tsx
```diff
@@ -0,0 +1,103 @@
...
+      <div className="align-center flex justify-center">
+        <img src={thumbnail} alt={name} className="m-3 text-center" />
+      </div>
+      <div className="flex flex-row items-center justify-between gap-x-4 break-all">
+        <Text fontSize="xl">{name}</Text>
+        <div>
+          <FiEdit2 />
+        </div>
+      </div>
+      <div className="flex flex-col">
```
likethemammal:
```suggestion
      <div className="grid grid-cols-[auto_auto] justify-center gap-y-1 text-left">
```

If the left column looks a little large or should be centered, `grid-cols-[max-content_auto]` should also work

packages/editor/src/panels/inspector/inspectoreditor.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { useTranslation } from 'react-i18next'
+import { FiEdit2 } from 'react-icons/fi'
+import { ClickPlacementState } from '../../systems/ClickPlacementSystem'
+
+const InspectorEditor = () => {
+  const { t } = useTranslation()
+  const { metadata } = useHookstate(getMutableState(ClickPlacementState)).value
+  const { thumbnail, name, type, author, dateCreated, fileSize, dimensions, mesh, resources, tags } = metadata
+  const rowCss = `flex flex-row gap-y-1`
+  const leftCellCss = `w-[40%] text-left font-bold`
```
likethemammal: Using an exact percentage like this will break the design on certain panel sizes. Thats why I mentioned `grid` over `flex`, because `grid` will inherit the width from the children in that column, where `flex` will not have a fixed size for the columns.

By switching to grid like suggested below, this line can become:

```suggestion
  const leftCellCss = `font-bold`
```

packages/editor/src/panels/inspector/inspectoreditor.tsx
```diff
@@ -0,0 +1,103 @@
...
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import { FiEdit2 } from 'react-icons/fi'
+import { ClickPlacementState } from '../../systems/ClickPlacementSystem'
+
+const InspectorEditor = () => {
+  const { t } = useTranslation()
+  const { metadata } = useHookstate(getMutableState(ClickPlacementState)).value
+  const { thumbnail, name, type, author, dateCreated, fileSize, dimensions, mesh, resources, tags } = metadata
+  const rowCss = `flex flex-row gap-y-1`
```
likethemammal: With the grid changes below, this variable and line arent needed.
```suggestion
```

packages/editor/src/panels/inspector/inspectoreditor.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const leftCellCss = `w-[40%] text-left font-bold`
+  const rightCellCss = `text-left`
```
likethemammal: With the grid changes below, this variable and line arent needed.
```suggestion
```

packages/editor/src/panels/inspector/inspectoreditor.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <Text fontSize="xl">{name}</Text>
+        <div>
+          <FiEdit2 />
+        </div>
+      </div>
+      <div className="flex flex-col">
+        <div className={rowCss}>
+          <div className={leftCellCss}>{t('editor:inspector.assetType')}</div>
+          <div className={rightCellCss}>{type}</div>
+        </div>
```
likethemammal: The way `grid` works, it takes the raw set of children and arranges them into columns and rows. That means these row divs arent necessary and will break how the `grid` styling works. If you remove these row divs the `grid` styling will work.
```suggestion
        <div className={leftCellCss}>{t('editor:inspector.assetType')}</div>
        <div>{type}</div>
```

Doing this for each row below as well..

likethemammal (changes requested):
The `w-[40%]` is still a signal that the design will break down. Heres an implementation with `grid` that wont have the same problems that `flex` has.

likethemammal (approved):
Looks great!

kattsushi (approved):
LGTM

MbfloydIR (approved):

packages/editor/src/panels/inspector/inspectoreditor.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { toDisplayDateTime } from '@ir-engine/common/src/utils/datetime-sql'
+import { getMutableState, useHookstate } from '@ir-engine/hyperflux'
+import Text from '@ir-engine/ui/src/primitives/tailwind/Text'
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import { FiEdit2 } from 'react-icons/fi'
+import { ClickPlacementState } from '../../systems/ClickPlacementSystem'
+
+const InspectorPanel = () => {
+  const { metadata } = useHookstate(getMutableState(ClickPlacementState)).value
```
Tortillaguy, line 37: I think `useState` is also applicable here. (Exported from the same `hyperflux` module)

Tortillaguy (commented):
I can help you with loading image assets into storybook. Otherwise the PR looks good to me 👍

Tortillaguy (approved):

packages/editor/src/panels/inspector/inspectoreditor.tsx
SYBIOTE: We prefer hookstate use over usestate

SYBIOTE (approved):

DanielBelmes (approved):

packages/editor/src/panels/inspector/inspectoreditor.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { FiEdit2 } from 'react-icons/fi'
+import { EditorState } from '../../services/EditorServices'
+import { ClickPlacementState } from '../../systems/ClickPlacementSystem'
+
+const InspectorPanel = () => {
+  const { metadata } = useHookstate(getMutableState(ClickPlacementState)).value
+  return <InspectorEditor data={metadata} />
+}
+
+export const InspectorEditor = ({ data }) => {
```
HexaField, line 41: Please ensure all variables have types

packages/editor/src/services/EditorServices.ts
```diff
@@ -40,6 +40,8 @@ export enum UIMode {
   ADVANCED = 'ADVANCED'
 }
 
+export type activeLowerPanel = 'properties' | 'inspector'
```
HexaField, line 43: Our standards require type interfaces (and all definitions) to be PascalCase

packages/editor/src/systems/ClickPlacementSystem.tsx
```diff
@@ -62,6 +62,25 @@ import { SelectionState } from '../services/SelectionServices'
 import { ObjectGridSnapState } from './ObjectGridSnapSystem'
 
 let placedCount = 0
+
+type AssetTag = string
```
HexaField, line 66: If this is a strict type, please use OpaqueType

packages/ui/.storybook/main.ts
```diff
@@ -55,6 +56,7 @@ const config: StorybookConfig = {
     getAbsolutePath('storybook-addon-react-router-v6'),
     getAbsolutePath('storybook-addon-sass-postcss')
   ],
+  staticDirs: ['../public'],
```
HexaField, line 59: if this file is just for tests, please put them in a 'tests' folder so they can be ignored from the docker image

packages/ui/src/components/editor/InspectorPanel/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+import { toDisplayDateTime } from '@ir-engine/common/src/utils/datetime-sql'
+import { EditorState } from '@ir-engine/editor/src/services/EditorServices'
+import { Checkbox } from '@ir-engine/ui'
+import Text from '@ir-engine/ui/src/primitives/tailwind/Text'
+import React, { useEffect, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { FiEdit2 } from 'react-icons/fi'
+
+export const InspectorEditor = ({ data }) => {
```
HexaField, line 34: it doesn't look like this file is used at all, and that it is a duplicate of the file in the editor folder, please make sure these are deduplicated

PR #1843, opened 2025-05-07
[IR-9927] [Release 1.0.2] Customer built space no longer loads properly after 1.0.2 Hotfix Deployed

MichaelEstes:
## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/physics/classes/Physics.ts
+++ b/packages/spatial/src/physics/classes/Physics.ts
@@ -51,7 +51,13 @@ import {
   Vector3
 } from 'three'
 
-import { getComponent, getOptionalComponent, hasComponent, setComponent } from '@ir-engine/ecs/src/ComponentFunctions'
+import {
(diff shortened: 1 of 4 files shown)


packages/spatial/src/physics/classes/Physics.ts
```diff
@@ -676,7 +682,7 @@ function attachCollider(
   rigidBodyEntity: Entity,
   colliderEntity: Entity
 ) {
-  if (world.Colliders.has(colliderEntity)) return
+  if (world.Colliders.has(colliderEntity) || !entityExists(colliderEntity)) return
```
SamMazerIR, line 685: should we swap this order of condition checks? I feel like it might not be needed but would prevent the physics lookup if an entity doesn't exist

SamMazerIR (approved):

DanielBelmes (approved):

PR #1850, opened 2025-05-08
[IR-9970] Add collapsible overflow menu to Viewport toolbar

likethemammal:
## Summary

This updates the Viewport toolbar to conditionally show the toolbar items as they are hidden in a separate Popup menu. This allows the Viewport panel to be resized and still usable.

The logic uses IntersectionObserver to check if the toolbar items are being hidden behind other elements, and then uses the `collapse` class to hide them without remounting the DOM elements.

https://github.com/user-attachments/assets/9c23c619-fefa-492e-abec-480860efa5ba

## References
- https://javascript.plainenglish.io/building-collapsible-menu-in-react-with-intersection-observer-64f355df9f0c
- https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

## QA Steps

- Open the editor
- Resize the Viewport panel
- The Toolbar items should disappear and an icon for the extra context menu should appear. The items that disappeared should be in the context menu


MbfloydIR (approved):

kattsushi (approved):
LGTM, but please take care of conflicts

SYBIOTE:
i love this update so much

PR #1854, opened 2025-05-09
Three 176

AidanCaruso:
## Summary

- Updates threejs and dependent packages (mesh-bvh, troika-text, quarks)
- Updates corresponding patches for threejs, including multiview support
- Updates patches for postprocessing adding ambient occlusion smoothing pass
- Refactors and simplifies CSM to the CSMComponent, no longer an instance of a class, uses material plugins, queries for material entities, now works for material arrays, and saves shadow types - so no more pixelated shadows.
- Updates the default project scenes to better utilize CSM, post processing, lighting
- Removes some old unused code, including TrailRenderer, and the old, never-functional lightmap baker

## References
closes https://tsu.atlassian.net/browse/IR-8472, https://tsu.atlassian.net/browse/IR-10043


packages/engine/tsconfig.json
```diff
@@ -26,5 +26,5 @@
     "declaration": false,
     "types": ["@types/node"]
   },
-  "include": ["../../__global.d.ts", "./src", "./index.ts"]
+  "include": ["../../__global.d.ts", "./src", "./index.ts", "../projects/AidanCaruso/terrainGen/src/TerrainComponent.ts", "../projects/AidanCaruso/terrainGen/src/systems/TerrainSystem.tsx"]
```
HexaField: Tsconfig lel

AidanCaruso: Oh my gosh how did this happen

AidanCaruso: Probably augment over reaching

HexaField: it can happen with auto imports when moving files around

AidanCaruso:
This is technically "ready for review" but I'm not marking as such until I update our projects

AidanCaruso:
Projects I've tested all look good so far, marking as ready.

HexaField (commented):
Screenshot 2025-05-26 at 11 47 33 am

Looking fantastic, the only thing I've noticed is *maybe* a change in the behaviour with directional lights. As pictured it seems to be a bit brighter than usual, though this might not be a regression and just a change in how the CSM works.

AidanCaruso:
> Screenshot 2025-05-26 at 11 47 33 am
> 
> Looking fantastic, the only thing I've noticed is _maybe_ a change in the behaviour with directional lights. As pictured it seems to be a bit brighter than usual, though this might not be a regression and just a change in how the CSM works.

In that screenshot it looks as if the CSM shader logic is not being injected into the materials in question. I just realized I forgot to add proper material entity logic to procedural primitive geometry so it won't get CSM set up properly either. 

EDIT: We need to go through the engine and ensure that we're always creating material entities for mesh materials. That's probably best done in a future PR with its own scope.

In general a dead giveaway of this (or any issue preventing csm logic shader injection) is directional lighting being too bright as all the cascades are basically drawn on top of each other.

HexaField (approved):

dinomut1 (approved):

CODEOWNERS
```diff
@@ -40,7 +40,6 @@
 /packages/spatial/src/renderer/ @AidanCaruso @DanielBelmes @MichaelEstes
 /scripts/ @barankyle @hanzlamateen @DanielBelmes
 /patches/postprocessing+6.34.1.patch @DanielBelmes
-/patches/three+0.158.0.patch @DanielBelmes
```
HexaField, line 43: please correct the codeowners for all patch files affected

DanielBelmes:
Video is appearing as a white box even when playing
image

Particle system wasn't working. I believe it has a default state and it wasn't spewing anything

The biggest issue I am seeing is that the highlight effect is persisting on the screen in the editor. If you click on the platform and then click something else. The last drawn effect between clicking on something new persists on the screen as a ghostly outline.
image

Also everything seems to be being highlighted
image

DanielBelmes:
Seeing an occasional race condition where lighting doesn't load. An odd way I can guarantee reproducing it is to quickly press "Files" when loading the scene 🤷🏼‍♂️ 
image

AidanCaruso:
@DanielBelmes I've fixed the highlight problems. 

Particle systems work fine for me. Spitting out particles on add and all.

Media component visualizers being white boxes was busted on dev, maybe with the UI visualization PR that was merged into dev previously (just a guess). Either way I've pushed a temporary fix to this PR that clears the white material so they're at least visible now for testing.

I've tried many times but was unable to reproduce the directional light race condition you're seeing, which is odd.

AidanCaruso:
Currently the server-core tests are failing - this is an issue in dev right now, unrelated to this branch. Seems all other relevant checks are passing.

AidanCaruso:
I actually did nail down the lighting issue @DanielBelmes was seeing - or at least I think so. I found an issue related to csm unmount component removal before calling setComponent not clearing out previous values. Don't know why but overwriting the properties works for now.

DanielBelmes (approved):
Yup all the bugs I found are fixed. Thank you

PR #1864, opened 2025-05-11
fix(prefab): load the new exported prefab [IR-9728]

mikeplascdev:
## Summary
- Add missing words for the prefab dialog
- Add loading state to avoid creating multiple prefabs
- Validate the object from the UUID component

## References
closes #_insert number here_

## QA Steps
1. Open studio project
2. Create new scene
3. Select Files menu dropdown > Export Lookdev Prefab
4. Name custhe tom prefab and confirm
5. Find the customized prefab in Files tab
6. Drag prefab into viewport

## Evidence
https://github.com/user-attachments/assets/f342ef7a-536e-471d-96f4-1a30b3558f39

## Jira
[IR-9728](https://tsu.atlassian.net/browse/IR-9728)

[IR-9728]: https://tsu.atlassian.net/browse/IR-9728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ecs/src/UUIDComponent.ts
```diff
@@ -186,7 +188,11 @@ export const UUIDComponent = defineComponent({
 
   /** Reactively gets an entity from the same source by ID */
   useEntityFromSameSourceByID(entity: Entity, id: EntityID, layer = Layers.Simulation as LayerID) {
+    if (!entity || entity === UndefinedEntity) return UndefinedEntity
```
HexaField: This will cause reactivity problems. React contexts must always run the same number of hooks each render. Early outs like this will break any react contexts that use them.

packages/ecs/src/UUIDComponent.ts
```diff
@@ -203,8 +209,12 @@ export const UUIDComponent = defineComponent({
   },
 
   useSourceEntity(entity: Entity) {
+    if (!entity || entity === UndefinedEntity) return UndefinedEntity
```
HexaField: Same here

packages/ecs/src/UUIDComponent.ts
```diff
@@ -227,7 +237,11 @@ export const UUIDComponent = defineComponent({
   use: (entity: Entity) => UUIDComponent.join(useComponent(entity, UUIDComponent).value),
 
   /** Joins an EntityUUIDPair into a string */
-  join: (idPair: EntityUUIDPair) => `${idPair.entitySourceID}${idPair.entityID}` as EntityUUID,
+  join: (idPair: EntityUUIDPair) => {
```
HexaField: These checks seem odd, we should be relying on tyoe safety here, if the value passed in is not valid, we should find what situation is causing this, rather than putting these early returns in. This will likely cause silent failures that will make it harder to find when assumptions are being made.

HexaField (changes requested):

packages/ecs/src/UUIDComponent.ts
mikeplascdev: done 4a9d4f728f546766e8ab13ec65283a469d058f54

packages/ecs/src/UUIDComponent.ts
mikeplascdev: done 4a9d4f728f546766e8ab13ec65283a469d058f54

packages/ecs/src/UUIDComponent.ts
mikeplascdev: done 4a9d4f728f546766e8ab13ec65283a469d058f54

mikeplascdev:
@HexaField, I have already reverted the validation changes
In the `overrideExporterExtension`, I validate that the entity is in valueMap

1d52ca78e06881bac65f280852dbfa15373f0625

Tortillaguy (approved):

HexaField (approved):

PR #1868, opened 2025-05-12
IR-9309 Add support of copy embed code in the publish modal

mshafiqmk:
## Summary
image

## References
[IR-9309](https://tsu.atlassian.net/browse/IR-9309)

## QA Steps

[IR-9309]: https://tsu.atlassian.net/browse/IR-9309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
```diff
@@ -611,6 +612,20 @@ export default function AddEditLocationModal(props: AddEditLocationModalProps) {
...
               </div>
             </div>
+            {scene.value && scenes.status === 'success' && (
+              <div className="border-t border-t-ui-outline py-6">
+                <EmbedCodeField
+                  url={(() => {
+                    const sceneResource = scenes.data.find((s) => s.id === scene.value)
+                    if (sceneResource) {
+                      const sceneName = sceneResource.key.split('/').pop()!.replace('.gltf', '')
+                      return `${config.client.clientUrl}/location/${sceneName}`
```
hanzlamateen: This url wouldn't work for MT scenarios. Use `location.url` instead. You might need to change above conditions accordingly.

packages/client-core/src/common/components/EmbedCodeField.tsx
```diff
@@ -0,0 +1,79 @@
...
+export const EmbedCodeField = ({
+  url,
+  containerClassName = 'text-text-secondary',
+  className = 'h-20 border-ui-tertiary bg-[#F6F8FA] dark:border-ui-outline dark:bg-surface-2',
+  labelClassname = 'text-xs text-text-secondary',
+  showLabel = true
+}: Props) => {
+  const { t } = useTranslation()
+
+  const embedCode = `<iframe style="border: 1px solid rgba (0, 0, 0, 0.1);" width="375" height="687" src="${url}" allowfullscreen></iframe>`
```
hanzlamateen: I am just wondering if the `EmbedCodeField` component can be kept generic and the iframe part can be created in the parent component. Also, the `url` prop can be renamed to be generic like content/text/code.

packages/client-core/src/common/components/popovers/CopyEmbedCodePopover.tsx
```diff
@@ -38,21 +36,6 @@ type Props = Readonly<{
 export const CopyEmbedCodePopover = ({ url }: Props) => {
```
hanzlamateen: Just wondering if `packages/ui/src/components/tailwind/InputDialog/index.tsx` InputDialog component can be extended to support codefields and then used instead of this new dedicated popover?

hanzlamateen (changes requested):

packages/client-core/src/common/components/EmbedCodeField.tsx
mshafiqmk: We can do that, but it would require creating the iframe three times in the parent components. In my opinion, the better approach is to use the reusable component we already have, since the exact same iframe is needed in three different places.

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
mshafiqmk: in MT we have different logic and i have used `locaction.url`

packages/client-core/src/common/components/EmbedCodeField.tsx
HexaField: there are other properties we need in the iframe
see https://github.com/ir-engine/ir-engine/blob/dev/packages/client-core/src/user/menus/ShareMenu.tsx#L168

HexaField (changes requested):

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
mshafiqmk: I'm using `location.url`, so we'll display this component only if the scene is published. If the scene isn't published, `location.url` will be undefined, and the component won't be shown.

packages/client-core/src/common/components/popovers/CopyEmbedCodePopover.tsx
mshafiqmk: I think we'll need to reuse all the logic from the Embed component in the InputDialog. In my opinion, it's better to keep the Embed component separate, since we might use it without a dialog — for example, in the Publish modal.

hanzlamateen: I think all we need is an if condition in InputDialog

mshafiqmk: DONE

hanzlamateen (approved):

PR #1876, opened 2025-05-13
[IR-8941] Enum schemas should save enum values instead of index

MichaelEstes:
## References
closes #_insert number here_


SYBIOTE (approved):
Reject enum !!
Return to const

HexaField (changes requested):
I think we should keep the Enum lingo since that is how JSON Schema defines this data type

https://json-schema.org/understanding-json-schema/reference/generic

MichaelEstes:
> I think we should keep the Enum lingo since that is how JSON Schema defines this data type
> 
> https://json-schema.org/understanding-json-schema/reference/generic

That is objectively not an enum, but sure, we can go back to enum

HexaField (changes requested):
Screenshot 2025-05-20 at 1 58 14 pm

In the test scene all geometries load as boxes, which likely indicates they are not migrating properly.

packages/engine/src/scene/components/PrimitiveGeometryComponent.ts
```diff
@@ -51,10 +51,10 @@ export const PrimitiveGeometryComponent = defineComponent({
   jsonID: 'EE_primitive_geometry',
 
   schema: S.Object({
-    geometryType: S.Enum(GeometryTypeEnum, {
+    geometryType: S.Enum(GeometryType, {
       $comment:
```
HexaField, line 55: oh, we also need to update the comments in each enum to not say it's a numerical value

HexaField (approved):

PR #1877, opened 2025-05-14
feat(IR-10053): disable compress and publish feature

kattsushi:
## Summary
feat(IR-10053): disable compress and publish feature
## References
closes #_insert number here_


BogieBogard (approved):
Approved, looks like a straightforward implementation of disabling the feature based on feature flags.

PR #1904, opened 2025-05-16
Rel1.0.3 Add a large fallback to kill field height

DanielBelmes:
## Summary
If the scene settings component somehow gets deleted this functionality didn't have a fallback. So I added a large  one. This does limit the build depth of the scene though. I also considered a large velocity check but I'd really like to avoid users falling for a long time in any instance.

## References
closes [IR-10172
](https://tsu.atlassian.net/browse/IR-10172)


BogieBogard (approved):
Approved. I confirmed the fallback works when the scene settings are deleted. The code changes look great too--targeted, minimal change to fix the problem.

PR #1908, opened 2025-05-16
update the spawn point avatar random placement logic

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-9412
## References
closes #_insert number here_


packages/engine/src/avatar/functions/getSpawnPoint.ts
```diff
@@ -51,7 +51,12 @@ export function getSpawnPoint(spawnPointNodeId: string, userId: UserID): { posit
 }
 
 const randomPositionCentered = (area: Vector3) => {
-  return new Vector3((Math.random() - 0.5) * area.x, (Math.random() - 0.5) * area.y, (Math.random() - 0.5) * area.z)
+  const radius = 0.5
+  const x = (Math.random() - 0.5) * 2 * radius + area.x
+  const z = (Math.random() - 0.5) * 2 * radius + area.z
+  console.log('mbf', area)
```
kattsushi: LGTM just remove the logs

kattsushi (approved):

jose-galvan (approved):

HexaField (changes requested):
we need to use the scale of the spawn point entity in order to decide the radius, otherwise we can't change the size of the spawn area

SYBIOTE (dismissed):
Connect radius to scale, for increasing spawn size, im assuming spawn points are cylindrical now

HexaField:
> Connect radius to scale, for increasing spawn size, im assuming spawn points are cylindrical now

they're not, we don't have a specification for that. really it should be labelled size not radius

Tortillaguy (approved):

packages/engine/src/avatar/functions/getSpawnPoint.ts
MbfloydIR: implemented


> LGTM just remove the logs

MbfloydIR:
> > we need to use the scale of the spawn point entity in order to decide the radius, otherwise we can't change the size of the spawn area

implemented

packages/engine/src/avatar/functions/getSpawnPoint.ts
```diff
@@ -69,15 +71,15 @@ export function getRandomSpawnPoint(userId: UserID): { position: Vector3; rotati
     const spawnTransform = getComponent(entity, TransformComponent)
     const worldPosition = TransformComponent.getWorldPosition(entity, new Vector3())
     return {
-      position: randomPositionCentered(worldPosition),
+      position: randomPositionCentered(worldPosition, spawnTransform.scale.clone()),
```
HexaField: we don't need to clone any of these values if they aren't being mutated

HexaField (dismissed):

packages/engine/src/avatar/functions/getSpawnPoint.ts
MbfloydIR: implemented

PR #1910, opened 2025-05-17
IR-10194 Add support for mt-prdmirr, mt-nightly, and mt-weekly app hosts

sfgarza:
## Summary
* Added environment detection for production mirror (mt-prdmirr), nightly build (mt-nightly), and weekly build (mt-weekly) deployments
* Updated builder scripts to properly handle these environments in GCP artifact repositories


infinitelos (approved):

HexaField (commented):
Why are we hardcoding these at all? For the engine to be fully reusable and self deployable we can't have these hard coded

sfgarza:
> Why are we hardcoding these at all? For the engine to be fully reusable and self deployable we can't have these hard coded

@HexaField These should get updated to be self deployable however right now are needed to match the quirks in the naming conventions of the artifact repositories created by the terraform scripts

PR #1912, opened 2025-05-17
revert material instance entity to id

HexaField:
## References
closes #_insert number here_


HexaField:
rebased here https://github.com/ir-engine/ir-engine/pull/1925

PR #1916, opened 2025-05-18
Fix: improve entity label rendering to handle new UUIDComponent

HexaField:
## References
closes #_insert number here_


packages/client-core/src/components/Debug/EntityDebug.tsx
```diff
@@ -80,7 +84,9 @@ const renderEntityTree = (entity: Entity) => {
             (r, child) =>
               Object.assign(r, {
                 [`${child} - ${
-                  getOptionalComponent(child, NameComponent) ?? getOptionalComponent(child, UUIDComponent)
+                  getOptionalComponent(child, NameComponent) ?? hasComponent(child, UUIDComponent)
```
AidanCaruso: This is causing errors when a component posesses a name component without a uuid component. The case for which UUIDComponent.get() is called should be only if the NameComponent does not exist and the UUIDComponent does exist.

AidanCaruso (dismissed):

the-reality-engineer (approved):

PR #1917, opened 2025-05-18
[v1.2] Upgrade typescript to 5.8 and enable erasable syntax only

HexaField:
## References
closes #_insert number here_


HexaField:
Marking as draft as we have ruled this potentially introducing too many regressions, will merge once v1.1 is cut.

PR #1922, opened 2025-05-19
Update gizmo visibility logic based on active helper component state

SYBIOTE:
Revise the gizmo visibility logic to consider the state of the active helper component, ensuring that visibility is only set when both editing is active and the helper component is enabled.


packages/editor/src/systems/ActiveHelperSystem.tsx
```diff
@@ -139,7 +140,7 @@ const ActiveHelperReactor = () => {
       }
       return iconGizmo
     },
-    activeHelperComponent.enabled.value,
+    editorHelperState.gizmoEnabled && activeHelperComponent.enabled.value,
```
HexaField: this needs to be `editorHelperState.gizmoEnabled.value`

HexaField (dismissed):

achen5671 (approved):

PR #1923, opened 2025-05-19
add asset processing system in FileThumbnailJobState

JT00y:
https://tsu.atlassian.net/browse/IR-9900


junhua-ir (approved):
Looks good. Tested on local machine and generated multi-views successfully.

PR #1924, opened 2025-05-20
IR-6715 Fix the Environment Map Bake component

koktavy:
## Summary
Restores envmap bake functionality. The "probe" option for the envmap settings dropdown is to be removed in [#7898](https://tsu.atlassian.net/browse/IR-7898).

## Subtasks / QA Checklist
- [ ] Respects the entity position
- [ ] Respects the Position Offset
- [ ] Respects the Scale
- [ ] Respects the Bake Resolution
- [ ] Respects the Box Projection toggle
- [ ] On save and refresh, envmap is positioned to match last bake

## Breaking Changes
N/A, restored functionality

## References
closes [#6715](https://tsu.atlassian.net/browse/IR-6715)


packages/editor/src/functions/uploadEnvMapBake.ts
```diff
@@ -71,46 +80,31 @@ const getScenePositionForBake = (entity?: Entity) => {
...
   if (!envmap) return null!
 
   const nameComponent = getComponent(entity, NameComponent)
   const editorState = getState(EditorState)
   const sceneName = editorState.sceneName!
   const projectName = editorState.projectName!
-  const filename = isSceneEntity ? `${sceneName}.envmap.ktx2` : `${sceneName}-${nameComponent.replace(' ', '-')}.ktx2`
+  const timestamp = Date.now()
+  // Add a cache-busting timestamp to the filename to ensure the envmap updates each bake
+  const filename = isSceneRootEntity
```
HexaField, line 103: we have cache busting already with the content hash returned on the url from static resources, which should be returned once it is uploaded. we should rework this such that once the upload is complete, we then set the EnvMapBakeComponent with the returned url with the hash

HexaField (changes requested):

packages/editor/src/functions/uploadEnvMapBake.ts
koktavy: Ah ok, existing code had the hash and search params cleared. Now updated to keep those in

HexaField (approved):

PR #1927, opened 2025-05-20
Fix sync local storage not always syncing state

HexaField:
## References
closes #_insert number here_


the-reality-engineer:
This change breaks the LoopAnimationComponent test.

HexaField:
> This change breaks the LoopAnimationComponent test.

It doesn't just had to run the test again. Still some non-determinism.

MichaelEstes (approved):

PR #1931, opened 2025-05-20
IR-10146 mt stg user spawns beneath the floor

quaiquai-ir:
## References
https://tsu.atlassian.net/browse/IR-10146

## QA Steps

- Create a new scene using the GLTF asset model in the ticket.
- Publish Scene
- Go to published location


MichaelEstes (approved):

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -164,6 +164,12 @@ const loadDependencies = {
       key: 'dependencies',
       eval: (dependencies?: ComponentDependencies) => componentDependenciesLoaded(dependencies)
     }
+  ],
+  ['EE_collider']: [
```
HexaField: For both this jsonID and the other existing once, we should pull them directly from their component definitions (eg `ColliderComponent.jsonID`)

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -164,6 +164,12 @@ const loadDependencies = {
...
+    {
+      key: 'shape',
+      eval: (shape) => !!shape
```
HexaField: I understand why this helps, but I think we want to go further and ensure that the collider is actually added to the physics world (and only when there is an ancestor with a rigidbody to avoid getting stuck) to evaluate that the physics world has actually created a collider for this entity.

HexaField (changes requested):

packages/engine/src/gltf/GLTFComponent.tsx
quaiquai-ir: If im understanding this correctly at the point of loading these dependencies the source componenet doesnt have access to a loaded physical world yet.  As this code doesn't allow me to get the physics world:
```
eval: (_, entity: Entity) => {
        const world = Physics.getWorld(entity)
        if (!world) return false
        
        const rigidbodyEntity = getAncestorWithComponents(entity, [RigidBodyComponent])
        if (!rigidbodyEntity) return true // Skip if no rigidbody ancestor
        
        return world.Colliders.has(entity) // Check if collider was actually created
      }
```

Could we add those colliders to a state list and check later if they have been applied to the physics world?

quaiquai-ir: I added a useCheckLoadedColliders() that checks the hoisted hasColliders variable in the ColliderComponents scheme before fully loading the GLTFComponent. If all loaded, it triggers GLTFComponent collidersLoaded variable to true.

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -159,11 +160,19 @@ const componentDependenciesLoaded = (dependencies?: ComponentDependencies) => {
...
+  [GLTFComponent.jsonID]: [
     {
       key: 'dependencies',
       eval: (dependencies?: ComponentDependencies) => componentDependenciesLoaded(dependencies)
     }
+  ],
+  [ColliderComponent.jsonID]: [
+    {
+      key: 'hasCollider',
+      eval: (hasCollider: boolean) => {
```
HexaField: just a small nitpick, let's inline this for brevity: 

`eval: (hasCollider: boolean) => hasCollider`

HexaField (changes requested):

HexaField (approved):

PR #1932, opened 2025-05-20
Storybook Additional Stories

Tortillaguy:
## References
closes #_insert number here_


package.json
```diff
@@ -169,5 +191,10 @@
     "patch-package": "^6.5.1",
     "ts-node": "10.9.2",
     "typescript": "5.6.3"
+  }, 
+  "eslintConfig": {
+    "extends": [
+      "plugin:storybook/recommended"
```
DanielBelmes: Should configure this in eslint.config.mjs

packages/ui/types.d.ts
```diff
@@ -0,0 +1,35 @@
...
+
+The Original Developer is the Initial Developer. The Initial Developer of the
+Original Code is the Infinite Reality Engine team.
+
+All portions of the code written by the Infinite Reality Engine team are Copyright © 2021-2025
+Infinite Reality Engine. All Rights Reserved.
+*/
+
+
+declare module '*.glb?url' {
```
DanielBelmes: just adding that we can put this in __global.d.ts. easier to have them in one place

packages/ui/public/loader_decoders/basis/basis_transcoder.js
DanielBelmes, line 1: I wonder if we can include some of these with staticDirs so they're not duplicated

likethemammal (approved):

PR #1933, opened 2025-05-20
Remove reflection probes implementation

koktavy:
## Summary
Since environment probes are [currently broken](https://tsu.atlassian.net/browse/IR-7898) and a larger discussion/decision is needed, this PR simply removes the UI option for probes from the EnvMap component.

## Subtasks Checklist
Comments included with links to ticket and discussion for future reference.

## Breaking Changes
Not that I found, unless probes are expected to be available for use by other projects.

## References
Closes [IR-7939](https://tsu.atlassian.net/browse/IR-7939) as a bandaid fix for [IR-7898](https://tsu.atlassian.net/browse/IR-7898)

## QA Steps
Add an Environment Map component to the scene and observe that there is no longer a Probes option.
image

[IR-7939]: https://tsu.atlassian.net/browse/IR-7939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IR-7898]: https://tsu.atlassian.net/browse/IR-7898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HexaField (changes requested):
i think we should just get rid of it as an option and an implementation altogether. if we need it again for some reason, we can get the code from this PR, but we have no plans so let's just remove it.

HexaField (approved):

PR #1943, opened 2025-05-21
Optimize flow should limit texture size to 1K

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-10160


BogieBogard (approved):
Approved. Nice fix, I think the previous code could have chosen the wrong extension if the URL had a query param or something like that. This update should more reliably pick the right extension.

I know some devs aren't fans of nested ternaries. But that was existing and it's only two so that's probably fine to leave as-is. We could add a "no-nested-ternary" eslint rule if we want to disallow that in the future.

hzraslan:
@BogieBogard Thanks! Glad the fix makes sense.Totally hear you on the nested ternaries. I’d be on board with adding a linter rule if we want to standardize that going forward.

PR #1945, opened 2025-05-21
Add Type naming conventions to Eslint Config

DanielBelmes:
## Summary
Given that types typically don't have a naming convention recommend in typscript for us to maintain ours we need a custom eslint configuration rule to adhere to our standard.


MichaelEstes (approved):

HexaField:
Awesome!

PR #1947, opened 2025-05-22
IR-9980: Fix lock and visibility icons shifting off-screen 

jose-galvan:
## Summary

With this change lock and visibility icons are always kept are the right edge of the list. 

image

**Ticket:**  https://tsu.atlassian.net/browse/IR-9980 

## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/panels/hierarchy/hierarchynode.tsx
+++ b/packages/editor/src/panels/hierarchy/hierarchynode.tsx
@@ -433,7 +433,7 @@ export default React.memo(function HierarchyTreeNode(props: ListChildComponentPr
             'flex items-center justify-between gap-x-2 bg-inherit pr-2',
             rootEntity === entity ? 'p-2' : 'py-1 pr-2'
           )}
-          style={{ width: `calc(100% - ${node.depth * 0.75}rem)`, marginLeft: `${node.depth * 0.75}rem` }}
+          style={{ marginLeft: `${node.depth * 0.75}rem` }}
           ref={onDropTarget}
         >
           {node.isLeaf ? (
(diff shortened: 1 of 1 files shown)


mikeplascdev (approved):

kattsushi (approved):

packages/editor/src/panels/hierarchy/hierarchynode.tsx
```diff
@@ -430,10 +430,10 @@ export default React.memo(function HierarchyTreeNode(props: ListChildComponentPr
         />
         <div
           className={twMerge(
-            'flex w-full items-center justify-between gap-x-2 bg-inherit pr-2',
-            rootEntity === entity ? 'p-2' : 'py-1 pl-10 pr-2'
+            'flex items-center justify-between gap-x-2 bg-inherit pr-2',
+            rootEntity === entity ? 'p-2' : 'py-1 pr-2'
           )}
-          style={{ marginLeft: `${node.depth * 0.75}rem` }}
+          style={{ width: `calc(100% - ${node.depth * 0.75}rem)`, marginLeft: `${node.depth * 0.75}rem` }}
```
likethemammal: 🐑 We shouldnt have to use `calc()` and JS to align something to the right edge, even with cascading children. I'm sure this works well and it will be fine for now, but its a pretty brittle solution for something that CSS can handle smoothly without JS or assumptions, like the `0.75`, which will prob change in the future, on certain screen sizes, or when someone changes/adds another element

likethemammal (approved):
lgtm, 1 non-critical (🐑) comment

packages/editor/src/panels/hierarchy/hierarchynode.tsx
jose-galvan: I just realized I can remove this. I'll update my PR and merge it.

PR #1948, opened 2025-05-22
unblock IR-8127: fix hardcode value of max user per instance number by default

kattsushi:
## References
closes #_insert number here_


barankyle (changes requested):
Couple of things around this new setting:

1. There needs to be a migration to add the new setting to existing deployments. engine-setting.seed.ts only gets used on initial DB seeding. `20250501600000_add_ip_geolocation_settings.ts` should be a good example of adding engine-settings.
2. I believe the new setting needs to be of `type` 'public', as only users with the scope `settings:read` can get private settings, and that's something we only grant to admins.

hzraslan (approved):
Approved from my side, pending feedback needs to be resolved

kattsushi:
> Couple of things around this new setting:
> 
>     1. There needs to be a migration to add the new setting to existing deployments. engine-setting.seed.ts only gets used on initial DB seeding. `20250501600000_add_ip_geolocation_settings.ts` should be a good example of adding engine-settings.
> 
>     2. I believe the new setting needs to be of `type` 'public', as only users with the scope `settings:read` can get private settings, and that's something we only grant to admins.

Thanks Kyle nice catch already addressed! ty

likethemammal (approved):

packages/client-core/src/admin/components/settings/tabs/instanceServer.tsx
```diff
@@ -295,6 +296,27 @@ const InstanceServerTab = forwardRef(({ open }: { open: boolean }, ref: React.Mu
...
+            position: 'top'
+          }}
+          value={maxUsersPerInstanceValue}
+          onChange={(e) => {
+            const value = parseInt(e.target.value)
+            const settingInDb = engineSettings.data.find(
+              (el) => el.key === EngineSettings.InstanceServer.MaxUsersPerInstance
+            )
+            if (!settingInDb) return
+            engineSettingMutation.patch(settingInDb.id, {
```
jose-galvan: Why saving changes when the input value changes if there's already a button to save changes?

jose-galvan (changes requested):

barankyle:
I push a small change up for engine-setting.seed.ts, needed to run generateSeedData on it separately so it could be made public on dev-reinit. Other than Jose's comment about removing the auto-update, looks good.

barankyle (approved):

packages/client-core/src/admin/components/settings/tabs/instanceServer.tsx
kattsushi: yea you are right, fixed ty!

jose-galvan (approved):

PR #1959, opened 2025-05-23
fix: activate bounding box system and add on-demand bbox calculation …

noodleA1:
…for agent placement

- Remove isEditing check from only the ObjectGridSnapSystem to ensure it runs in world builder mode
- Process ALL models instead of just entering ones to handle async GLTF loading, in case assets are already in scene, or brought into the scene
- Add calculateBoundingBox function for on-demand bbox calculation when component is missing
- Add alignObjects function to properly align objects using face-to-face positioning
- Add height check to prevent objects from sinking below floor on horizontal alignment
- Add comprehensive documentation explaining why these workarounds are necessary

This addresses issues with:
1. GLTF models loading asynchronously causing missing bounding boxes
2. ObjectGridSnapSystem not running in world builder enabled mode
3. Some corrupted GLTF files failing to load properly
4. Race conditions between model loading and bbox calculation

## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/functions/EditorControlFunctions.ts
+++ b/packages/editor/src/functions/EditorControlFunctions.ts
@@ -23,7 +23,7 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
 Infinite Reality Engine. All Rights Reserved.
 */
 
-import { Euler, Matrix4, Quaternion, Vector3 } from 'three'
+import { Box3, Euler, Matrix4, Quaternion, Vector3 } from 'three'
 
 import {
   EntityTreeComponent,
@@ -31,6 +31,7 @@ import {
   findRootAncestors,
   getAncestorWithComponents,
   getChildrenWithComponents,
+  getOptionalComponent,
   iterateEntityNode,
   UUIDComponent
 } from '@ir-engine/ecs'
@@ -57,7 +58,9 @@ import { ComponentJsonType } from '@ir-engine/engine/src/scene/types/SceneTypes'
 import { getMutableState, getState, setNestedObject } from '@ir-engine/hyperflux'
 import { DirectionalLightComponent, HemisphereLightComponent } from '@ir-engine/spatial'
 import { TransformSpace } from '@ir-engine/spatial/src/common/constants/TransformConstants'
+import { MeshComponent } from '@ir-engine/spatial/src/renderer/components/MeshComponent'
 import { VisibleComponent } from '@ir-engine/spatial/src/renderer/components/VisibleComponent'
+import { BoundingBoxComponent } from '@ir-engine/spatial/src/transform/components/BoundingBoxComponent'
 
 import { GLTFComponent } from '@ir-engine/engine/src/gltf/GLTFComponent'
 import { serializeEntity } from '@ir-engine/engine/src/scene/functions/serializeWorld'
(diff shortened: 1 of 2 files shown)


HexaField:
Please include the ticket for this PR

HexaField (changes requested):
We have plans for a better bounding box implement. An 'on demand' system is not the right approach, we have a bounding box component that should be always up to date. What about it is insufficient currently?

noodleA1:
This is in regards to one ticket and a second that's being blocked: IR-9226 to allow the feedback layer to intelligently place assets in space relative to other assets, and IR-10252, which is just to enable the bounding box again. 

I don't think the current bounding box module is insufficient. I think my own issues with the engine necessitated the on demand calculation for when the bounding box doesn't exist. In case the bounding box did not load (which it wasn't on my build of the engine), this would allow the agent to continue to place assets, while only performing a calculation when needed. 

For me, the gltf assets were not loading completely (only about half of assets would load), and so there was trouble with the bounding boxes existing in the first place, causing failure.

HexaField:
> This is in regards to one ticket and a second that's being blocked: IR-9226 to allow the feedback layer to intelligently place assets in space relative to other assets, and IR-10252, which is just to enable the bounding box again.
> 
> I don't think the current bounding box module is insufficient. I think my own issues with the engine necessitated the on demand calculation for when the bounding box doesn't exist. In case the bounding box did not load (which it wasn't on my build of the engine), this would allow the agent to continue to place assets, while only performing a calculation when needed.
> 
> For me, the gltf assets were not loading completely (only about half of assets would load), and so there was trouble with the bounding boxes existing in the first place, causing failure.

My recommendation would be trying to set the BoundingBoxComponent inside the MeshComponent's onSet, which is the intended behaviour, and the bounding box will be up to date.

See https://tsu.atlassian.net/browse/IR-8616

dinomut1:
> > This is in regards to one ticket and a second that's being blocked: IR-9226 to allow the feedback layer to intelligently place assets in space relative to other assets, and IR-10252, which is just to enable the bounding box again.
> > I don't think the current bounding box module is insufficient. I think my own issues with the engine necessitated the on demand calculation for when the bounding box doesn't exist. In case the bounding box did not load (which it wasn't on my build of the engine), this would allow the agent to continue to place assets, while only performing a calculation when needed.
> > For me, the gltf assets were not loading completely (only about half of assets would load), and so there was trouble with the bounding boxes existing in the first place, causing failure.
> 
> My recommendation would be trying to set the BoundingBoxComponent inside the MeshComponent's onSet, which is the intended behaviour, and the bounding box will be up to date.
> 
> See https://tsu.atlassian.net/browse/IR-8616

We need bounding boxes for gltf assets as a whole, not just the meshes. We should add a system which makes reactors for GLTF components, and, when they finish loading, calculates a bounding box for it and its children.

HexaField:
> We need bounding boxes for gltf assets as a whole, not just the meshes. We should add a system which makes reactors for GLTF components, and, when they finish loading, calculates a bounding box for it and its children.

Yes I agree, I've clarified that in the ticket but didn't update my comment here.

packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -566,6 +571,80 @@ const addToSelection = (entities: EntityUUID[]) => {
   SelectionState.updateSelection(entities)
 }
 
+const alignObjects = (
```
HexaField, line 607: please add a jsdoc comment to explain what this function does

packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -566,6 +571,80 @@ const addToSelection = (entities: EntityUUID[]) => {
...
+  sourceEntity: Entity,
+  targetEntity: Entity,
+  srcDirection: Vector3,
+  dstDirection: Vector3
+): boolean => {
+  // Get bounding boxes from BoundingBoxComponent
```
HexaField: please remove unnecessary ai generated code comments

HexaField (changes requested):

packages/editor/src/functions/EditorControlFunctions.ts
noodleA1: just completed

packages/editor/src/functions/EditorControlFunctions.ts
noodleA1: wiped those out

packages/spatial/src/renderer/components/MeshComponent.ts
```diff
@@ -37,7 +38,13 @@ export const MeshComponent = defineComponent({
 
   onSet(entity, component, json) {
     setComponent(entity, ObjectComponent, json as Mesh)
+    setComponent(entity, BoundingBoxComponent, {})
```
HexaField: we changed what we wanted, please revert the changes to MeshComponent and BoundingBoxComponent

See https://github.com/ir-engine/ir-engine/pull/2018

HexaField (changes requested):

packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -566,6 +572,82 @@ const addToSelection = (entities: EntityUUID[]) => {
...
 }
 
+/** Aligns source entity to target entity using bounding box face directions */
+const alignObjects = (
+  sourceEntity: Entity,
+  targetEntity: Entity,
+  srcDirection: Vector3,
+  dstDirection: Vector3
+): boolean => {
+  let srcBoundingBox = getOptionalComponent(sourceEntity, BoundingBoxComponent)
```
HexaField: we don't want to set bounding boxes on just any entity, i'd suggest falling back to the MeshComponent.geometry.boundingBox if the BoundingBoxComponent does not exist. also depends if this needs to be the box for a whole hierarchy or just the entity in question. also, the BoundingBoxComponent is AABB, not OABB

HexaField (changes requested):

packages/spatial/src/transform/components/BoundingBoxComponent.ts
dinomut1, line 1: Should not need to change the bounding box calculations at all for this. Revert this file

packages/spatial/src/transform/components/BoundingBoxComponent.ts
```diff
@@ -53,15 +53,17 @@ import { TransformComponent } from './TransformComponent'
 
 export const BoundingBoxComponent = defineComponent({
   name: 'BoundingBoxComponent',
+  jsonID: 'EE_bounding_box',
 
   schema: S.Object({
     box: T.Box3(),
+    worldBox: T.Box3(),
```
dinomut1, line 60: don't store worldspace version of the data in the component. This can be calculated from the box + matrixWorld on the transform component

packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -566,6 +569,99 @@ const addToSelection = (entities: EntityUUID[]) => {
...
+ * @param entity - Entity to get bounding box for
+ * @returns Box3 in world space or null if no bounding box available
+ */
+const getEntityBoundingBox = (entity: Entity): Box3 | null => {
+  // Try BoundingBoxComponent first (preferred for GLTF entities)
+  const boundingBoxComponent = getOptionalComponent(entity, BoundingBoxComponent)
+  if (boundingBoxComponent && !boundingBoxComponent.worldBox.isEmpty()) {
+    console.log(`Entity ${entity}: Using BoundingBoxComponent.worldBox`)
+    return boundingBoxComponent.worldBox
+  }
```
dinomut1, line 583: Calculate the world box using world matrix of transform component

packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -566,6 +569,99 @@ const addToSelection = (entities: EntityUUID[]) => {
...
+    }
+
+    if (meshComponent.geometry.boundingBox) {
+      console.log(`Entity ${entity}: Using MeshComponent.geometry.boundingBox fallback`)
+      const transform = getComponent(entity, TransformComponent)
+      const worldBBox = new Box3().copy(meshComponent.geometry.boundingBox)
+      worldBBox.applyMatrix4(transform.matrixWorld)
+      return worldBBox
+    }
+  }
```
dinomut1, line 600: Don't use the mesh component bounding box. We should delete this

packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -566,6 +569,99 @@ const addToSelection = (entities: EntityUUID[]) => {
...
+      desiredWorldPos.y = dstWorldPos.y + dstHeight
+    }
+  }
+
+  positionObject([sourceEntity], [desiredWorldPos], TransformSpace.world)
+
+  EditorState.markModifiedScene(sourceEntity)
+
+  return true
+}
```
dinomut1, line 663: Keep only this function, delete everything else in this PR

dinomut1 (commented):
None of this is necessary except for the `placeAlignedObject` function, which should be implemented in the frontend instead of in the core engine

dinomut1:
This PR is unnecessary, as we should be using built in bounding box system, and implementing aligned placement in the builder frontend API rather than the core repo

PR #1963, opened 2025-05-26
IR-10033: Update file-browser update to allow copying of public/ folder

MoizAdnan:
https://tsu.atlassian.net/browse/IR-10033


packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -431,7 +440,7 @@ export class FileBrowserService
       const oldItemPath = path.join(oldPath, item.name)
       const newItemPath = path.join(newPath, item.name)
 
-      if (item.type === 'directory') {
+      if (item.type === 'directory' || item.type === 'folder') {
```
HexaField: what's the difference between a folder and directory? why do we need a distinction?

MoizAdnan: `storageProvider.listFolderContent` returns project subdirectories as `folder` instead of `directory` locally

hanzlamateen (approved):
@HexaField can you give it a 2nd look.

packages/server-core/src/media/file-browser/file-browser.class.ts
HexaField: are we able to standardize this? all storage provider interfaces should be strongly typed and isomorphic when possible

packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -351,9 +356,13 @@ export class FileBrowserService
       await storageProvider.moveObject(oldName, fileName, oldDirectory, newDirectory, data.isCopy)
     }
 
+    const key = data.isProject ? `${path.join(oldDirectory, oldName)}/` : path.join(oldDirectory, oldName)
```
HexaField: is `isProject` necessary to distinguish between these two cases? what currently is the behavior that hits this code path with isProject false, and would that be able to use this new condition instead?

If it is matching all files under a directory rather than just the directory itself, maybe we should rename it something like `recursive` but I'm not certain that's what the difference is.

MoizAdnan: `isProject` is being used in 2 cases, so I don't think renaming it to `recursive` would work:
1. Used to skip conditions that make sure the file path is limited to `assets` and `public` directory.
2. Used to add `/` at the end of key for 359. Because of the issue that you stated, `static-resource` service returns records for all files under the directory. When `isProject` is false, it should run like it currently runs on Dev and get all resources that are `like` key.

I can add another variable `recursive` that is true on the basis of `isProject` for better readability.

HexaField: > 1. Used to skip conditions that make sure the file path is limited to assets and public directory.

This doesn't sounds like the right level of abstraction to be doing that on, I don't think we should have folder specific logic at this level (beyond the validation that exists). Can't we just run the operation on both those sub-folders?


> 2. Used to add / at the end of key for 359. Because of the issue that you stated, static-resource service returns records for all files under the directory. When isProject is false, it should run like it currently runs on Dev and get all resources that are like key.

What does the system not already do that this enables? It should already be handling this with the `isDirectory` condition on L#343 on dev branch, unless I'm missing some edge condition, a project copy is just a folder copy, correct?

MoizAdnan: 1. Before 1.1 we didn't have the `assets` and `public` directory validation. We can run this operation on these sub-folders, but that would essentially mean we call `filebrowser.update` for each folder rather than once for the whole project, right?

2. On L#343 it only moves the folders in storageprovider, doesn't update the `key`. We could update the `key` in the `isDirectory` condition so we don't need `isProject`, but im not sure on why we aren't adding `/` in key for folders in the first place. Would it break the existing logic?

PR #1971, opened 2025-05-28
[IR-9892] Integrate resource loading into GLTF dependencies

MichaelEstes:
## References
closes #_insert number here_


HexaField:
Looks good, will wait for the resource caching PR to land before merging this

HexaField (changes requested):
Screenshot from 2025-06-03 17-36-28

sky station works fine, but the apartment does not load and throws lots of logs, likely due to missing assets, potentially even ones not necessarily part of the scene

HexaField (approved):

PR #1975, opened 2025-05-28
Monitoring otel

infinitelos:
## Summary
Add prometheus client endpoint

## References
closes #_insert number here_


packages/client-core/src/transports/p2p/PeerToPeerNetworkState.tsx
```diff
@@ -140,6 +141,30 @@ const ConnectionReactor = (props: { instanceID: InstanceID; topic: Topic }) => {
       API.instance.service(instanceSignalingPath).get({ instanceID })
     }, 5000)
 
+    // Track P2P metrics if the service is available (enabled via environment config)
+    const instance = getState(PeerToPeerNetworkState)[instanceID]
```
HexaField: can we do this on the backend instead

packages/server-core/src/monitoring/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import { Application } from '@feathersjs/koa'
+import configureMetrics from './metrics.service'
+import configureP2PMetrics from './p2p-metrics'
+import configureTracing from './tracing.service'
+
+/**
+ * Configure monitoring for the application
+ */
+export default (options = {}) => {
```
HexaField: this isn't how the rest of the codebase handles exporting a group of services, instead we just export an array of the service factory functions

packages/server-core/src/monitoring/metrics.service.ts
```diff
@@ -0,0 +1,652 @@
...
+import { Context, Next } from 'koa'
+import * as promClient from 'prom-client'
+
+import { buildStatusPath } from '@ir-engine/common/src/schemas/cluster/build-status.schema'
+import { logger } from '../ServerLogger'
+
+/**
+ * MetricsService class for handling Prometheus metrics
+ */
+export class MetricsService {
```
HexaField, line 36: shouldn't this data go into the database such that we don't store data in memory? How does this work across multiple API servers?

packages/server-core/src/monitoring/tracing.service.ts
```diff
@@ -0,0 +1,244 @@
...
+import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions'
+import { Context, Next } from 'koa'
+
+import { logger } from '../ServerLogger'
+import { default as config } from '../appconfig'
+
+/**
+ * TracingService class for handling OpenTelemetry tracing
+ */
+export class TracingService {
```
HexaField: we prefer functions over classes, please make this a functional implementation if possible

HexaField (changes requested):

PR #1976, opened 2025-05-28
fix(particle-system): enum to record [IR-9745]

mikeplascdev:
## Summary
- Map enum to record fro the `S.Enum`

## References
closes #_insert number here_


mikeplascdev:
In this [PR 1876](https://github.com/ir-engine/ir-engine/pull/1876), the change was made, but the type is still `Record<string, Values>` 
https://github.com/ir-engine/ir-engine/blob/dcf03f597366770d5bcf1cdd821f5eb87f791dd5/packages/ecs/src/schemas/JSONSchemas.ts#L109
so the particle system should not work.

With this change, it is now fixed. I just wanted to confirm with @HexaField, since there's a [comment in the PR](https://github.com/ir-engine/ir-engine/pull/1876#pullrequestreview-2852501608) with evidence showing that the particle system does work and I want to validate if that evidence was before the change or after because it shouldn't work.

HexaField (commented):
There is a better solution in this PR
https://github.com/ir-engine/ir-engine/pull/1854


`renderMode: S.LiteralUnion(Object.values(RenderMode), {`

DanielBelmes (approved):

PR #1977, opened 2025-05-28
IR-10179: Update EditorContainer logic for new wizard input prompt sc…

BogieBogard:
## Summary
WIP

## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/components/EditorContainer.tsx
+++ b/packages/editor/src/components/EditorContainer.tsx
@@ -6,8 +6,8 @@ Version 1.0. (the "License"); you may not use this file except in compliance
 with the License. You may obtain a copy of the License at
 https://github.com/ir-engine/ir-engine/blob/dev/LICENSE.
 The License is based on the Mozilla Public License Version 1.1, but Sections 14
-and 15 have been added to cover use of software over a computer network and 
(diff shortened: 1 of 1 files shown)


packages/editor/src/components/EditorContainer.tsx
```diff
@@ -157,7 +157,13 @@ const EditorContainer = () => {
   const editorUIAddon = useMutableState(UIAddonsState).editor
   const currentLoadedSceneURL = useHookstate(null as string | null)
 
-  useEngineCanvas(canvasRef.value as React.RefObject<HTMLElement> | null)
+  // Check if wizard is enabled from URL params
```
HexaField, line 160: we really do not want to have to reference the wizard in the core engine if possible. why is this necessary?

HexaField (changes requested):

packages/editor/src/components/EditorContainer.tsx
BogieBogard: Yes, that makes sense. I should be able to refactor to keep the logic changes in wizard. I added this to fix a loading state stuck issue that occurred in the new wip wizard flow for ticket 10179.

PR #1980, opened 2025-05-28
Flip Glassmorphism feature flag

likethemammal:
## Summary

Flipping the glassmorphism feature flag so the new UI isnt visible by default

### DIFF
--- a/packages/client-core/src/world/Location.tsx
+++ b/packages/client-core/src/world/Location.tsx
@@ -60,7 +60,7 @@ const LocationPage = ({ online }: Props) => {
   const params = useParams()
   const ready = useMutableState(LoadingUISystemState).ready
 
-  const [glassEnabled] = useFeatureFlags([FeatureFlags.Client.Glass])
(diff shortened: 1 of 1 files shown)


HexaField (changes requested):
can we change the naming of the feature flag to reflect the expected behaviour?

likethemammal:
> can we change the naming of the feature flag to reflect the expected behaviour?

I updated the PR with a clearer variable, but this change isnt going to make sense and reflect the expected behavior because we're doing it to hide the feature from the other devs who have feature flags on by default.

HexaField (approved):

HexaField:
> > can we change the naming of the feature flag to reflect the expected behaviour?
> 
> I updated the PR with a clearer variable, but this change isnt going to make sense and reflect the expected behavior because we're doing it to hide the feature from the other devs who have feature flags on by default.

All feature flags are positive essentially, so if you need a feature flag to be off by default, it needs to be state as a 'disable' flag. This is expected. I understand this may not be the best, and we probably should improve it to be clearer.

PR #1982, opened 2025-05-29
adds a check for a RigidBodyComponenet on entity Ancestor

quaiquai-ir:
## Summary
Adds a RigidBodyComponent check on entity ancestors before checking for colliders to avoid detached colliders from never loading. If there's no rigidbody ancestor, consider the collider as loaded to prevent scenes with detached colliders from never loading. Also checks if entity is part of the simulation layer before checking on colliders.

### DIFF
--- a/packages/engine/src/gltf/GLTFComponent.tsx
+++ b/packages/engine/src/gltf/GLTFComponent.tsx
@@ -162,10 +162,7 @@ const componentDependenciesLoaded = (dependencies?: ComponentDependencies) => {
 }
 
 const checkCollider = (hasCollider: boolean, entity: Entity) => {
-  if (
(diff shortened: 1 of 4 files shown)


packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -267,10 +279,9 @@ export const GLTFComponentReactor = () => {
   const scene = useOptionalComponent(entity, SceneComponent)
```
HexaField: `  const scene = useOptionalComponent(entity, SceneComponent)` is also unused now right? we can remove that line too

kattsushi (approved):
LGTM

PR #1986, opened 2025-05-29
IR-9699b: add more feature flag functionalities to publish menu

svyoung:
## Summary
As a creator, I want to configure which UI elements are visible in Viewer so that I can customize the experience for my users. The publish modal is critical to settings for the users' saved scenes. 

 We will not support VR or jump controls, and thus we should hide these behind a feature flag, for now. 

## References
[IR-9599](https://tsu.atlassian.net/browse/IR-9599)

## QA Steps
image
image

[IR-9599]: https://tsu.atlassian.net/browse/IR-9599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/client-core/i18n/en/editor.json
```diff
@@ -257,7 +257,11 @@
       "publishSuccess": "Space Published Successfully",
       "noThumbnailPlaceholder": "Publish Scene to Generate a Preview Image",
       "createCompressedScenePublish": "Compress and Publish",
-      "createCompressedScenePublishInfo": "This is a beta feature that optimizes your space for mobile devices. If you encounter errors, please refresh your browser and try again, or use the regular “Publish” function for now."
+      "createCompressedScenePublishInfo": "This is a beta feature that optimizes your space for mobile devices. If you encounter errors, please refresh your browser and try again, or use the regular “Publish” function for now.",
+       "jumpControlFeature": "Jump Control",
+      "jumpControlFeatureDesc": "Enable your avatar’s ability to jump.",
+      "vrCapabilitiesFeature": "VR Capabilities",
+      "vrCapabilitiesFeatureDesc": "Enable your eligible devices to render AR experiences."
```
SYBIOTE, line 264: I believe we serve both VR and AR features, so please include VR here as well

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
```diff
@@ -120,14 +120,6 @@ export default function AddEditLocationModal(props: AddEditLocationModalProps) {
 
   const locationMutation = useMutation(locationPath)
 
-  const instanceEngineSettings = useFind(engineSettingPath, {
```
SYBIOTE: why do we remove this? Wouldn't we want the instance server to defined the max number of users allowed in the scene?

likethemammal (approved):

Tortillaguy (approved):

Tortillaguy:
Seems like it's missing the `jumpControlEnabled` column for that DB table? 
You'll need a migration if that's the case..

packages/client-core/i18n/en/editor.json
svyoung: VR is included

HexaField (dismissed):
Usually, UI comes last after the functionality is implemented, otherwise we end up with buttons that do nothing.

packages/client-core/src/user/components/LocationIconButton.tsx
```diff
@@ -46,20 +47,49 @@ function LocationIconButton({
 }: LocationIconButtonProps) {
   const Button = () => {
     const { ref, className, ...restIconProps } = iconProps || {}
+    /** @todo arbitrary flag for now */
+    const { glassMorphismDisabled } = props
```
likethemammal: Since this work isnt related to the publishing menu, can you remove these changes to LocationIconButton from this PR? I'd prefer to keep the glassmorphism logic separate so we dont have to fight these styles.

DanielBelmes (approved):

PR #1995, opened 2025-05-30
Prevent infinite re-rendering of EditorContainer

MichaelEstes:
## References
closes #_insert number here_


DanielBelmes (approved):
Thank you so much! I think the original ticket I had assigned to me for this has been repurposed for the API slowness under Matthew.

PR #2004, opened 2025-06-02
IR-10320 Handle errors during avatar unloading

koktavy:
## Summary
Added safety checks for entities related to avatars.
Added safety checks for elements related to peer audio/video.

## Subtasks Checklist
N/A

## Breaking Changes
N/A

## References
closes [IR-10320](https://tsu.atlassian.net/browse/IR-10320)

## QA Steps
Open https://localhost:3000/location/sky-station
Open the console and clear logs
Click anywhere in the canvas to ensure media playback is active
Open the same location in an incognito window
Click both windows and move both avatars
Close the incognito window
Again, click anywhere in the canvas
Observe clean console logs

[IR-10320]: https://tsu.atlassian.net/browse/IR-10320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/spatial/src/common/debug/useHelperEntity.ts
```diff
@@ -61,8 +62,10 @@ export function useHelperEntity<TObject extends DisposableObject3D>(
   const transform = useOptionalComponent(helperEntityState.value, TransformComponent)
 
   useEffect(() => {
-    setComponent(parentEntity, ActiveHelperComponent)
-  }, [])
+    if (parentEntity && entityExists(parentEntity)) {
+      setComponent(parentEntity, ActiveHelperComponent)
```
MichaelEstes: Typically anywhere that we set a component in a useEffect with dependencies we want to remove the component in a cleanup function, so if the dependencies change we don't end up with dangling components. 
`useEffect(() => {
    if (parentEntity && entityExists(parentEntity)) {
      setComponent(parentEntity, ActiveHelperComponent)
      return () => {
        if (entityExists(parentEntity)) removeComponent(parentEntity, ActiveHelperComponent)
      }
    }
  }, [parentEntity])`

packages/client-core/src/user/VideoWindows/hook.tsx
```diff
@@ -109,15 +109,17 @@ export const useUserMediaWindowHook = ({ peerID, type }: Props) => {
...
+      if (currentVideoElement?.srcObject) currentVideoElement.play()
+      if (currentAudioElement?.srcObject) currentAudioElement.play()
       harkListener?.value?.resume()
     }
     window.addEventListener('pointerup', onUserInteraction)
     return () => {
       window.removeEventListener('pointerup', onUserInteraction)
     }
-  }, [videoElement, audioElement, harkListener?.value])
+  }, [peerID]) // Use peerID to gracefully handle players leaving
```
HexaField: this change doesn't make any sense

packages/engine/src/avatar/systems/AvatarTransparencySystem.tsx
```diff
@@ -100,6 +101,8 @@ export const AvatarTransparencySystem = defineSystem({
 
 const AvatarReactor = (props: { entity: Entity }) => {
   const entity = props.entity
+  if (!entity || !entityExists(entity)) return null
```
HexaField: this should be impossible, subreactors to a query will always have non-zero entity passed in

HexaField (changes requested):

packages/client-core/src/user/VideoWindows/hook.tsx
koktavy: Dependency array change reverted, the errors from play() are now handled by checking that the video/audio srcObject exists.

packages/engine/src/avatar/systems/AvatarTransparencySystem.tsx
koktavy: The error was actually coming from the line just below this:
`const sourceID = UUIDComponent.getAsSourceID(entity)`

So instead I now check if the avatar entity has a UUIDComponent and return null if it does not:
`if (!hasComponent(entity, UUIDComponent)) return null`

packages/spatial/src/common/debug/useHelperEntity.ts
koktavy: Thanks for the guidance. Seems the error this change fixed is now resolved by other dev updates that were merged in.

packages/engine/src/avatar/systems/AvatarTransparencySystem.tsx
```diff
@@ -100,6 +101,8 @@ export const AvatarTransparencySystem = defineSystem({
 
 const AvatarReactor = (props: { entity: Entity }) => {
   const entity = props.entity
+  if (!hasComponent(entity, UUIDComponent)) return null // Early exit if the UUID on this avatar has been removed, which can happen when another player leaves the scene
```
HexaField: a better fix would be to just have a `useComponent(entity, UUIDComponent)` which will suspend the context if the component doesn't exist

HexaField (changes requested):

HexaField:
Closing as these changes don't make sense. Will re-evaluate the ticket.

HexaField (approved):

PR #2005, opened 2025-06-02
add check for null parent in engine canvas reactor

dinomut1:
Fixes race condition crash where parent.appendChild throws an error when parent is null

### DIFF
--- a/packages/spatial/src/renderer/functions/useEngineCanvas.ts
+++ b/packages/spatial/src/renderer/functions/useEngineCanvas.ts
@@ -34,7 +34,7 @@ export const useEngineCanvas = (ref: React.RefObject<HTMLElement> | null) => {
   useEffect(() => {
     if (!ref) return
     const parent = ref.current as HTMLElement
-
+    if (!parent) return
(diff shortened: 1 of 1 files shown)


the-reality-engineer:
Refs aren't reactive, so if we need that effect to run when the element becomes available, we need to store and pass it in a reactive state.

junhuac (approved):

dinomut1:
> Refs aren't reactive, so if we need that effect to run when the element becomes available, we need to store and pass it in a reactive state.

Any reason we shouldn't just refactor useEngineCanvas to use a hookstate HTMLCanvasElement as an input instead of a ref?

HexaField:
> Any reason we shouldn't just refactor useEngineCanvas to use a hookstate HTMLCanvasElement as an input instead of a ref?

No reason, honestly we don't even need that since we aren't mutating within the hook, so just passing the target parent HTMLElement we want should be simpler, and then outside the hook we can use a ref or a useHookstate or whatever we want,

Rezmason (approved):
LGTM

dinomut1:
This unblocks us, I'm merging it in as is for now

PR #2007, opened 2025-06-03
IR-10191 mt stg studio flips green channel on normal maps

quaiquai-ir:
## References
https://tsu.atlassian.net/browse/IR-10191

## QA Steps
Add normal mapped gltf attached in ticket to a new scene.
Observe that the normal maps on the small quad display correctly


packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -748,7 +742,7 @@ const loadMaterial = async (options: GLTFParserOptions, materialIndex: number) =
     const scale = materialDef.normalTexture.scale
     materialConstructorParameters.normalScale = new Vector2(scale, scale)
   } else {
-    materialConstructorParameters.normalScale = new Vector2(1, 1)
+    materialConstructorParameters.normalScale = new Vector2(1, -1)
```
HexaField: I think we still need to use the condition from the previous code

packages/engine/src/gltf/GLTFLoaderFunctions.ts
```diff
@@ -157,18 +157,12 @@ export function getImageURIMimeType(uri) {
 
 const assignFinalMaterial = (primitiveDef: GLTF.IMeshPrimitive, entity: Entity) => {
   const material = getComponent(entity, MaterialStateComponent).material as MeshPhysicalMaterial
-  const useDerivativeTangents = primitiveDef.attributes.TANGENT === undefined
   const useVertexColors = primitiveDef.attributes.COLOR_0 !== undefined
   const useFlatShading = primitiveDef.attributes.NORMAL === undefined
 
   if (useVertexColors) material.vertexColors = true
```
HexaField: Let's also move the other two conditions here, as they likely also are problematic if they stay here

HexaField (changes requested):

packages/engine/src/gltf/GLTFLoaderFunctions.ts
quaiquai-ir: yeah, was just marking this as a draft and was going to move that over in the morning along with the tests:)

quaiquai-ir:
@HexaField 
We have a problem with this code: https://github.com/ir-engine/ir-engine/pull/2007/files#diff-63e41036118b337f383eb2a3d6e9fbeac555b58fc4915809114adb0e68f9906bR157-R179

We need to switch from using computeTangents() to MikkTSpace because computeTangents() breaks how our normal maps are imported—especially with GLTF files that use OpenGL-style tangent space. When we delete tangents for merging and use computeTangents(), the resulting tangents don't include the correct handedness (w component), which leads to flipped or incorrect lighting when using a normalScale of (1, 1) (which we choose for included tangents). ThreeJS even states that if we are going to calculate the Tangents we must flip the green channel. Currently with or without tangents we are computing the tangents. This mismatch causes normal maps that should work out of the box to render incorrectly appearing inverted. MikkTSpace solves this by generating tangents that match the standard expected by GLTF exporters and is recommended by Threejs to use. The current quick bandaid fix would be to always have the normal scale flipped for all gltfs.

Another way to get around this is to not delete tangents and comment out this code: https://github.com/ir-engine/ir-engine/blob/dev/packages/spatial/src/common/classes/BufferGeometryUtils.ts#L104

I tested that on many scenes and didnt find a problem but may be a super niche edge case where it would fail. Currently this problem is only with gltfs that have tangents

**### Incorrect (Current implementation)**
incorrect

**### Correct**
correct

HexaField (approved):

PR #2008, opened 2025-06-03
Account Page

Tortillaguy:
## Summary
Support magic link and Oauth support, username modifications, identity provider modifications, and deleting the user's own account.

https://www.loom.com/share/80aece376abe47ddb772a8ec8e68e190

## References
closes [IR-10552]

## QA Steps

[IR-10552]: https://tsu.atlassian.net/browse/IR-10552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/client-core/src/components/Glass/ToolbarAndSidebar.tsx
```diff
@@ -195,6 +195,8 @@ const contentStyles = `
   
   lg:px-14
   lg:pt-12
+
+  w-full
```
likethemammal: The `absolute` and `inset-0` should be making this element full height and width of the parent already. Also I think `w-full` wouldnt work as expected cus its `absolute`

packages/client-core/src/components/Glass/ToolbarAndSidebar.tsx
```diff
@@ -161,18 +161,15 @@ const Header = ({ tabs = [], heading, handleSidebarClose, handleSidebarBack, has
...
-        <div className={twMerge(buttonContainer_base, closeButtonStyles)}>
-          <MenuButton className={`text-3xl`} onClick={handleSidebarClose}>
-            <XCloseLg />
-          </MenuButton>
-        </div>
         <div style={{ textShadow: `0 0.025em 0.08em hsla(0, 0%, 0%, 0.2)` }} className={headingsStyles}>
-          {hasHistory ? <div className={headerBackButtonStyles}>{backButton}</div> : <></>}
-          <h2 className={twMerge(`lg:block`, tabs.length ? `hidden` : ``)}>{heading}</h2>
+          <div className={twMerge(headerBackButtonStyles, hasHistory ? '' : 'invisible')}>{backButton}</div>
+          <h2 className={twMerge(`mx-auto font-dm-sans lg:block`, tabs.length ? `hidden` : ``)}>{heading}</h2>
```
likethemammal: The `mx-auto` breaks the design on desktop. The mobile version of the heading that has smaller centered text should already be visible on smaller screen sizes

packages/client-core/src/components/Glass/ToolbarAndSidebar.tsx
```diff
@@ -161,18 +161,15 @@ const Header = ({ tabs = [], heading, handleSidebarClose, handleSidebarBack, has
...
-        {hasHistory ? <div className={twMerge(buttonContainer_base, backButtonStyles)}>{backButton}</div> : <></>}
-        <div className={twMerge(buttonContainer_base, closeButtonStyles)}>
-          <MenuButton className={`text-3xl`} onClick={handleSidebarClose}>
-            <XCloseLg />
-          </MenuButton>
-        </div>
         <div style={{ textShadow: `0 0.025em 0.08em hsla(0, 0%, 0%, 0.2)` }} className={headingsStyles}>
-          {hasHistory ? <div className={headerBackButtonStyles}>{backButton}</div> : <></>}
-          <h2 className={twMerge(`lg:block`, tabs.length ? `hidden` : ``)}>{heading}</h2>
+          <div className={twMerge(headerBackButtonStyles, hasHistory ? '' : 'invisible')}>{backButton}</div>
```
likethemammal: We want this button to not change the styles when it shouldnt exist to match the design. This pushes the `h2` to the right, even when the back button shouldnt exist

packages/client-core/src/components/Glass/ToolbarAndSidebar.tsx
```diff
@@ -161,18 +161,15 @@ const Header = ({ tabs = [], heading, handleSidebarClose, handleSidebarBack, has
...
+          <h2 className={twMerge(`mx-auto font-dm-sans lg:block`, tabs.length ? `hidden` : ``)}>{heading}</h2>
           {tabs.map((tabProps) => {
             return <Tab {...tabProps} />
           })}
+          <MenuButton className={`text-3xl`} onClick={handleSidebarClose}>
```
likethemammal: I'm confused why this needed to get moved

packages/client-core/src/components/Settings/ButtonGroup.tsx
```diff
@@ -0,0 +1,86 @@
+/*
```
likethemammal, line 1: 🐑 This is another reason I wasnt making these Button components yet. What kinds of buttons does this button group consider? Cus theres the ToolbarMenu buttons, ur wider text-based GlassButtons, and a few other buttons floating around the design. I'm guessing this ButtonGroup was only created for one case but since its generic sounding it will be confusing that it doesnt apply to all cases

packages/client-core/src/components/Settings/SSOScreen.tsx
```diff
@@ -65,43 +56,36 @@ const Socials = [
...
   const disableProvider = (client: string) => {
-    AuthService.removeUserOAuth(client)
+    deleteSSO.set(client)
   }
 
-  const connectedProviders = Socials.filter((p) => oauthConnectedState[p.client].value && authState.value[p.client])
-  const disconnectedProviders = Socials.filter((p) => !oauthConnectedState[p.client].value && authState.value[p.client])
+  if (deleteSSO.value) {
+    return (
+      <div className="mx-auto flex h-full w-3/4 flex-col items-center justify-between pb-2">
```
likethemammal: 🐑 We generally shouldnt be using arbitrary values for things. `width: 75%` is def going to break in a number of cases and isnt a sustainable way to style a button, container, etc. In general, using `w-full` with padding in the container or using a min-width is the most flexible and responsive approach

packages/client-core/src/components/Settings/Section.tsx
```diff
@@ -28,17 +28,20 @@ import React from 'react'
...
   <div
-    className={`overflow-hidden rounded-xl shadow-sm ${className} `}
+    className={`overflow-hidden rounded-xl  font-dm-sans text-xs shadow-sm md:text-base ${
+      disabled ? 'pointer-events-none opacity-50' : ''
+    } ${className} `}
     style={{
       background: 'linear-gradient(145deg, rgba(255,255,255,0.1), rgba(255,255,255,0.05))',
-      border: '1px solid rgba(255, 255, 255, 0.05)'
+      border: '1px solid rgba(255, 255, 255, 0.05)',
+      letterSpacing: '0.025em'
```
likethemammal: Can you use tailwind and  `tracking-[0.025em]` for this

packages/client-core/src/components/Settings/ToggleItem.tsx
```diff
@@ -26,16 +26,16 @@ Infinite Reality Engine. All Rights Reserved.
...
   checked?: boolean
   onClick?: () => void
 }
 
-const ToggleItem: React.FC<ToggleItemProps> = ({ label, checked = false, onClick }) => {
+const ToggleItem: React.FC<ToggleItemProps> = ({ label, checked = false, onClick, children }) => {
   return (
     <div className="flex items-center justify-between px-4 py-3.5 text-white/90">
-      <span className="font-medium">{label}</span>
+      {children ? children : <span className="font-medium">{label}</span>}
```
likethemammal: 🐑 This could be `children || <......`

packages/client-core/src/components/Settings/UsernameScreen.tsx
```diff
@@ -0,0 +1,94 @@
...
+  navigateTo: (screenKey: string, historyKey) => void
+}
+const logger = multiLogger.child({ component: 'engine:ecs:DisplayName', modifier: clientContextParams })
+
+const UsernameScreen: React.FC<DisplayNameScreenProps> = () => {
+  const { name, id } = useMutableState(AuthState).user
+  const displayName = useHookstate(name.value as string)
+  const saved = useHookstate(false)
+
+  const isUsernameValid = useMemo(() => {
```
likethemammal: 🐑 Idk where we landed on the discussion, but its my understanding that `Username` in all these cases is actually closer to `Display name`. I think its worth being clear about that in the code too cus I would assume Username means critical for auth while Display is just a piece of text being passed around for ease of account creation

packages/client/index.html
```diff
@@ -11,48 +11,54 @@
...
+    />
 
     <link rel="stylesheet" type="text/css" href="/src/themes/base.css" />
     <link rel="stylesheet" type="text/css" href="/src/themes/components.css" />
     <link rel="stylesheet" type="text/css" href="/src/themes/utilities.css" />
 
     <link rel="apple-touch-icon" sizes="180x180" href="<%- appleTouchIcon %>" />
-    <link rel="manifest" href="<%- siteManifest %>">
-    <link rel="mask-icon" href="<%- safariPinnedTab %>" color="#5bbad5">
+    <link rel="manifest" href="<%- siteManifest %>" />
```
likethemammal, line 39: Good catch

packages/client/index.html
```diff
@@ -61,8 +67,16 @@
       const w = 400
       const h = 400
 
-      let width = window.innerWidth ? window.innerWidth : document.documentElement.clientWidth ? document.documentElement.clientWidth : screen.width;
-      let height = window.innerHeight ? window.innerHeight : document.documentElement.clientHeight ? document.documentElement.clientHeight : screen.height;
+      let width = window.innerWidth
```
likethemammal: Is there a reason we're using `let` for these? Ik its the old code but it looks like these could be `const`s

packages/client/index.html
```diff
@@ -76,22 +90,66 @@
...
+      src="<%- rootCookieAccessor %>"
+      sandbox="allow-storage-access-by-user-activation allow-scripts allow-same-origin"
+    ></iframe>
     <script type="module" src="/src/main.tsx"></script>
-    <canvas hidden id='engine-renderer-canvas' tabindex="1" style='outline: none; z-index: 0; width: 100%; height: 100%; position: fixed; -webkit-user-select: none; pointer-events: auto; user-select: none'></canvas>
+    <canvas
+      hidden
+      id="engine-renderer-canvas"
+      tabindex="1"
+      style="
```
likethemammal, line 143: 🐑 We should be able to use Tailwind classes for this but that might be complicated since its so early in the render

packages/ui/src/components/viewer/Button/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+Infinite Reality Engine. All Rights Reserved.
+*/
+
+import React from 'react'
+import { twMerge } from 'tailwind-merge'
+
+export const GlassButton = (props: React.ButtonHTMLAttributes<HTMLButtonElement>) => {
+  const { className, disabled } = props
+  const style = `
+  flex items-center justify-center  w-[80%] rounded-full border border-white/20 bg-white/15 px-6 py-4 text-lg font-bold text-white/90 shadow-lg drop-shadow-xl backdrop-blur-sm
```
likethemammal: 🐑 I mentioned the arbitrary percent earlier. The styles prob hold up in most cases since its just a button. Ideally this gets changed to be more sustainable, but I figure we'll also change it when we consolidate the button styles

packages/ui/src/components/viewer/Divider/index.tsx
```diff
@@ -35,7 +35,7 @@ export interface DividerProps {
  * Uses a centered, rounded horizontal line with customizable width and opacity
  */
 const Divider: React.FC<DividerProps> = ({ className = '' }) => (
-  <div className={`mx-auto h-px rounded-full ${className} w-[80%] bg-white/10 ${className}`} />
+  <div className={`mx-auto h-px w-[80%] rounded-full bg-white/10 ${className}`} />
```
likethemammal, line 38: 🐑 The arbitrary percent applies here too but I dont think this needs to change. Obviously its just a divider so it doesnt matter too much, but it would be more flexible and predictable using `w-full` and then giving the parent horizontal padding to push the divider size in a bit. When I make dividers they're generally 2 divs (parent and child) so you can create this effect without expecting the parent of the divider to handle the padding

packages/ui/src/components/viewer/InputField/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * A reusable input field component with reset functionality
+ */
+const InputField = React.forwardRef<HTMLInputElement, InputFieldProps>(
+  ({ label, value, isPassword = false, onChange, isDirty = false, onReset, className = '', ...props }, ref) => (
+    <div className={`flex max-w-[25ch] flex-1 items-center gap-2 ${className}`}>
+      <input
+        ref={ref}
+        type={isPassword ? 'password' : 'text'}
```
likethemammal: 🐑 I'm guessing for toggling visibility (show and hide password) on password fields we'll also want this logic to switch it to a `text` field. The `isPassword` variable in that case doesnt make sense and fights this logic

tailwind.config.js
```diff
@@ -162,7 +162,8 @@ module.exports = {
         sans: ['Inter', 'sans-serif'],
         figtree: ['Figtree', 'sans-serif'],
         campton: ['Campton', 'sans-serif'],
-        'dm-sans': ['"DM Sans"', 'sans-serif']
+        'dm-sans': ['"DM Sans"', 'sans-serif'],
+        manrope: ['Manrope', 'sans-serif']
```
likethemammal, line 166: Thanks for adding the fonts

likethemammal (changes requested):
Left some comments. 🐑 = I'm saying something redundant, not critical, or not required

packages/client-core/src/components/Glass/ToolbarAndSidebar.tsx
Tortillaguy: I didn't see the latest Figma design with the sidebar open in deskop

packages/client-core/src/components/Glass/ToolbarAndSidebar.tsx
Tortillaguy: See my other response on this file

packages/client-core/src/components/Settings/ButtonGroup.tsx
Tortillaguy: I re-used it in two places/ confirmation prompts, so I exported it to a component. I'm not picky about the names right now

packages/client-core/src/components/Settings/UsernameScreen.tsx
Tortillaguy: Updated the naming

packages/client/index.html
Tortillaguy: I think this needs to be revisited by the core team. There was a recent editor bug where the canvas gets deleted when maximizing the viewport

packages/ui/src/components/viewer/Divider/index.tsx
Tortillaguy: I'm going to keep it this way for now. This component isn't reusable anywhere outside the Section component for the UI, and saves me from adding w-[80%] strings per <Divider/> component

packages/ui/src/components/viewer/InputField/index.tsx
Tortillaguy: The input props are spread over with {...props} but I changed the line to be more explicit now.

packages/client/index.html
Tortillaguy: I think this needs to be revisited by the core team. There was a recent editor bug where the canvas gets deleted when maximizing the viewport

likethemammal (approved):

PR #2016, opened 2025-06-03
Error accessing published scene

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-10564


packages/client-core/src/systems/MediaControlSystem.tsx
```diff
@@ -78,6 +78,7 @@ const onUpdate = (entity: Entity) => {
   const mediaComponent = getMutableComponent(entity, MediaComponent)
   if (!mediaComponent.controls.value) return
   const xrui = getComponent(mediaComponent.xruiEntity.value, XRUIComponent)
```
DanielBelmes: if xrui is not guaranteed defined we should change getComponent to getOptionalComponent so the typing indicates it can be undefined.

DanielBelmes (changes requested):

DanielBelmes (approved):

PR #2017, opened 2025-06-03
[IR-9602] Attach Viewer chat styles to state

likethemammal:
## Summary

image

## QA Steps

- In the viewer, open the chat menu
- You should be able to click the input field and type text
- Pressing the button or hitting Enter should send the message
  - Send button should not be visible if no message text is typed

### DIFF
--- a/packages/client-core/src/components/Glass/ChatMenu.tsx
+++ b/packages/client-core/src/components/Glass/ChatMenu.tsx
@@ -23,13 +23,16 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
 Infinite Reality Engine. All Rights Reserved.
 */
 
-import { useMutableState } from '@ir-engine/hyperflux'
 import React from 'react'
 import { HiChatBubbleLeftRight } from 'react-icons/hi2'
 import { twMerge } from 'tailwind-merge'
 import { ModalState } from '../../common/services/ModalState'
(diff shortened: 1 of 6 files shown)


packages/client-core/src/components/Glass/ChatMenu.tsx
```diff
@@ -47,27 +50,33 @@ const messageBaseStyles = `
...
+  bg-gradient-to-r
+  from-blue-500
+  from-40%
+
+  via-blue-400
+  via-90%
+
+  to-blue-400
+  to-100%
+`
```
Tortillaguy, line 63: I don't mind this here but if/when we do re-use common styles.. should they exist in the utilties.css or somewhere else?

packages/client-core/src/components/Glass/ChatMenu.tsx
```diff
@@ -84,12 +93,59 @@ const OtherChat = ({ children }) => (
...
+  bottom-0
+  left-0
+  right-0
+  bg-black/30
+  pb-4 px-4 pt-4
+`
+
 export const ChatMenu = () => {
-  const { type } = useMutableState(AuthState).authUser.identityProvider.get()
+  const user = useMutableState(AuthState).user
```
Tortillaguy, line 140: this feels way cleaner then useHookState(getMutableState())

Tortillaguy (approved):

PR #2018, opened 2025-06-03
Ir 8616 revive bounding box

danskate:
## Summary
This change adds a BoundingBoxComponent to any GLTFComponent and fixes the BoundingBox calculation. 

## References
closes #_insert number here_

## QA Steps

Before Fix:
https://github.com/user-attachments/assets/eb0cd4da-ce7f-40cc-90f2-1019bef407d3

After Fix:
https://github.com/user-attachments/assets/ab697f55-c2de-409d-8455-f17ede082ae8


packages/spatial/src/transform/components/BoundingBoxComponent.ts
```diff
@@ -119,12 +119,23 @@ export const updateBoundingBox = (entity: Entity) => {
     return
   }
 
+  // Ensure all children transforms are computed before calculating bounding box
```
HexaField: just need to clean up unnecessary comments

HexaField (changes requested):

danskate:
I pushed new changes that introduced a new worldBox variable to the BoundingBoxComponent. Some uses require local or world space bounding boxes so I made .box local and added worldBox to be the world space box.

HexaField (changes requested):
we should probably split these into AABBComponent and OABBComponent for axis aligned and object aligned respectively, though i'm willing to keep it as is. it can become a performance concern in larger scenes, especially if updated dynamically.

HexaField:
what use case do we have currently for OABBs? my expectation previously was AABB

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -305,6 +307,11 @@ export const GLTFComponentReactor = () => {
     }
   }, [gltfComponent.document])
 
+  useEffect(() => {
+    if (!sceneLoaded) return
+    setComponent(entity, BoundingBoxComponent)
```
HexaField, line 312: we probably only want the bounding box on the simulation layer, so we should wrap this in

```ts
const simulationEntity = getSimulationCounterpart(entity)
setComponent(simulationEntity, BoundingBoxComponent)
```

packages/spatial/src/transform/components/BoundingBoxComponent.ts
```diff
@@ -131,7 +134,11 @@ export const updateBoundingBox = (entity: Entity) => {
 
   const helperObject = getComponent(helperEntity, ObjectComponent) as any as Box3Helper
   helperObject.updateMatrixWorld(true)
-  helperObject.position.set(0, 0, 0)
+
```
HexaField, line 137: depending on how the box helper works internally, we should be able to completely remove the helper update here, and use a ComputedTransformComponent when initializing the helper entity that copies the world position of the box entity to the local position of the helper entity. we want to call updateMatrixWorld on the helper in the computed transform function rather than here too, as there is a change it will be slightly faster and also keeps the scope of responsibility clear for this function.

HexaField (changes requested):

dinomut1 (changes requested):
Bounding boxes are being calculated in world space. They should be calculated in object space.
image

packages/engine/src/gltf/GLTFComponent.tsx
dinomut1: World builder needs the bounding boxes for authoring calculations (placing + rearranging assets, alignment, etc), so would like to have bounding boxes in authoring layer for that.

HexaField (changes requested):
Let's add unit tests to prove this functionality.

PR #2020, opened 2025-06-04
[IR-10677] Add History to sidebar and connect SettingsMenu

likethemammal:
## Summary

Migrated the SettingsMenu to use the Sidebar instead of being an isolated modal. Switched to using a history state and a NavigationProvider that sets the state of the sidebar so that any element can change state and history of the sidebar

image

## QA Steps

- Open toolbar to show Settings cog icon
- Click Settings icon
- SettingsMenu should open in the Sidebar
- Navigating the Settings should work
- The close and back buttons should work to close the sidebar or go back a page in the Settings menu flow

### DIFF
--- a/packages/client-core/src/components/Glass/NavigationProvider.tsx
+++ b/packages/client-core/src/components/Glass/NavigationProvider.tsx
@@ -1,0 +1,108 @@
+/*
+CPAL-1.0 License
+
+The contents of this file are subject to the Common Public Attribution License
+Version 1.0. (the "License"); you may not use this file except in compliance
+with the License. You may obtain a copy of the License at
+https://github.com/ir-engine/ir-engine/blob/dev/LICENSE.
+The License is based on the Mozilla Public License Version 1.1, but Sections 14
+and 15 have been added to cover use of software over a computer network and
+provide for limited attribution for the Original Developer. In addition,
+Exhibit A has been modified to be consistent with Exhibit B.
+
+Software distributed under the License is distributed on an "AS IS" basis,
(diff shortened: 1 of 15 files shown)


packages/client-core/src/components/Glass/NavigationProvider.tsx
```diff
@@ -0,0 +1,108 @@
...
+  hasHistory: boolean
+  direction: DirectionType
+}
+
+const useSidebarNavigation = () => {
+  const [history, setHistory] = useState<string[]>([])
+  const [direction, setDirection] = useState<DirectionType>(1) // 1 for forward, -1 for backward
+  const [sidebarKey, setSidebarKey] = useState(``)
+
+  const navigateTo = (screenKey: string, historyKey: string): void => {
```
Tortillaguy, line 48: Do we need both string params? I see `navigateTo("Settings", ..)` in most places. If the plan is to expect `screenKey` to be something else this is fine with me

Tortillaguy (approved):

packages/client-core/src/components/Glass/NavigationProvider.tsx
likethemammal: I was considering that, but we'll prob need to be able to set a different page other than Settings at some point.

I thought it was cleaner to extend `navigateTo` now, instead of the NavigationProvider assuming the `history` only applies to the Settings page. In the latter case, whoever later adds the other page logic would have to know to make the logic less naive in NavigationProvider.

PR #2021, opened 2025-06-04
Lightmap Baker

AidanCaruso:
## Summary
Adds a lightmap component, which uses a bounding box to volumetrically query visible authoring layer meshes to perform UV atlasing and lightmap baking on. Lightmaps are exported as ktx2, and atlas mesh buffer data is exported as gltfs. Both are loaded and applied to meshes asynchronously on load. 

The idea is to work towards natively lightmapped kitbashed scenes in the engine, to enable functionality with the world builder, etc.

Notes:

I modified and rebuilt the xatlas wasm to use uint32 arrays for mesh indices in order to support the complex meshes we use in the engine, hence the inclusion of the wasm in the client package for loading.

Currently atlas loading and application is handled via lightmap component reactor, asynchonously and inelegantly. In the future, this will be factored into gltf loading.

image

## References
closes https://tsu.atlassian.net/browse/IR-10640


packages/engine/tsconfig.json
```diff
@@ -26,5 +26,5 @@
     "declaration": false,
     "types": ["@types/node"]
   },
-  "include": ["../../__global.d.ts", "./src", "./index.ts"]
+  "include": ["../../__global.d.ts", "./src", "./index.ts", "../editor/src/lightmapper"]
```
HexaField: annoying tsconfig auto-import  :/

packages/editor/tsconfig.json
```diff
@@ -27,5 +27,5 @@
     "types": ["@types/node"]
   },
   "exclude": ["**/node_modules/**", "./src/**/*.test.tsx"],
-  "include": ["../../__global.d.ts", "../client-core/src", "../server-core/src", "./src", "./index.ts"]
+  "include": ["../../__global.d.ts", "../client-core/src", "../server-core/src", "./src", "./index.ts", "../engine/src/lightmap/LightmapComponent.ts"]
```
HexaField: tsconfig nonsense

AidanCaruso: I'll get it. I keep forgetting to scrutinize my git diff before committing

packages/engine/tsconfig.json
HexaField: still not resolved

AidanCaruso: Oops. Thought I got this one.

AidanCaruso:
Marking as draft again as I need to fix issues with the lightmap component's volumetric querying

PR #2022, opened 2025-06-04
[IR-10247] KTX Compression causing extreme performance hit

MichaelEstes:
## References
closes #_insert number here_


HexaField (approved):
awesome

PR #2025, opened 2025-06-05
fix(editor): validate if the entity has component [IR-9706]

mikeplascdev:
## Summary
- Update the empty state for instancingGeometry
- Validate if the entity has a component after release file

## References
closes #_insert number here_

## QA Steps
1. Open studio scene
2. Add particle system
3. Navigate to Particle Mesh field in properties
4. Click and drag a 3D mesh into the field
5. Observe field

## Evidence
https://github.com/user-attachments/assets/2233a14f-4fe2-4a0b-b89d-d19cf617c31e


packages/engine/src/scene/components/ParticleSystemComponent.ts
```diff
@@ -104,11 +106,23 @@ const createBatchedRenderer = (entity: Entity) => {
...
     setComponent(particleRendererEntity, NameComponent, 'Particle Renderer')
     const sceneEntity = getAncestorWithComponents(entity, [SceneComponent])
-    setComponent(particleRendererEntity, UUIDComponent, {
-      entitySourceID: getComponent(sceneEntity, UUIDComponent).entitySourceID,
-      entityID: UUIDComponent.generateUUID()
-    })
-    setComponent(particleRendererEntity, EntityTreeComponent, { parentEntity: sceneEntity })
+
+    // Check if sceneEntity exists and has UUIDComponent before accessing it
+    if (sceneEntity && hasComponent(sceneEntity, UUIDComponent)) {
```
HexaField: i would argue that the rendererEntity and thus sceneEntity MUST exist for this logic to make sense. If not, then the particle is not part of a scene, and therefore is not part of a simulation, and therefore able to be rendered.

HexaField (changes requested):

packages/engine/src/scene/components/ParticleSystemComponent.ts
mikeplascdev: done 68255666c3964ab37c0fe4effe00f774b4dcb854

Tortillaguy (approved):

packages/engine/src/scene/components/ParticleSystemComponent.ts
```diff
@@ -104,10 +104,13 @@ const createBatchedRenderer = (entity: Entity) => {
     setComponent(particleRendererEntity, VisibleComponent)
     setComponent(particleRendererEntity, NameComponent, 'Particle Renderer')
     const sceneEntity = getAncestorWithComponents(entity, [SceneComponent])
-    setComponent(particleRendererEntity, UUIDComponent, {
-      entitySourceID: getComponent(sceneEntity, UUIDComponent).entitySourceID,
-      entityID: UUIDComponent.generateUUID()
-    })
+    const uuidComponent = getComponent(sceneEntity, UUIDComponent)
+    if (uuidComponent) {
```
HexaField: the if condition here is unnecessary

HexaField (changes requested):

packages/engine/src/scene/components/ParticleSystemComponent.ts
mikeplascdev: Hey @HexaField, when I didn’t check if the `uuidComponent` exists and tried to replace a particle mesh file, the app would break. However, with this validation, I can overwrite the files without any issues.

### Evidence without validation
https://github.com/user-attachments/assets/cfb9756a-74df-4a8b-b04a-2d4f301275f3

### Evidence with validation
https://github.com/user-attachments/assets/079cbcf3-f193-44c8-a571-a18af2d569ce

HexaField: this doesn't make any sense, as the scene entity should always have a scene component

mikeplascdev: done 72ce35708e7894c05151cc5774f36bb5e4a4040a

HexaField (approved):

PR #2026, opened 2025-06-05
Avatar preview menu fix and proper entity array boundary usage

HexaField:
## References
closes #_insert number here_


jose-galvan (commented):
I tested these changes because they also resolved another issue where duplicated assets in the scene were only rendering once.

However, I noticed a problem when updating transform properties from an entity in the hierarchy, the changes aren't persisted. The position reverts instantly, even though the input reflects the updated value, it doesn't seem to apply correctly to the component. See the video attached. 

https://github.com/user-attachments/assets/48d87691-8635-496e-afda-7a03075c9774

HexaField:
> However, I noticed a problem when updating transform properties from an entity in the hierarchy, the changes aren't persisted. The position reverts instantly, even though the input reflects the updated value, it doesn't seem to apply correctly to the component. See the video attached.

Addressed.

jose-galvan (approved):

PR #2028, opened 2025-06-05
[MT-INT] Studio: After duplicating an entity, the duplicate is not selected

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-9962


jose-galvan (approved):
There's another line down below that's supposed to do the same thing but it doesn't seem to be working? should that be removed? 

https://github.com/ir-engine/ir-engine/blob/adbe0733239566bfd7628196c114b2f69e5125b2/packages/editor/src/functions/EditorControlFunctions.ts#L267

svyoung (approved):

hzraslan:
Closing that, because ticket is for mt-int env

PR #2030, opened 2025-06-05
IR-6715 Fix Environment Map Baking (Pt. II)

koktavy:
## Summary
Restores envmap bake functionality _and_ captures all entities in the scene. This was achieved by ensuring the simulation layer entities match the visibility of their authoring layer counterparts, and using the existing scene for the bake instead of making a copy. This should also be more performant than making a temp scene copy on each bake. It also has the nifty effect of allowing multiple sequential bakes to show recursive reflections, "infinity mirror" style.

I renamed these component parameter titles for clarity:
- Position Offset -> Bake Position Offset
- Scale -> Box Projection Scale

See commits for further breakdown.

## Subtasks Checklist
N/A

## Breaking Changes
N/A, restored functionality

## References
closes [#6715](https://tsu.atlassian.net/browse/IR-6715)

## QA Steps
- [ ] Create a new empty entity in the scene
- [ ] Add the **Environment Map Bake** component to the empty entity
- [ ] Add the **Environment Map** component to any model in the scene with reflective materials
- [ ] Select "Bake" for Environment Map Source, then select the empty entity as the baking source
- [ ] Observe that the model's reflective materials now show the environment map
- [ ] With Box Projection checked, observe the new box helper to determine the projection of the scene to capture
- [ ] Adjust the Bake Position Offset and Box Projection Scale values to see the projection preview update
- [ ] Bake and save to capture the envmap, refresh to confirm consistency


packages/editor/src/functions/uploadEnvMapBake.ts
```diff
@@ -136,18 +142,24 @@ export const generateEnvmapBake = (
...
-
+  const rendererComponent = getComponent(viewerEntity, RendererComponent)
+  const renderer = rendererComponent.renderer!
+  const scene = rendererComponent.scene!
   const cubemapCapturer = new CubemapCapturer(renderer, scene, resolution)
+
+  // Match the visibility of the simulation layer to the authoring layer, since the cubemap capturer uses the simulation layer
+  const visibilityChanges = new Map<Object3D, boolean>()
+  const simulationQuery = defineQuery([ObjectComponent, LayerComponents[Layers.Simulation]])
+  for (const entity of simulationQuery()) {
```
HexaField: rather than mutating visible, this would probably be easier to use the original logic with a filter ensuring the upstream authoring entity is also visible

HexaField (changes requested):

packages/editor/src/functions/uploadEnvMapBake.ts
koktavy: The original logic has some flaws. It fails to show other entities in the scene when using `scene.children = sceneData.children` and I get an error `Cannot set parent of proxified object` if I try to scene.add directly. In the apartment I can see the main GLTF reflected, but not spotlights or primitives. All reflective surfaces show up black in the bake too, taking away some realism from the nested bake effect.

Another approach would be to make a full copy of the scene, entities and all, but that feels less performant than toggling the visibility.

HexaField: oh, we probably need to get the simulation counterpart for the rootEntity, rather than use the authoring layer for that entity. Can you try the following with the original logic? If this still has issues, we should figure out what entities are still not visible. we should be OK with regards to camera layers, as it defaults to the correct layer.

```ts
const rootEntity = getState(EditorState).rootEntity
const rootEntitySimulation = getSimulationCounterpart(rootEntity) // new line
const entitiesToRender = getNestedVisibleChildren(rootEntitySimulation)
  ```

HexaField (changes requested):

HexaField (approved):

PR #2037, opened 2025-06-06
Add experimental support for WebGPU

AidanCaruso:
## Summary
Completely busted. This PR is a total hack that was functional around 2 weeks ago.

## References
closes #_insert number here_


packages/engine/src/assets/state/AssetLoaderState.ts
```diff
@@ -53,9 +55,9 @@ export const AssetLoaderState = defineState({
     const ktx2Loader = new KTX2Loader()
     ktx2Loader.setTranscoderPath(getState(DomainConfigState).publicDomain + '/loader_decoders/basis/')
     if (isClient) {
-      const renderer = new WebGLRenderer()
-      ktx2Loader.detectSupport(renderer)
-      renderer.dispose()
+      const viewerEntity = getState(ReferenceSpaceState).viewerEntity
```
HexaField, line 62: We can't guarantee the viewer entitiy will exist, we probably just want to conditionally use webgpu same as the renderer component.

packages/engine/src/scene/classes/CubemapCapturer.ts
```diff
@@ -33,23 +33,24 @@ import {
...
-  constructor(renderer: WebGLRenderer, sceneToRender: Scene, resolution: number) {
+  constructor(renderer: WebGLRenderer | WebGPURenderer, sceneToRender: Scene, resolution: number) {
     this.width = resolution
     this.height = resolution
     this.sceneToRender = sceneToRender
     this.renderer = renderer
     this.cubeCamera = null!
     const gl = this.renderer.getContext()
-    const cubeMapSize = Math.min(resolution, gl.getParameter(gl.MAX_CUBE_MAP_TEXTURE_SIZE))
+    const cubeMapSize = gl instanceof WebGLRenderingContext ? gl.getParameter(gl.MAX_CUBE_MAP_TEXTURE_SIZE) : 2048
```
HexaField: Must retain the min(resolution here still, otherwise we could be rendering rather large textures....

packages/spatial/src/common/functions/OnBeforeCompilePlugin.ts
```diff
@@ -175,6 +178,17 @@ export function overrideOnBeforeCompile() {
 
     Material.prototype._onBeforeCompile = function (shader, renderer) {
       if (!this.shader) this.shader = shader
+
+      const rendererEntity = Engine.instance.viewerEntity
```
HexaField: Again here, viewer entity is the main renderer but not the only one.

If need be, we can assign our own properties to the renderer to easily know if it's webgl or webgpu

HexaField (changes requested):

PR #2044, opened 2025-06-09
[IR-10380] Add emote menu to Viewer 1.1

likethemammal:
## Summary

https://github.com/user-attachments/assets/e6c4bf8f-e83f-41f1-b1c9-dd8725c3e13d

## QA Steps

- Open toolbar menu (Arrow button)
- Click emote icon
- The menu should be swapped out for a list of emote icons
- Clicking any of the emote icons should trigger the respective emote on the player's avatar
- Clicking the arrow in the toolbar should swap the toolbar back to the full opened state, no longer showing the emote icons


packages/client-core/src/components/Glass/EmoteMenu.tsx
```diff
@@ -0,0 +1,210 @@
...
+import { emoteAnimations, preloadedAnimations } from '@ir-engine/engine/src/avatar/animation/Util'
+import { AvatarComponent } from '@ir-engine/engine/src/avatar/components/AvatarComponent'
+import { AvatarNetworkAction } from '@ir-engine/engine/src/avatar/state/AvatarNetworkActions'
+import { dispatchAction } from '@ir-engine/hyperflux'
+import React, { SVGProps } from 'react'
+import { MenuButton } from './MenuButton'
+
+const icons = [
+  {
+    Icon: (props: SVGProps<SVGSVGElement>) => (
```
SYBIOTE: i think making a new file for the icon in the UI package, and importing it here is the better option to simply putting this here directly in the Emote Menu file

SYBIOTE (changes requested):

Tortillaguy (changes requested):
Export the SVGs to their own files and this LGTM. Don't worry about which `icon/` folder to put this under for now. 
Also, I find it hard to mentally connect the style strings to their divs because vertically they're so far apart. I do like the newline breaking for grouping classNames though. 
I think if we just made pure components for these divs with hefty styles it will help with readability.

Tortillaguy:
> Export the SVGs to their own files and this LGTM. Don't worry about which `icon/` folder to put this under for now. Also, I find it hard to mentally connect the style strings to their divs because vertically they're so far apart. I do like the newline breaking for grouping classNames though. I think if we just made pure components for these divs with hefty styles it will help with readability.

Even if the names of the components end up being `<ContentWrapper/>` `<TitleBarWrapper/>` etc it helps me understand where in the hierary I'm at

packages/client-core/src/components/Glass/EmoteMenu.tsx
likethemammal: Since we dont have an organized place for these new icons I'm including them relative to this file and will organize them in a final pass later

likethemammal:
> > Export the SVGs to their own files and this LGTM. Don't worry about which `icon/` folder to put this under for now. Also, I find it hard to mentally connect the style strings to their divs because vertically they're so far apart. I do like the newline breaking for grouping classNames though. I think if we just made pure components for these divs with hefty styles it will help with readability.
> 
> Even if the names of the components end up being `<ContentWrapper/>` `<TitleBarWrapper/>` etc it helps me understand where in the hierary I'm at

I see what you're saying but I really dont like creating components unless theres an actual need for them. In my experience it tends to confuse the purpose and importance of the component when they are just an element and styling and not using hooks, providers, or any props. `<ContentWrapper>` seems like its doing something beyond styling and who knows what element or elements its exporting. `<div className={contentWrapperStyles}>` is obviously only a div that has styling on it, no extra assumptions like what html element it is, what props it inherits, or react lifecycles its using.

I am down to reorganize the styles into a separate `ExampleName.styles.js` file to make it easier to follow, instead of keeping the styles at the top. Or maybe we can change our naming convention for style variables to something more readable. But separating the styles from the JSX will become even more necessary when we use tools like `class-variance-authority`

likethemammal:
> Export the SVGs to their own files and this LGTM. Don't worry about which `icon/` folder to put this under for now. Also, I find it hard to mentally connect the style strings to their divs because vertically they're so far apart. I do like the newline breaking for grouping classNames though. I think if we just made pure components for these divs with hefty styles it will help with readability.

Edit: Sorry ur right, I misunderstood, I'll throw these Icons into separate files and keep them relative to this file

SYBIOTE (approved):
Emote menu looks a lot more cleaner with the icons in thier own files, I do think the icons will eventually end up in the UI package , not a priority 
LGTM

Tortillaguy (approved):

jose-galvan (approved):

PR #2052, opened 2025-06-10
refactor(particle-system): extract types and constants to separate files and update imports

mikeplascdev:
## Summary
This PR improves the structure and maintainability of the codebase by:
- Extracting type definitions from components into dedicated `types` files
- Moving constant values into separate `constants` files
- Updating import paths across components to use the new files

These changes simplify component files and encourage the reusability of types and constants

## Subtasks Checklist
- [x] Created `types` files for components
- [x] Created `constants` files where needed
- [x] Removed inline type and constant definitions from components
- [x] Updated all relevant import statements
- [x] Verified that the app builds and runs correctly

## Breaking Changes
_No breaking changes introduced._  
(All changes are internal and non-functional refactors.)

## References
Closes #_insert number here_


HexaField:
Is there a ticket for this?

HexaField (changes requested):
I'd prefer to keep types and constants all in a single file. there's no need to refactor into so many files. this achieves the opposite of maintainability. 3 files is probably fine: one for defaults & types, one for the system, one for the component.

The folder and file names are also not quite what is standardized elsewhere in the codebase. I don't think this needs it's own folder. How it was previously was fine.

The state definition also does not belong in a constants file.

mikeplascdev:
> I'd prefer to keep types and constants all in a single file. there's no need to refactor into so many files. this achieves the opposite of maintainability. 3 files is probably fine: one for defaults & types, one for the system, one for the component.
> 
> The folder and file names are also not quite what is standardized elsewhere in the codebase. I don't think this needs it's own folder. How it was previously was fine.
> 
> The state definition also does not belong in a constants file.

@HexaField 
Done 
- fb673ea52565c7464a5b6dc06f92d645293ecf92
- 97cc666ce47a83ac56cb2e5f8e961c79e6683fda
- 6616c85f0287fdd99d1b9f82cb5c1cb4fc9a82a2

and we don't have a ticket for this, I'm working on particle system issues, and I wanted to add these changes

HexaField (approved):

PR #2056, opened 2025-06-10
[IR-8457] Scene deltas as GLTF loading dependency

MichaelEstes:
## Summary
Also prevents the user from saving a scene that has not fully loaded yet
## References
closes #_insert number here_


packages/engine/src/authoring/AuthoringState.tsx
```diff
@@ -177,14 +178,15 @@ export const AuthoringState = defineState({
 
   reactor: () => {
     const state = useMutableState(AuthoringState)
+    const layer = useMutableState(EngineState).isEditing.value ? Layers.Authoring : Layers.Simulation
```
HexaField: rather than isEditing, we should use getAuthoringCounterpart on each entity returned from the query, (which will fall back to the simulation layer if an authoring layer entity doesn't exist)

we also need to make sure this does not run for avatars, without referencing avatars.

HexaField (changes requested):

packages/engine/src/authoring/AuthoringState.tsx
```diff
@@ -267,21 +274,31 @@ export const AuthoringState = defineState({
...
+    const sourceID = UUIDComponent.getAsSourceID(entity)
+    const sourceData = getSourceSnapshot(sourceID, layer)
 
     dispatchAction(AuthoringActions.initialize({ sourceID, partialState: sourceData }))
 
     return () => {
       dispatchAction(AuthoringActions.uninitialize({ sourceID }))
     }
-  }, [loaded])
+  }, [loaded, valid])
```
HexaField: do we need to merge this into a single bool dep to avoid it retriggering due to potential race conditions with the network object component?

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -393,16 +425,41 @@ const DependencyEntryReactor = (props: { gltfComponentEntity: Entity; uuid: Enti
...
+
   useEffect(() => {
     return () => {
       removeError(gltfComponentEntity, GLTFComponent, 'LOADING_ERROR')
       removeError(gltfComponentEntity, GLTFComponent, 'INVALID_SOURCE')
     }
   }, [])
 
+  useEffect(() => {
+    const scene = commands['scene'].get(NO_PROXY) as HistoryCommand[] | undefined
```
HexaField: must use SceneUser variable instead of string

HexaField (changes requested):

HexaField (approved):

PR #2059, opened 2025-06-11
fix: disable client and add script to create default buckets

jose-galvan:
## Summary

MinIO has removed the admin web interface from the community version, which causes the container to stop shortly after starting in newer versions.

This update skips the client setup (which causes the crash) and ensures the default buckets are still created via a startup script.

https://cloudian.com/blog/minios-ui-removal-leaves-organizations-searching-for-alternatives/

The certificates are also regenerated since they expired a few days ago. 

## References
closes #_insert number here_


hanzlamateen (changes requested):
Getting error in dev-reinit.  Previously on reinit container was started


https://github.com/user-attachments/assets/111eb5a6-8f37-4149-b1e2-2989ffea1373

hanzlamateen:
@jose-galvan I removed `chmod -x` from entrypoint command in docker-compose and its working fine now.

Does removing it impacts anything on your end?

jose-galvan:
> @jose-galvan I removed `chmod -x` from entrypoint command in docker-compose and its working fine now.
> 
> Does removing it impacts anything on your end?

it needs to execute the script to create the default buckets, if they are already created it's not an issue but if they aren't it will cause some issues when running the engine. 

I made some improvements, can you test again, please?

hanzlamateen (changes requested):
Again the init container is failing on chmod

image

hanzlamateen:
Merging with reverted file and chmod command removed.

PR #2070, opened 2025-06-11
[IR-10447] Rework image loading limits on iOS

MichaelEstes:
## Summary
Adds queries to Resource state around asset memory usage and budgets 
Limit concurrent images being loaded
Wait for images to be discarded if over budget before loading
## References
closes #_insert number here_


packages/spatial/src/resources/ResourceState.tsx
```diff
@@ -710,13 +810,22 @@ const getAllResourcesOfType = (type: ResourceType) => {
   return result
 }
 
+const assetDiscarded = () => {
```
HexaField, line 820: is it necessary to have listeners? can't we just react to the `discarded` state in ResourceState for each resource?

HexaField (changes requested):

packages/spatial/src/resources/ResourceState.tsx
MichaelEstes: There's no reactive context in the texture loader to listen to those changes

HexaField: we can do `const reactor = startReactor(() => { ... })` to reactivify a promise, we do it in a few places already

HexaField (changes requested):

the-reality-engineer (approved):

HexaField:
If we no longer need this for iOS to run consistently (we can re-evaluate after the ECS refactor and optimization flow is more comprehensive), then i think we should mark as draft and park it here. I want to minimize the additional complexity in our loaders.

PR #2079, opened 2025-06-12
[MT-STG] - Loading Spinner constantly present in the scene, including wizard flow

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-10746?linkSource=email

### DIFF
--- a/packages/engine/src/gltf/GLTFComponent.tsx
+++ b/packages/engine/src/gltf/GLTFComponent.tsx
@@ -508,7 +508,6 @@ export const parseGLTFFile = (
     json = JSON.parse(JSON.stringify(json))
 
     json = migrateEEMaterial(json)
-
     return [parseStorageProviderURLs(json), body]
   } catch (error) {
     if (onError) onError(error)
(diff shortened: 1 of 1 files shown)


MbfloydIR (approved):

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -525,10 +524,11 @@ const useGLTFDocument = (entity: Entity) => {
   const isEditing = layer === Layers.Authoring
 
   const dynamicLoadAndNotEditing = !isEditing && !!dynamicLoadComponent && !dynamicLoadComponent?.loaded?.value
-
+  const gltfComponent = getMutableComponent(entity, GLTFComponent)
   useEffect(() => {
     if (dynamicLoadAndNotEditing) return
     if (!url) {
+      gltfComponent.progress.set(100)
```
DanielBelmes, line 531: is it possible to move the logic of not showing the spinner to the ownership of the spinner? Can it not check for the url itself and just not be added?(early return)

hzraslan: We are already doing that in the spinner and the issue is, When we add 3d blank model to scene, at the beginning it already has an url and then in the process we remove it. So my observation is, without progress 100 it actually fails to add 3d blank model to scene without any error.

DanielBelmes: Odd yea I check the LoadingSpinnerReactor and it is checking for it properly

PR #2081, opened 2025-06-12
Fix Multiplayer Toggle

Tortillaguy:
## Summary
Multiplayer functionality is controlled by the `online` arg of `useNetwork` when rendering the `/location` route. 
The arg was hard-coded as true, but is now reactively reading from the NetworkState object

https://www.loom.com/share/3435ea9a38c546aab773804eab28ff05

## References
closes [IR-10560] [IR-11084]

## QA Steps

[IR-10560]: https://tsu.atlassian.net/browse/IR-10560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[IR-11084]: https://tsu.atlassian.net/browse/IR-11084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/client-core/src/components/World/EngineHooks.tsx
```diff
@@ -80,7 +81,23 @@ export const useNetwork = (props: { online?: boolean }) => {
...
+    const networkState = getMutableState(NetworkState)
+
+    // worldId initializes as null
+    const worldId = networkState.hostIds.world.value
+
+    const selfHosting = worldId === (userID as any as InstanceID)
+    // Refresh the page to return online
+    // worldId becomes undefined after disconnecting
+    if (selfHosting || worldId === undefined) {
+      window.location.reload()
```
HexaField: there is no need to reload when going from offline to online or vice versa

packages/client-core/src/world/Location.tsx
```diff
@@ -64,7 +60,11 @@ const LocationPage = ({ online }: Props) => {
   let [glassDisabled] = useFeatureFlags([FeatureFlags.Client.Glass])
   glassDisabled = glassDisabled && searchParams.get('glassUI') === null
 
-  useNetwork({ online })
+  const {
```
HexaField: this doen't make any sense, you are passing in values that are then used to set the same state

HexaField (changes requested):

HexaField (changes requested):
I would highly recommend creating a separate hyperflux state that handles the MainMenu buttons, and instead of mutating the NetworkConfig directly, it should be used in EngineHooks as another potential disabling bool for first useEffect.

There is no need to reload either, joining and leaving the network will be taken care of if the existing states are used correctly.

packages/client-core/src/common/services/MultiplayerState.ts
```diff
@@ -0,0 +1,34 @@
...
+Original Code is the Infinite Reality Engine team.
+
+All portions of the code written by the Infinite Reality Engine team are Copyright © 2021-2025
+Infinite Reality Engine. All Rights Reserved.
+*/
+
+import { defineState, syncStateWithLocalStorage } from '@ir-engine/hyperflux'
+
+export const MultiplayerState = defineState({
+  name: 'MultiPlayer State',
```
HexaField: state name should be something like `ir.client.user.MultiplayerState` to avoid collisions

packages/client-core/src/components/Settings/MainMenu.tsx
```diff
@@ -53,8 +54,10 @@ const MainMenu: React.FC<ScreenProps> = ({ navigateTo }) => {
         <Divider />
         <ToggleItem
           label="Video Communication"
-          checked={videoCommunication}
-          onClick={() => setVideoCommunication(!videoCommunication)}
+          checked={config.media.value}
```
HexaField: can we do the same thing for media? it should also be in the MultiplayerState

HexaField (changes requested):

packages/client-core/src/common/services/MultiplayerState.ts
```diff
@@ -26,9 +26,10 @@ Infinite Reality Engine. All Rights Reserved.
 import { defineState, syncStateWithLocalStorage } from '@ir-engine/hyperflux'
 
 export const MultiplayerState = defineState({
-  name: 'MultiPlayer State',
+  name: 'ir.client.user.MultiplayerState',
   initial: () => ({
-    enabled: false
+    enabled: false,
```
HexaField: let's rename this to 'world' and 'media' to reflect what is NetworkState

HexaField (dismissed):

packages/client/src/pages/location/location.tsx
```diff
@@ -51,11 +52,16 @@ const LocationRoutes = () => {
 
   const projectsLoaded = useEngineInjection()
 
+  const multiplayer = useMutableState(MultiplayerState).world
+  useEffect(() => {
+    multiplayer.set(true)
```
HexaField: what's the intention behind setting this to true here?

DanielBelmes:
I was gonna hook up the media controls too but I need to find the PDR to find what they are even supposed to do. I assume disable video/audio chat completely. Which we can work on tomorrow really quickly together.

packages/client/src/pages/location/location.tsx
Tortillaguy: The initial value in the state is `false`. Do you want it to be true? 
This is currently following the same methdology as the `online` prop

Tortillaguy: Also I want to note this is touching live code. Changing behavior here will affect the next release even if we're not using the new UI yet

HexaField: Yeah we want this to be true by default

DanielBelmes (approved):

PR #2082, opened 2025-06-12
IR-10563 Avatar invisible in published scene

danskate:
## References
closes #_insert number here_


packages/engine/src/avatar/state/AvatarNetworkState.tsx
```diff
@@ -99,11 +100,40 @@ const AvatarReactor = ({ entityUUID }: { entityUUID: EntityUUID }) => {
   useEffect(() => {
     if (!entity || !avatarURL.value) return
 
-    setComponent(entity, GLTFComponent, { src: avatarURL.value })
+    const testAvatarURL = async (url: string) => {
```
HexaField, line 103: this logic should already be more or less captured in GLTFComponent. we also have a 'recovery' mechanism in AvatarSpawnState that attempts to change your avatar if it errors when loading. that logic probably isn't happening, and we should figure out why.

HexaField (changes requested):

packages/engine/src/avatar/state/AvatarNetworkState.tsx
danskate: I see, I'll look into the existing recovery mechanism.

danskate: Testing the URL wasn't necessary. I think all that did was introduce a delay between removing the GLTFComponent and re-adding it. If I replace testAvatarURL with a 1ms timeout it also allows the avatar to load correctly. This is a bandaid though and I'm still looking for the root cause.

HexaField: merge dev in and see if the problem still occurs, a bug related to loading models was fixed here https://github.com/ir-engine/ir-engine/pull/2026

danskate: I am still seeing avatar issues in dev. The avatars aren't appearing after selection and they are appearing in strange poses in the preview after visiting the change avatar page a second time.

https://github.com/user-attachments/assets/12104e2d-26d0-4b82-9a32-34efb2d236d4

PR #2087, opened 2025-06-13
IR-7193 link component interaction

SamMazerIR:
## Summary
fixed issue with input processing that prevented anything other than the first key of a button alias from having its state evaluated (alias for interact for example is **left click** OR **controller trigger** OR **E key**)

## References
[https://tsu.atlassian.net/browse/IR-7193](https://tsu.atlassian.net/browse/IR-7193)

## QA Steps
set up a link component by adding one to a primitive geometry or mesh
walk up to it in preview/location and press E (clicking always worked because it was the first...and only... key checked in 'Interact' button alias of click/trigger/E)

### DIFF
--- a/packages/spatial/src/input/components/InputComponent.test.tsx
+++ b/packages/spatial/src/input/components/InputComponent.test.tsx
@@ -490,6 +490,76 @@ describe('InputComponent', () => {
     })
   })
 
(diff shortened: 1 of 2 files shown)


hzraslan (approved):
test is failed, otherwise approved

PR #2088, opened 2025-06-13
IR-10378

svyoung:
## Summary
[IR-10378](https://tsu.atlassian.net/browse/IR-10378)
For user moderation, we are required legally to have basic reporting and moderation functionality.  Without a full-blown social system (we don’t/won’t have one of these for now), we will simply implement the existing functionality for reporting that we have today. 

## References
[IR-10378](https://tsu.atlassian.net/browse/IR-10378)

## QA Steps
moderation_button1
moderation_button2

[IR-10378]: https://tsu.atlassian.net/browse/IR-10378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IR-10378]: https://tsu.atlassian.net/browse/IR-10378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/client-core/src/components/Glass/VideoMenu.tsx
```diff
@@ -168,6 +172,12 @@ const Video = ({ peerID, type }: WindowType) => {
   const camVideoOn = isSelf ? isCamVideoEnabled : !videoStreamPaused
   const camAudioOn = isSelf ? isCamAudioEnabled : !audioStreamPaused
 
+  const reportUser = () => {
+    navigateTo('ReportUser', '')
+    ReportUserState.setReportedPeerId(peerID)
```
likethemammal, line 176: 🐑 Itd be better for UX to set these state changes above and then `navigateTo`. That way the UI is already ready when the user navigates and sees it.

packages/client-core/src/components/ReportUser/index.tsx
```diff
@@ -0,0 +1,350 @@
...
+      case REPORT_INPROGRESS:
+        return reportProgress
+      case REPORT_SUCCESS:
+        return reportSucess
+      default:
+        return reportProgress
+    }
+  }
+
+  return <div className="report-user-menu w-full">{showContents()}</div>
```
likethemammal: I dont think we need this class `report-user-menu` class

packages/client-core/src/components/Glass/VideoMenu.tsx
```diff
@@ -281,7 +291,10 @@ export const VideoMenu = ({ videos = [] }: { videos: WindowType[] }) => {
 
       <div className={arrowsContainer}>
         <button
-          onClick={() => setPageIndex(pageIndex - 1)}
+          onClick={() => {
+            setPageIndex(pageIndex - 1)
+            console.log('well hellooooo')
```
likethemammal: 😋 extra console.log

packages/client-core/src/components/ReportUser/index.tsx
```diff
@@ -0,0 +1,350 @@
...
+    details: '',
+    files: ''
+  })
+
+  /** may need to move this to its own provider if used in more than several areas */
+  const fieldOptions = {
+    abuseType: {
+      label: t('user:usermenu.profile.typeAbuse'),
+      validate: (input: string) => {
+        if (input === null || input === '' || input === 'null') {
```
likethemammal, line 108: 🐑 Its normally worth trimming the input value before validating, otherwise an empty space/tab/return will pass this validation.

```suggestion
        input = input?.trim()
        if (input === null || input === '' || input === 'null') {
```

packages/client-core/src/components/ReportUser/index.tsx
```diff
@@ -0,0 +1,350 @@
...
+    ReportUserState.resetReportUser()
+    setContent(REPORT_INPROGRESS)
+    handleChange('', 'abuseType')
+    handleChange('', 'details')
+    errors.abuseType.set('')
+    errors.details.set('')
+  }
+
+  const reportProgress = (
+    <div className={twMerge(inputDivStyles, 'gap-y-6')}>
```
likethemammal: Could you make this a different class and name it differently, something like `containerStyles`, since the styles have a different purpose. The only common styles between them is the `flex flex-col` which isnt work making a class for and will cause confusion later

packages/client-core/src/util/ReportUserState.tsx
```diff
@@ -28,12 +28,19 @@ import { defineState, getMutableState, PeerID } from '@ir-engine/hyperflux'
 export const ReportUserState = defineState({
```
likethemammal: I dont think you are using this `reportingUser` state anymore, so the changes to this file arent needed and the function calls for `toggleReportUser` and `resetReportUser` arent needed in the other files

packages/ui/src/components/viewer/Dropdown/index.tsx
```diff
@@ -89,9 +94,9 @@ const Dropdown: React.FC<DropdownProps> = ({
         onClick={handleToggle}
         disabled={disabled}
         className={`
-          flex w-full items-center justify-between rounded-lg border border-white/20 
-          bg-white/10 px-4 py-3 text-left text-white backdrop-blur-sm
-          ${disabled ? 'cursor-not-allowed opacity-50' : 'cursor-pointer hover:bg-white/15'}
+          flex w-full items-center justify-between rounded-lg ${border && 'border'} border-white/20 
+          bg-${bgColor}/10 px-4 py-3 text-left text-white backdrop-blur-sm
```
likethemammal: We shouldnt construct dynamic classes like this `bg-${bgColor}/10` because Tailwind generates these classes after the fact. This only work because somewhere else in the codebase we happen to have statically written `bg-white/10` and `bg-black/10`.

It'd be better to use Tailwind merge with a conditional for either adding `bg-white/10` or `bg-black/10`

likethemammal: 🐑 This will end up a lot cleaner once we use `class-variance-authority` and have a clear way of defining variants

packages/client-core/src/user/VideoWindows/window.tsx
```diff
@@ -33,8 +33,7 @@ import {
   webcamAudioMediaChannelType,
```
likethemammal: The changes to this file arent needed since this component isnt used with the Glass components

likethemammal (changes requested):

packages/client-core/src/components/Glass/VideoMenu.tsx
svyoung: woops!

packages/client-core/src/components/ReportUser/index.tsx
```diff
@@ -0,0 +1,313 @@
...
+    setContent(REPORT_INPROGRESS)
+    handleChange('', 'abuseType')
+    handleChange('', 'details')
+    errors.abuseType.set('')
+    errors.details.set('')
+  }
+
+  const reportProgress = (
+    <div className={twMerge(containerStyles, 'gap-y-6')}>
+      <div className={twMerge(containerStyles, 'gap-y-2')}>
```
likethemammal: Sorry what I meant was

1. `inputDivStyles` shouldnt be reused for both the div wrapping the input and the div that wraps everything. The elements serve different purposes
2. I see why you wanted to share the styles, but they are so minimal that a common class like that is less immediately understandable and causes confusion if the styles need changing.

I would do something like this:

```ts
const baseContainerStyles = `
  flex flex-col
  text-white
`

const containerStyles = twMerge(baseContainerStyles, 'gap-y-6')

const inputContainerStyles = twMerge(baseContainerStyles, 'gap-y-2')

...
```

```tsx
<div className={containerStyles}>
  ...
  <div className={inputContainerStyles}>
    ..
  </div>
  <div className={inputContainerStyles}>
    ..
  </div>
```

And then you can reuse `inputContainerStyles` for the other input container divs.

likethemammal (approved):

PR #2089, opened 2025-06-13
Moved Engine Render Canvas to React

DanielBelmes:
## Summary
We've been having issues in the editor for when rc-dock full rerenders the moved canvas it loses the canvas since react doesn't know it exists there. With this change we now just define a canvas directly and call useEngineCanvas on it. This is mostly in a functioning state but WebGL still errors when rc-dock rerenders the canvas.


HexaField (changes requested):
we need to maintain the styling we have or things will break. i have wanted to make this more generic like with how `useRender3DPanelSystem` works so this is a good step, i reckon we should go all the way if possible.

packages/client/src/pages/location/location.tsx
```diff
@@ -63,6 +64,7 @@ const LocationRoutes = () => {
           <LoadingView fullScreen animated title={t('common:loader.loadingApp')} titleClassname="text-black" />
         </div>
       )}
+      {createPortal(<canvas id="engine-renderer-canvas" tabIndex={1} ref={canvasRef}></canvas>, document.body)}
```
DanielBelmes: leaving a note here that this can likely be temporary as I just have to fix input handling since a lot of our input handling relied on the canvas being in document.body

PR #2090, opened 2025-06-13
Optimize flow should limit texture size to 1K

hzraslan:
## References
closes  https://tsu.atlassian.net/browse/IR-10160


packages/common/src/model/ModelTransformFunctions.ts
```diff
@@ -1282,57 +1285,21 @@ export async function safeCompressGLTFWeb(
...
-        error: 0.001,
+        ratio: params.simplifyRatio,
+        error: params.simplifyErrorThreshold,
         simplifier: MeshoptSimplifier
       })
     )
 
     document.createExtension(KHRMeshQuantization)?.setRequired(true)
     onProgress?.(0.8, Status.WritingFiles)
-    const eeMaterialExtension: EEMaterialExtension | undefined = document
```
SamMazerIR, line 1293: not sure if we should remove this portion, @AidanCaruso would be good to check with

HexaField: yeah ee material is deprecated, can be removed

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
```diff
@@ -366,14 +333,16 @@ export default function AddEditLocationModal(props: AddEditLocationModalProps) {
...
         await handlePublish(true)
         //re-open the original scene
         const studioUrl = `${window.location.origin}/studio?project=${projectName}&scenePath=${scenePath}`
-        window.open(studioUrl, '_blank')?.focus()
+        // window.open(studioUrl, '_blank')?.focus()
         ModalState.closeModal()
         progressState.set({ progress: 0, caption: '' })
       }
     } catch (error) {
+      console.log('error1', error)
```
DanielBelmes: You can console.error out the error

packages/engine/src/assets/compression/extensions/EEPRO_EcommerceProductTransformer.ts
```diff
@@ -0,0 +1,256 @@
...
+  activationDistance: number
+  useOverlay: boolean
+  xOffset: number
+  yOffset: number
+  gltfSources: string[]
+  mediaSources: string[]
+  featuredMediaSource: string
+}
+
+export class EEPROEcommerceProduct extends ExtensionProperty<IEEPROEcommerceProduct> {
```
HexaField: we cannot have the ecommerce extension in the core engine. furthermore, we have hundreds of serializable components, this is not a scalable solution. the components themselves already define schemas, this needs to be driven by the component definitions we already have.

HexaField (dismissed):

HexaField (approved):

PR #2097, opened 2025-06-16
IR-10715 Users soft-delete instead of hard-delete feature

kattsushi:
### Implement User Soft Delete Functionality
This PR modifies the core authentication service to support user deactivation instead of permanent deletion. The changes enable a "soft delete" approach where user accounts are preserved but marked as deactivated.

### Problem Addressed
Users who delete their accounts are currently permanently removed from the system. This causes several issues:

- Inability for users to reactivate their previous accounts
- Potential creation of duplicate accounts when users re-register

### Solution
The authentication service now checks for a user's deactivation status during login and redirects deactivated users to a dedicated page. This preserves user records while preventing access to the platform.

### Benefits
Account Recovery: Allows for potential account reactivation in the future
Improved User Management: Provides better visibility into user lifecycle

### Testing
Verified that deactivated users cannot access protected routes
Confirmed proper redirection to the deactivation notification page
Tested authentication flows with both active and deactivated accounts

## References

Ticket: [IR-10715](https://tsu.atlassian.net/browse/IR-10715)
This PR is required for irpro-multitenancy project PR as part of completion of this issue here [1054](https://github.com/theinfinitereality/irpro-multitenancy/pull/1054) 

## QA Steps

[IR-10715]: https://tsu.atlassian.net/browse/IR-10715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/server-core/src/user/user/user.resolvers.ts
```diff
@@ -63,7 +64,13 @@ export const userDataResolver = resolve<UserType, HookContext>({
...
 
 export const userPatchResolver = resolve<UserType, HookContext>({
-  updatedAt: getDateTimeSql
+  updatedAt: getDateTimeSql,
+  deactivatedAt: async (deactivatedAt) => {
+    if (deactivatedAt) {
+      return getDateTimeSql()
+    }
+    return undefined
+  }
```
hanzlamateen: This check should be based on `isDeactivated` prop

```suggestion
  deactivatedAt: async (_, userData) => {
    if (userData.isDeactivated) {
      return getDateTimeSql()
    }
    return undefined
  }
```

packages/server-core/src/user/user/user.hooks.ts
```diff
@@ -82,12 +82,15 @@ const restrictUserPatch = async (context: HookContext<UserService>) => {
...
-  // If a user without admin and user:write scope is patching themself, only allow changes to avatarId and name
+  // If a user without admin and user:write scope is patching themself, only allow changes to specific properties
   const process = (item: UserType) => {
     const data = {} as UserPatch
     // selective define allowed props as not to accidentally pass an undefined value (which will be interpreted as NULL)
     if (typeof item.name !== 'undefined') data.name = item.name
     if (typeof item.ageVerified !== 'undefined') data.ageVerified = item.ageVerified
+    // Allow users to deactivate their own accounts
+    if (typeof item.isDeactivated !== 'undefined') data.isDeactivated = item.isDeactivated
+    if (typeof item.deactivatedAt !== 'undefined') data.deactivatedAt = item.deactivatedAt
```
hanzlamateen: `deactivatedAt` shouldnt be passed from client.

packages/client-core/src/admin/components/user/UserTable.tsx
```diff
@@ -48,19 +48,22 @@ import ActionButton from '../ActionButton'
...
   },
   users: UserType[]
 ) => {
   modalProcessing.set(true)
   await Promise.all(
     users.map((user) => {
-      adminUserRemove(user.id)
+      adminUserPatch(user.id, {
+        isDeactivated: true,
+        deactivatedAt: new Date().toISOString()
```
hanzlamateen: We shouldnt pass deactivatedAt here.

packages/server-core/src/user/user/user.hooks.ts
```diff
@@ -82,12 +82,15 @@ const restrictUserPatch = async (context: HookContext<UserService>) => {
   if (loggedInUser.id !== context.id)
     throw new Error("Must be an admin with user:write scope to patch another user's data")
 
-  // If a user without admin and user:write scope is patching themself, only allow changes to avatarId and name
+  // If a user without admin and user:write scope is patching themself, only allow changes to specific properties
```
hanzlamateen, line 85: Since we are not longer using remove method of user service. Therefore should remove it from `userMethods` array in schema file and hooks?

hanzlamateen, line 85: Also, how are you checking that removed user is no longer able to make api calls using user-api-key token service?

packages/client-core/src/user/menus/ProfileMenu.tsx
```diff
@@ -703,34 +702,6 @@ const ProfileMenu = ({ hideLogin, onClose }: Props): JSX.Element => {
             <LogIn01Lg />
             {t('user:usermenu.profile.logout.submit')}
           </button>
-
-          <button
```
hanzlamateen, line 707: I also see a `packages/client-core/src/components/Settings/DeleteAccountScreen.tsx` control which can delete a user. Where is it used? It still introduces the loop hole to cause the original issue.

We can get rid of `AuthService.removeUser` method if above question is sorted out.

hanzlamateen (changes requested):

packages/client-core/src/admin/components/user/UserTable.tsx
```diff
@@ -48,19 +48,22 @@ import ActionButton from '../ActionButton'
 import AccountIdentifiers from './AccountIdentifiers'
 import AddEditUserModal from './AddEditUserModal'
```
hanzlamateen, line 49: Also add the isDeactivated data in table's column to display it in admin panel. And show all users instead of restricting to only active ones.

packages/client-core/src/admin/components/user/UserTable.tsx
```diff
@@ -48,19 +48,22 @@ import ActionButton from '../ActionButton'
 import AccountIdentifiers from './AccountIdentifiers'
```
hanzlamateen, line 48: Also, how are you ensuring in core engine that a deactivated user is not able to login?

hanzlamateen (changes requested):

packages/server-core/src/user/user/user.resolvers.ts
kattsushi: addressed ty!

packages/client-core/src/admin/components/user/UserTable.tsx
kattsushi: are able to login but without access to the dashboad either the viewer, redirect to deactivated page instead

packages/server-core/src/user/user/user.hooks.ts
kattsushi: good catch up ty, added the flag to the authenticated hook and create a little test for it

packages/server-core/src/user/magic-link/magic-link.class.ts
```diff
@@ -153,6 +154,21 @@ export class MagicLinkService implements ServiceInterface<MagicLinkParams> {
       })
     ).data
 
+    if (identityProviders.length > 0) {
+      try {
+        const user = await this.app.service(userPath).get(identityProviders[0].userId)
```
barankyle: Should this iterate over all of the identityProviders it gets back if length > 0? If an email was used to make an account, then deactivated, then makes a second account, wouldn't this possibly have the deactivated IP/user in the 0th position in the array, and the activated IP/user in the 1st position? This check as it currently stands would change IP.length to 0, and a new IP/user would be created even though there's an active user using that email.

kattsushi: Ty already fixed

barankyle (approved):

hanzlamateen (approved):

MoizAdnan (approved):

PR #2099, opened 2025-06-16
IR-9784 fix for camera mode reactivity

SamMazerIR:
## Summary
adding subreactor to handle reactivity for cameraSettingsComponent and entity. gives reactive dep for useEffect

## References
[https://tsu.atlassian.net/browse/IR-9784](https://tsu.atlassian.net/browse/IR-9784)

## QA Steps
1. create at least 2 POI entities in scene, change camera settings to POI mode, add at least 2 POI entities, choose desired settings for snapping/scrolling etc
2. publish a scene with the poi camera set up
3. confirm published scene is in POI camera mode (scroll/snap between POI entity locations with camera on scrollwheel or UI button press as described in ticket)


packages/client-core/src/networking/AvatarSpawnSystem.tsx
```diff
@@ -211,11 +216,35 @@ const reactor = () => {
...
+  const userID = useMutableState(EngineState).userID.value
+  const locationSceneURL = useHookstate(getMutableState(LocationState).currentLocation.location.sceneURL).value
+  const sceneEntity = useLoadedSceneEntity(locationSceneURL)
+  const gltfLoaded = GLTFComponent.useSceneLoaded(sceneEntity)
+
+  const cameraSettingsComponents = useChildrenWithComponents(sceneEntity, [CameraSettingsComponent])
+  const cameraSettingsEntity = cameraSettingsComponents.length > 0 ? cameraSettingsComponents[0] : null
+  const currentCameraMode = useHookstate<CameraModeType>(CameraMode.FOLLOW)
+
+  const handleCameraModeChange = (cameraMode: CameraModeType) => {
```
HexaField, line 223: rather than hoisting state, the AvatarSpawnReactor should also just get the camera mode and ensure it is FOLLOW

HexaField (dismissed):

packages/client-core/src/networking/AvatarSpawnSystem.tsx
SamMazerIR: I'm confused, the AvatarSpawnReactor should run or not depending on the currentCameraMode. This is being updated from the CameraSettingsReactor's onCameraModeChange, and that reactor early outs when the entity is null because we're checking an array from useChildrenWithComponents to avoid a query. I could have set this all up with a useQuery instead but thought we preferred useChildrenWithComponents in situations like this.

DanielBelmes (approved):

PR #2101, opened 2025-06-16
use the useMutableState hook to prevent error on mt-dev

DanielBelmes:
## Summary
useHookstate here was inconsistently throwing an error on mt-dev. I believe using useMutableState has some internal logic that prevents this error. It isn't super clear.

## References
closes #_insert number here_


MichaelEstes (approved):

HexaField:
That's not what's going on here. `useHookstate(getMutableState(` is identitcal to `useMutableState(`

This is fixed by https://github.com/ir-engine/ir-engine/pull/2094 it's just reporting on the next hook since it's trying to render with a different number, which is what React 103 means

PR #2108, opened 2025-06-17
[IR-10992] Ensure resource progress component is always properly removed when resources are finished loading

MichaelEstes:
## References
closes #_insert number here_


HexaField (changes requested):
Screenshot 2025-06-18 at 8 00 30 pm

I'm seeing the ResourceProgressComponent sticking around on the Settings entity on the default scene still

MichaelEstes:
> Screenshot 2025-06-18 at 8 00 30 pm
> 
> I'm seeing the ResourceProgressComponent sticking around on the Settings entity on the default scene still

I can't seem to reproduce this, this is from loading the default scene location as is?

HexaField:
> I can't seem to reproduce this, this is from loading the default scene location as is?

Yeah. Odd I'll try again.

packages/ecs/src/UUIDComponent.ts
```diff
@@ -241,6 +242,11 @@ export const UUIDComponent = defineComponent({
   /** Construct a new SourceID from the concatenated values of the source entity */
   getAsSourceID: (entity: Entity) => UUIDComponent.join(getComponent(entity, UUIDComponent)) as any as SourceID,
 
+  useAsSourceID: (entity: Entity) => {
+    const uuidComponent = useOptionalComponent(entity, UUIDComponent)
```
HexaField, line 246: we may considering changing this to a useComponent to forcibly suspend the context any time we expect a UUIDComponent

HexaField (approved):

PR #2117, opened 2025-06-18
[IR-11027] Attach routing to sidebar and add dynamic routing

likethemammal:
## Summary

- NavigationProvider needed a way to add dynamic routes so the Shopify Cart could inject itself into the Viewer.
- This makes the routing logic more flexible, allowing for any depth of route, and it connect the routing to the actual React router and URL
- This provides browser history functionality like back and refresh on any sidebar page

https://github.com/user-attachments/assets/635ac19d-428a-4fe3-a855-ca12a0c10ff3

## QA Steps

- Open viewer scene
- Clicking any button that would open the sidebar should open the sidebar
- When the sidebar is open, the url should show the route
- Clicking the back button in the browser should navigate backwards
- When refreshing the page, if the sidebar was open it should be open to the same page


svyoung (approved):
Looks good

PR #2124, opened 2025-06-18
Cleanup GLTF body reference

the-reality-engineer:
## Summary
This allows for the GLTFComponent body ArrayBuffer to be cleaned up when it is no longer needed. 

This represents about a 20% memory reduction in a typical location.


packages/engine/src/assets/functions/resourceLoaderFunctions.ts
```diff
@@ -117,7 +117,10 @@ export const loadResource = <T extends ResourceAssetType>(
         onError(new Error('Resource removed before load finished'))
         return
       }
-      resource.asset.set(response)
+      // only store cloneable assets
+      if (isCloneable(resourceType)) {
```
MichaelEstes, line 121: Is this right? We need the asset reference to call the dispose function on the asset if it exists

the-reality-engineer: We were previously storing several copies of ArrayBuffers for entire gltf files in several places throughout the codebase; this change means we only store clonable texture / material / geometry / etc objects. Would you recommend a different approach?

MichaelEstes: Ya, only textures are cloneable, but that does seem to be the only thing we would need to care about in this flow

MichaelEstes (approved):
Looks good, everything seems to be loading in fine. Minor nitpick that having a field named "loaded" on the GLTF component might lead to confusion on whether it represents the GLTF file being loaded vs the GLTF and all it's dependencies having been loaded.

HexaField:
> Looks good, everything seems to be loading in fine. Minor nitpick that having a field named "loaded" on the GLTF component might lead to confusion on whether it represents the GLTF file being loaded vs the GLTF and all it's dependencies having been loaded.

i agree, we can clean this up in a follow up as we need to get iOS working

PR #2130, opened 2025-06-19
fix particles being removed when other particles are deleted

SYBIOTE:
fix other particles disappearing when a particle system is removed


BogieBogard (approved):

HexaField:
Will this cause other particles to be restarted when another particle is added or removed? @DanielBelmes i spoke to Rahul about this, this is a hack fix that is obscuring the core problem (bug with reference counting for particle renderers it seems) that likely will cause other bugs

SYBIOTE:
from what I have noticed, triggering the particle system reactor itself, causes all particles in the scene to restart, since the reactor reacts on visibility, while testing i was able to confirm that the particles in the scene do restart when a different particle system is made invisible

i noticed when we delete a particle system it deletes the shared renderer Entity along with it ( my conclusion ), making all the particles disappear in the scene, when the reactors of these particle systems are triggered again( by changing a value from the editor ), it creates a new renderer and the particle system restarts

my hack, listens for removals with query and reacts on the same to ensure that we have a new renderer up and ready the moment we delete a particle system

the issue i encountered is that while i know the rendered entity is being deleted, i failed to find the how it gets deleted 
the reactor for the component does not run when i delete a particle system entity.  and the deletion code is triggered in the reactor.

PR #2135, opened 2025-06-20
[IR-11024] Create Viewer button stories

likethemammal:
## Summary

Created components and stories for `IconButton`, `MenuIconButton`, and `TextButton`

I used `class-variance-authority` for these. It helps keep the common styles and logic organized, and the variants can be automatically imported into Storybook controls

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,6 +160,7 @@
     "@hookstate/core": "4.0.1",
     "@swc/core": "1.7.35",
     "app-root-path": "3.1.0",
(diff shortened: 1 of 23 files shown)


Tortillaguy:
Not a big issue, but we don't have light mode colors and it makes the button story look broken
If we can default the background global arg (the toolbar one) to dark I think that suffices for now

image

Tortillaguy (approved):

PR #2147, opened 2025-06-20
Poll for Updated Messages

Tortillaguy:
## Summary
Add a refetch poll in the case that another user has updated their username. 
This will fetch the same messages but with the updated username saved per message object

## References
closes [IR-11115]

## QA Steps

[IR-11115]: https://tsu.atlassian.net/browse/IR-11115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/client-core/src/components/Glass/ChatProvider.tsx
```diff
@@ -68,6 +68,17 @@ const useChatMessages = (): ChatMessagesType => {
     }
   })
 
+  // Poll for new messages every 10 seconds
+  useEffect(() => {
```
HexaField: This shouldn't be necessary, new messages should be pushed from the backend via feathers publish events.

I would look into the backend, rather than the front end to solve this.

Tortillaguy: I believe this is an issue where the front-end is caching previous messages with the outdated username record in them

Tortillaguy: To clarify, this isn't a problem with newer messages. This is addressing a problem with older messages having outdated sender names

HexaField (changes requested):
we need to chain useGet with the userID for each message to display the name instead of getting it from the sender user object resolved, such that the username is pushed by feathers from the backend.

likethemammal (approved):

AidanCaruso:
Smoke tested, seems to work. useGet makes sense.

AidanCaruso (approved):

HexaField (approved):

PR #2154, opened 2025-06-21
Remove mipmaps for skyboxes, remove clone texture for envmaps(Improves WebGL crashes)

DanielBelmes:
## Summary
This was done purely by informed brute forcing. @MichaelEstes has watched my debugging. I and he found that when adding in a new skybox the GPU Process memory would skyrocket. 

To repeat the same steps:

1. Grab a couple of your favorite 8k skybox texture
2. Upload to ir-engine
3. Change the skybox URL to a new one
4. Check task manager for chrome and see that on first change you'll jump 200ish mega bytes
5. Change the skybox url to a new one
6. Check task manager for chrome and see that you'll jump a gigabyte
7. Keep repeating as it exponentially increases in memory usage until the GPU process crashes. You lose webgl context and it restarts. 

Through the day the first success I found was that setting generateMipMaps to false greatly reduced the jump in memory. Post that change I went from only being able to change the skybox 4-5ish times. To being able to change it 10ish times before GPU Process ran out of memory.

The next hurdle was that when dragging in models I kept seeing these exponential jumps in memory and more models also coincided with the Skybox upload GPU process memory increasing faster. I found this change in EnvSystem that clones the Skybox texture. Removing that made adding in apartment glb into the scene only increase GPU Process memory by less than 100mb.

**At this current point there is still something occurring with skybox changing that is massively increasing the memory pressure on the internal GPU Process. It happens much slower now but there's still something left which is why this is a draft for now. But honestly we should really merge it in as Wizard is unusable.**


DanielBelmes:
Had @MichaelEstes fix it in release1.1

PR #2155, opened 2025-06-21
If isDev resolve age verification to True.

DanielBelmes:
## References
closes #_insert number here_


SYBIOTE (approved):

HexaField (changes requested):
we have an env variable for this

PR #2157, opened 2025-06-23
IR-10330 - Update iR Logo in top left corner of /admin panel to take user back to /dashboard page

hurairahmateen:
## References
https://tsu.atlassian.net/browse/IR-10330


CITIZENDOT (changes requested):
image

Although this works in terms of purpose, Ticket mentions a dedicated **Back to Dashboard** Button. May be check with Kara if this is sufficient.

hurairahmateen:
> Although this works in terms of purpose, Ticket mentions a dedicated **Back to Dashboard** Button. May be check with Kara if this is sufficient.

I believe it was decided that only make the logo navigate. The ticket details are a bit outdated but let me confirm from Chris.

packages/client-core/src/admin/adminRoutes.tsx
```diff
@@ -64,12 +64,15 @@ const AdminTopBar = () => {
...
 
   return (
-    <div className="flex h-16 w-full items-center justify-between  px-8 py-4">
-      <img
-        src="static/app_logo.svg"
-        alt="Napster Engine Logo"
-        className={`h-7 w-7${theme.value === 'light' ? ' invert' : ''}`}
-      />
+    <div className="pointer-events-auto flex h-16 w-full items-center justify-between px-8 py-4">
+      <a href="/dashboard">
```
HexaField: the dashboard doesn't exist in the core engine. it should redirect to `/` and that will redirect to `/dashboard`

HexaField (changes requested):

CITIZENDOT (approved):

hanzlamateen (approved):

PR #2173, opened 2025-06-24
Add opacity transitions for light helper components

SYBIOTE:
Introduce opacity transitions for various light helper components, enhancing visual feedback during interactions. Define related timeout and interpolation settings for smoother transitions.


packages/editor/src/classes/helper/DirectionalLightHelper.ts
```diff
@@ -110,13 +112,31 @@ export const DirectionalLightHelperReactor: React.FC = (props: { parentEntity; i
     setComponent(helperEntity, LineSegmentComponent, {
       name: 'directional-light-helper',
       geometry: mergedGeometry?.clone(),
-      color: directionalLight.color.value
+      color: directionalLight.color.value,
+      opacity: 0
     })
+    // @ts-ignore causes issues with the type system value inferred as never
```
HexaField, line 118: we definitely can't be adding ts-ignore calls here.

@speigg looks like we do indeed have problems with transition types :(

SYBIOTE: No problem i can call the function by wrapping it in an any

HexaField: that's equivalent. we should try to fix this in a follow up PR.

SYBIOTE: Ok so no action needed right now?

MbfloydIR: approving since @HexaField said we can fix in a followup PR

MbfloydIR (approved):

PR #2188, opened 2025-06-25
Optimize entity tree hooks and other reactors

HexaField:
## References
closes #_insert number here_


packages/engine/src/avatar/systems/AvatarTransparencySystem.tsx
```diff
@@ -114,18 +117,17 @@ const AvatarReactor = (props: { entity: Entity }) => {
...
+    getMutableComponent(props.rootEntity, TransparencyDitheringRootComponent).materials.merge([props.entity])
+    setComponent(entity, TransparencyDitheringPluginComponent)
+    return () => {
+      getMutableComponent(props.rootEntity, TransparencyDitheringRootComponent).materials.set((current) =>
+        current.splice(current.indexOf(props.entity), 1)
+      )
+      removeComponent(entity, TransparencyDitheringPluginComponent)
     }
-  }, [materialComponentEntities, !!rootDitheringComponent])
+  }, [])
```
AidanCaruso: On avatar switch the dithering effect is no longer applied to the new materials, we should be using the material component as a dependency here.

AidanCaruso (changes requested):
Looks good, these are some nice optimizations for now. I did notice an issue with avatar transparency not getting applied on avatar switch.

AidanCaruso (approved):

PR #2197, opened 2025-06-25
Remove code from AvatarInputSystem that erroneously zooms the camera out

AidanCaruso:
## References
closes #_insert number here_


the-reality-engineer (approved):
Intention here was to back up the camera somewhat while the avatar is moving towards the camera so that the user can see a bit more of where they are going, but we can implement this better later.

PR #2200, opened 2025-06-25
[Release 1.1][IR-11094] Env map breaking textures

MichaelEstes:
## References
closes #_insert number here_


AidanCaruso:
Interesting, going to smoke test this shortly. I noticed this is just a small fraction of the counterpart aimed at dev, does this fix less? And does unwrapping the state value with .get(NO_PROXY) also work to prevent hookstate errors or is it just the removal of the .clone() call that prevents problems?

AidanCaruso (dismissed):
Screenshot shows what happens after the avatar preview renderer is disposed.
Screenshot 2025-06-25 at 5 16 23 PM
I recall the clone call this PR removes being added to prevent the texture being disposed of on the disposal of renderers using that envmap texture as a background, which became an issue with the changes made in the resource caching PR. I feel like removing the clone call may be the right move but maybe we'd also have to prevent the disposal of textures shared across renderers?

MichaelEstes:
> Interesting, going to smoke test this shortly. I noticed this is just a small fraction of the counterpart aimed at dev, does this fix less? And does unwrapping the state value with .get(NO_PROXY) also work to prevent hookstate errors or is it just the removal of the .clone() call that prevents problems?

The other changes in the dev PR were around cleaning up the skybox component reactor flow and how it's resources are cleaned up, but they're not needed for release.

.clone will cause a texture upload now if the TextureMemoryManager has overwritten the source object, but because the TextureMemoryManager has overwritten the source object there's no data in it to upload which causes errors. .value might work, but there might be edge cases with it being a proxy object in Three's code.

@DanielBelmes and I are looking into the multiple renderer issue, we should have a fix for that soon

DanielBelmes (approved):

PR #2224, opened 2025-06-26
Wizard Welcome Flow Compress & Publish Step Broken

hzraslan:
## References
closes https://tsu.atlassian.net/browse/IR-11247

### DIFF
--- a/packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
+++ b/packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
@@ -105,8 +105,6 @@ type AddEditLocationModalProps = Readonly<{
   sceneModified?: boolean
   inStudio?: boolean
   projectFullName?: string
-  onCompress?: (newGltfUrl: string) => void
-
   onPublish?: () => Promise<void>
   onPublishSuccess?: (location: LocationType) => void
 }>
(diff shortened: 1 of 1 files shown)


packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
```diff
@@ -105,6 +105,7 @@ type AddEditLocationModalProps = Readonly<{
   sceneModified?: boolean
   inStudio?: boolean
   projectFullName?: string
+  wizardState?: any
```
DanielBelmes: Could we change this to an on Compress functionality

DanielBelmes (changes requested):

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
```diff
@@ -331,6 +333,7 @@ export default function AddEditLocationModal(props: AddEditLocationModalProps) {
 
             setComponent(gltfEntity, NameComponent, fileName + '-compressed')
             setComponent(gltfEntity, GLTFComponent, { src: newGLTFURL })
+            props?.onCompress && fileName !== scenename && props.onCompress(newGLTFURL)
```
DanielBelmes: So this calls onCompress for every compressed entity except the scene? Might be annoying but maybe it should be called onCompressing?

hzraslan: I am not sure if even that function is working, so removed

DanielBelmes (approved):

PR #2265, opened 2025-06-30
[MT-DEV]Editor | Toolbar: Selected Render Mode Persists in Toolbar After Refresh, but Visual Effect Is Lost on Scene

MbfloydIR:
https://tsu.atlassian.net/browse/IR-11282
update the useEffect that updates the RenderMode settings to react to the renderer's effect composer also

## References
closes #_insert number here_


BogieBogard (approved):
Approved. Nice job identifying the root cause and providing a clean, minimal fix here.

PR #2273, opened 2025-07-01
Michael/ir 9972 video controls

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-9972
Updated the Client Input Hooks to use a pointer event obj with the “mapped” pointer Id value

Fixed the Video xrui controls from being able to be clicked on

## References
closes #_insert number here_


BogieBogard (approved):
Approved. I think your approach of using `mappedPointEvent` with the clone is a sound logic strategy for solving the problem of users not being able to pause/play videos from the expected distance.

However, I haven't been able to verify this one on my local yet. I'm going an error in my local spaces currently with videos (both release1.1 and your branch). I'll fix it shortly. Once resolved, I'll circle back if this update doesn't work as expected. Approving now to not slow you down.

the-reality-engineer:
Looks reasonable to me!

PR #2274, opened 2025-07-01
Ir 10563 Add avatar confirm button, remove avatar change on exit, and fix mobile avatar view

DanielBelmes:
## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/src/common/components/Avatar/Avatar2.tsx
+++ b/packages/client-core/src/common/components/Avatar/Avatar2.tsx
@@ -78,9 +78,9 @@ const Avatar = ({
         <div
           className={twMerge(
             'relative overflow-hidden rounded-2xl bg-gradient-to-b from-blue-400 to-blue-600',
-            size ? `w-[${size}px] h-[${size}px]` : '',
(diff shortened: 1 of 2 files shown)


packages/client-core/src/common/components/Avatar/Avatar2.tsx
```diff
@@ -78,7 +78,7 @@ const Avatar = ({
         <div
           className={twMerge(
             'relative overflow-hidden rounded-2xl bg-gradient-to-b from-blue-400 to-blue-600',
-            size ? `w-[${size}px] h-[${size}px]` : 'h-32 w-32',
+            size ? `w-[${size}px] h-[${size}px]` : '',
```
likethemammal: Dynamic classes like this dont actually get generated by Tailwind like you would expect here. You'll have to use style to set the width and height.

Also, ideally we use `rem` over `px` where we can.

```
style={size ? { width: `${size}rem`, height: `${size}rem` } : {}}

likethemammal (changes requested):

likethemammal (approved):

PR #2275, opened 2025-07-01
Michael/ir 11484 vidoe backside controls

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-11484

Added a backside xrui for the video media controls.
Parented it to the front side and rotated it 180 degrees.
Updated the "update" logic that handles the scale / opacity for transitions in/out to use a list ( parent xrui and any children xrui)

## References
closes #_insert number here_


DanielBelmes (approved):
I think it's fine for now. Ideally i'd like to explore detecting what side is being viewed and transform the xrui correctly.

PR #2276, opened 2025-07-02
IR-11343 mt dev qat client viewer not all users in same instance can see each other

quaiquai-ir:
## References
https://tsu.atlassian.net/browse/IR-11343

Screenshot 2025-06-30 at 11 12 33 AM


packages/engine/src/avatar/components/AvatarAnimationComponent.ts
```diff
@@ -235,20 +269,35 @@ const toesAngle = new Euler(Math.PI / 6, 0, 0)
 
 /**Rewrites avatar's bone quaternions and matrices to create a T-Pose, assuming all bones are the identity quaternion */
 export const enforceTPose = (entity: Entity) => {
-  const bones = getComponent(entity, AvatarRigComponent).bonesToEntities
+  const rigComponent = getOptionalComponent(entity, AvatarRigComponent)
+  if (!entityExists(entity) || !rigComponent) {
```
AidanCaruso: Does this entity actually not exist at times? If so it's an issue upstream with the useQuery in the AvatarAnimationSystem reactor. It should not be running and returning EIDs that don't exist - we shouldn't have to be checking for this. If this is sometimes causing errors, I understand it may be a valid solution for 1.1, otherwise we've got some serious issues to address with the useQuery pattern/implementation.

packages/engine/src/avatar/components/AvatarAnimationComponent.ts
```diff
@@ -84,19 +84,30 @@ export const AvatarRigComponent = defineComponent({
   }),
 
   setBone: (toRigEntity: Entity, boneEntity: Entity, boneName: VRMHumanBoneName) => {
+    // Check if entities exist before setting bone mapping
+    if (!entityExists(toRigEntity) || !entityExists(boneEntity)) return
```
AidanCaruso: We should probably be ensuring that references to entities in the AvatarRigComponent are always cleared out when those entities are unloaded, which is a less brittle alternative to checking if the entities exist whenever we attempt to access components from them

AidanCaruso:
The NetworkState/SpawnSystem changes all make sense to me, I'm just worried about the entityExists checks which are indicative of some deeper problems with reference cleanup and possibly useQuery issues. If each of those checks is important for preventing errors, we can merge as-is for 1.1 and then we can dig into the core of these issues after release.

quaiquai-ir:
> The NetworkState/SpawnSystem changes all make sense to me, I'm just worried about the entityExists checks which are indicative of some deeper problems with reference cleanup and possibly useQuery issues. If each of those checks is important for preventing errors, we can merge as-is for 1.1 and then we can dig into the core of these issues after release.

You're right. Those were some older checks I added before I solved the problem in the NetworkState and SpawnSystem. Removing them still solves the issue.

AidanCaruso (approved):

PR #2277, opened 2025-07-02
Solid color skybox fallback on iOS

AidanCaruso:
## Summary
On iOS it should look something like this if all is working properly
image

## References
closes https://tsu.atlassian.net/browse/IR-11467

### DIFF
--- a/packages/engine/src/scene/components/SkyboxComponent.ts
+++ b/packages/engine/src/scene/components/SkyboxComponent.ts
@@ -105,6 +105,7 @@ export const SkyboxComponent = defineComponent({
       // temporary logic to force solid color on iOS. We use a separate variable to keep track of this
       // so we can fall back to a sensical default value (in spite of bad serialized color data)
       // in the event the user did not set up a color themselves
+      /**@todo implement smart asset LOD system to load lower resolution skybox textures on iOS */
       if (iOS) forceColorFallback.set(true)
     }, [])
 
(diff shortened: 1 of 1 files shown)


koktavy (approved):
Tested and confirmed on iOS Safari that the skybox does fallback to a solid color.

https://ir-engine.qat-dev.theinfinitereality.io/location/default?instandId=2462078e-7fa0-4e25-a820-5a60e1cf5f9e

packages/engine/src/scene/components/SkyboxComponent.ts
```diff
@@ -97,6 +100,14 @@ export const SkyboxComponent = defineComponent({
       }
     }, [])
 
+    const forceColorFallback = useHookstate(false)
+    useEffect(() => {
+      // temporary logic to force solid color on iOS. We use a separate variable to keep track of this
```
koktavy, line 105: Only thing I might add is a description of what the permanent logic might be or when to expect that, so we don't have this temp logic sitting forever.

AidanCaruso: Yeah fair point. I'll add a todo

koktavy: Not sure if there's already a Jira ticket for a more permanent/ideal fix but would be great to link that # in the comment

koktavy: Ah perfect

PR #2282, opened 2025-07-02
remove unwanted DELTAY objects created when adding colliders/rigidbodies

danskate:
## References
closes [#IR-11485](https://tsu.atlassian.net/browse/IR-11485)

## QA Steps
Create scene
Press ` for debug menu
Upload a model
Drag and drop into scene
Toggle Apply Colliders
Press blue “Add rigidbody” button
Observe performance


MichaelEstes (approved):
This does seem to fix the lag and I don't see any behaviors changing from removing this. lineEntitiesState is no longer being used with this change though and can probably be removed.

the-reality-engineer:
What was the performance bottleneck in the removed code?

SYBIOTE (changes requested):
we need the deltaY entity for on icon hover, to show distance from the horizontal plane 
it was implemented as part of the design direction, 

is this causing editor lag?

SYBIOTE:
adding colliders and rigid bodies to an entity will add an icon to them in the viewport as well, 

the removed code will end up affecting all gizmo icons

PR #2285, opened 2025-07-02
IR-11474: show size unit in properties menu for selected files

jose-galvan:
## Summary

Show size unit in properties menu for selected files
**ticket:** https://tsu.atlassian.net/browse/IR-11474

Screen Recording 2025-07-02 at 1 46 09 p m

## References
closes #_insert number here_


BogieBogard (approved):
Approved. Verified that the size text is rendering as expected with this update. The solution is clean and self-contained to the one new useEffect.

PR #2289, opened 2025-07-03
fix(IR-11489): fix copy/paste array filtering and sourceId validation

kattsushi:
## Summary
fix(editor): resolve copy/paste issues in hierarchy panel

- Fix array filtering bug where delete components[UUIDComponent.jsonID] was incorrectly trying to delete from array instead of filtering

Fixes issue where:
1. Pasted assets would disappear from viewport when moved/translated
2. Original entities would disappear from hierarchy panel when copied after being transformed
3. Copy/paste would fail for entities with composite sourceIds

Behavior before:

https://github.com/user-attachments/assets/66b78077-c254-402e-9251-403fefd9082b

---

Behavior fixed

https://github.com/user-attachments/assets/b99c5fb2-aea0-4cf5-a34d-a8d52ecbe7fe

## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/panels/hierarchy/helpers.ts
+++ b/packages/editor/src/panels/hierarchy/helpers.ts
@@ -104,8 +104,9 @@ export const pasteNodes = (parentEntity?: Entity) => {
 
   const ProcessEntityData = (parentEntity: Entity | undefined, nodeEntitiesData: EntityCopyDataType[]) => {
     nodeEntitiesData.forEach((nodeEntityData) => {
-      const components = nodeEntityData.components.map((c) => ({ name: c.name, props: c.json }) as ComponentJsonType)
-      delete components[UUIDComponent.jsonID]
+      const components = nodeEntityData.components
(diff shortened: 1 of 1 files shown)


BogieBogard:
@kattsushi I noticed this PR is pointed to dev, but the ticket is set for fix version 1.1. Does this fix need to be based off of release1.1 and be targeting that branch?

BogieBogard (approved):
Approved. This logic fix makes sense to me. I added a question asking if this should be set to dev branch or release1.1 branch though, because it's not quite clear.

Also, I reproduced the bug on my local, then ran your branch and confirmed it worked as expected.

kattsushi:
> Approved. This logic fix makes sense to me. I added a question asking if this should be set to dev branch or release1.1 branch though, because it's not quite clear.
> 
> Also, I reproduced the bug on my local, then ran your branch and confirmed it worked as expected.

Thank you Matt, yes this was happen in dev and qa, so next after merge i'm going to cherry pick the commit and create a PR against to release1.1 branch as well

PR #2297, opened 2025-07-04
Refactor CallbackComponent for authoring entity management

SYBIOTE:
Refactor the CallbackComponent to utilize the authoring entity for managing callbacks, ensuring that callbacks set on the authoring entity automatically reflect in the simulation. This change enhances the consistency and reliability of callback handling.


HexaField (commented):
When/why are we adding callbacks on authoring entities? Callbacks are not serializable, therefore should not be on the authoring layer.

SYBIOTE:
> When/why are we adding callbacks on authoring entities? Callbacks are not serializable, therefore should not be on the authoring layer.

Hmm, all entities in the scene already have callbacks for visibility, because of the object component 

The trigger component 
Properties panel, already queries over callbacks in the authoring layer

I assume then we dont want absolutely any callbacks in the authoring layer

MbfloydIR (approved):

PR #2301, opened 2025-07-07
[IR-11029] Add current speaker highlight

likethemammal:
## Summary

image

## QA Steps

- Go to viewer
- Go to videos page
- Turn on mic and start speaking
- Observe that a border around the speaker's video or thumbnail appears as the speaker speaks

### DIFF
--- a/packages/client-core/src/components/Glass/VideoMenu.tsx
+++ b/packages/client-core/src/components/Glass/VideoMenu.tsx
@@ -24,7 +24,7 @@ Infinite Reality Engine. All Rights Reserved.
 */
 
 import { ChevronLeftMd as ArrowLeftIcon, ChevronRightMd as ArrowRightIcon } from '@ir-engine/ui/src/icons'
-import React, { useEffect, useMemo, useRef, useState } from 'react'
+import React, { createContext, useContext, useEffect, useMemo, useRef, useState } from 'react'
 import {
   BsEyeSlashFill as CameraOffIcon,
   BsEyeFill as CameraOnIcon,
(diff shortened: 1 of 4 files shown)


svyoung (approved):
Looks good

Tortillaguy:
I'd opt for using a state object over context/ provider but this is fine

Tortillaguy (approved):

likethemammal:
> I'd opt for using a state object over context/ provider but this is fine

Not sure exactly what you mean. Most of the logic for these changes to work is in the hook `useUserMediaWindowsHook`. The context wraps it so the logic isnt doubled up for the 2 areas of the UI that need it. Most of the state values are already coming from Hookstate

packages/client-core/src/components/Glass/VideoMenu.tsx
```diff
@@ -272,15 +275,37 @@ const arrowsContainer = `
...
 const numVideosPerPage = 6
 
-export const VideoMenu = ({
-  videos = [],
-  soundIndicators
-}: {
-  videos: WindowType[]
-  soundIndicators: State<Record<string, boolean>>
-}) => {
+const VideosProviderContext = createContext(
```
Tortillaguy, line 278: I was referring to this provider/ context pair. 
I don't see why you wouldn't just `useUserMediaWindowsHook` where needed. This provider is just renaming `_windows` to `videos` in a wrapper use hook

Tortillaguy: And I was referring to the hyperflux (`defineState`) state objects, not `useState`

Tortillaguy (approved):

packages/client-core/src/components/Glass/VideoMenu.tsx
likethemammal: `useUserMediaWindowsHook` spins up a bunch logic and instances of state that would be redundant to call more than once and ends up causing bugs (duplicate audio, etc). It could prob be rewritten to only use hyperflux but the goal was to keep the fix simple and only call the hook once. Context allows me to do that without requiring the hook call in the top parent and rerendering all of the Glass UI

PR #2305, opened 2025-07-08
[IR-11508] MT Int the video s audio gain changes dramatically after engaging chat

MichaelEstes:
## Summary
Makes sure the gain value isn't reset to 1 when the MediaStreamState reactor is run
Lower scene audio elements' volume when voice chat is occurring 
## References
closes #_insert number here_


packages/hyperflux/src/media/WebRTCPeerConnection.tsx
```diff
@@ -23,6 +23,7 @@ All portions of the code written by the Infinite Reality Engine team are Copyrig
 Infinite Reality Engine. All Rights Reserved.
 */
 
+import hark from 'hark'
```
HexaField: Hark is a front-end module, so if we can avoid putting this in hyperflux that'd be great.

HexaField (dismissed):

packages/hyperflux/src/media/WebRTCPeerConnection.tsx
```diff
@@ -397,6 +399,29 @@ export const MediaReceiveChannelReactor = (props: {
...
+    const cleanup = () => {
+      unmounted = true
+      AudioDuckingState.removeEvent(peer)
+      harkEvents.stop()
+    }
+
+    harkEvents.on('stopped_speaking', cleanup)
+
+    return cleanup
+  })
```
HexaField: Needs empty deps array

Tortillaguy (approved):

AidanCaruso (approved):

PR #2306, opened 2025-07-08
Grass System WIP

Tortillaguy:
## References
closes #_insert number here_


HexaField (changes requested):
FYI we have expansive plans for the instancing and variant systems already underway you can see here. https://github.com/ir-engine/ir-engine/pull/2091

It's best to not overload primitive concepts such as the instancing component with functionality, higher level components should take advantage of them rather than build opinions directly into them.

packages/engine/src/scene/functions/useEntity.ts
```diff
@@ -0,0 +1,51 @@
...
+import { useHookstate } from '@hookstate/core'
+import { createEntity, Entity, removeEntity, UndefinedEntity } from '@ir-engine/ecs'
+import { useEffect } from 'react'
+
+interface EntityOptions {
+  setup?: (entity: Entity) => void
+  cleanup?: (entity: Entity) => void
+}
+
+export const useEntity = ({ setup, cleanup }: EntityOptions = {}) => {
```
HexaField, line 35: These kinds of APIs generally aren't as helpful as they appear to be. I'd prefer to keep hook composition like this to the bare essentially, especially as we continue to improve the ECS and build higher level APIs that makes this sort of utility unnecessary.

packages/engine/src/scene/systems/InstancingSystem.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { InstancingComponent } from '../components/InstancingComponent'
+import { VariantComponent } from '../components/VariantComponent'
+import { useEntity } from '../functions/useEntity'
+
+interface Level {
+  src: string
+  metadata: Record<string, any>
+}
+
+export const InstancingSystem = defineSystem({
```
HexaField, line 54: This system is making some assumptions about what instances and variants will be used for. We have plans to refactor these that will achieve the same thing without making assumptions.

HexaField (changes requested):

PR #2343, opened 2025-07-14
Michael/ir 11910 bounding box offset

MbfloydIR:
## Summary
https://tsu.atlassian.net/browse/IR-11910

The bounding box helper was being placed at the origin of the entity, instead of being placed at the center of the Bounding Box Component's box.

## References
closes #_insert number here_


SamMazerIR (approved):
just need to fix failing test, otherwise looks good

PR #2346, opened 2025-07-15
fix: scene was not loading due to null === object

DanielBelmes:
## References
closes #_insert number here_


kattsushi (approved):
LGTM

AidanCaruso (approved):

HexaField (commented):
please add a regression test for this

PR #2352, opened 2025-07-16
IR-11585 lowering minimum radius of primitive geo shapes

SamMazerIR:
## Summary
lowering minimum radius, setting a minimum for cylinders so they don't visually break at 0

## References
[https://tsu.atlassian.net/browse/IR-11585](https://tsu.atlassian.net/browse/IR-11585)


kattsushi (approved):
LGTM

PR #2353, opened 2025-07-16
IR-11725  fix rotating camera on Mobile viewer

svyoung:
## Summary
Description: On iOS mobile, it seems that pressing the rotate camera button option, it is instead going through all the camera zoom levels prior to rotating the camera.

Expected: Pressing rotate button switches between the user-facing and the forward facing camera on iPhone.
Actual: Pressing the rotate button seems to go from user facing, to forwards facing, to then zooming multiple times and then finally back to user facing.

Reproduction:
Enter published scene on iOS.
Enable camera
Press rotate camera button
Rotate camera to forward camera
Rotate camera to face camera
Rotate camera to forward camera again.
Observe that camera will go through different zoom levels prior to rotating back to face camera.

## References
[IR-11725](https://tsu.atlassian.net/browse/IR-11725)

## QA Steps
- Go to a location on mobile
- Enable camera
- Toggle between camera switch
- It should only be displaying forward and back facing camera

[IR-11725]: https://tsu.atlassian.net/browse/IR-11725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/hyperflux/src/media/MediaStreamState.ts
+++ b/packages/hyperflux/src/media/MediaStreamState.ts
@@ -55,7 +55,7 @@ export const MediaStreamState = defineState({
     /** Whether the audio is enabled or not. */
     microphoneEnabled: false,
     /** Whether the face tracking is enabled or not. */
-    /** @deprecated face tracking has been disabled */
+    /** @deprecated - face tracking has been disabled */
     faceTracking: false,
     /** Video stream for streaming data. */
     webcamMediaStream: null as MediaStream | null,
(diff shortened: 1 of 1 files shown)


packages/hyperflux/src/media/MediaStreamState.ts
```diff
@@ -295,6 +295,31 @@ export const MediaStreamService = {
       return false
     }
     console.log('Cycle camera')
+
+    const isMobile = /Android|iPhone|iPad|iPod/i.test(navigator.userAgent)
```
MbfloydIR, line 299: Do we already have a method to determine if the device is mobile that we can use?
If not, we should move this to a util file that can be reused for other places that need to determine if device is mobile.

PR #2358, opened 2025-07-17
Replace remaining onBeforeCompile plugins with defineMaterialPlugin

HexaField:
## Summary

- [x] VariantComponent
- [x] applyScreenshareToTexture
- [x] FogSystem
- [x] XRDepthOcclusion

## References
closes #_insert number here_


HexaField:
Will await https://github.com/ir-engine/ir-engine/pull/2359 such that defineMaterialPlugin and some uses of it can be moved to the spatial package

